PR #2, opened 2015-05-05
Fixup roles so that fresh provision succeeds

CloudNiner:
Moves some prereqs for pip installing the ashlar dependencies to the `driver.ashlar` role.


kshepard:
+1

PR #5, opened 2015-05-07
Feature/travis ci

CloudNiner:


app/driver/settings_travis.py
```diff
@@ -0,0 +1,147 @@
+
+"""
+Django settings for DRIVER project testing.
```
CloudNiner: Is there a better way to override this file? All that is different is the DATABASES key.

If there is an easier way using the base settings.py file I don't have secrets.yml file either. So that would have to be stubbed out somehow.

kshepard: As discussed, may be worth trying to add a try/except when loading the secrets file with a dict of defaults when it's not available. Then the travis settings file can just import everything and override only what it needs.

kshepard:
+1

app/driver/settings_travis.py
CloudNiner: Done. This seemed to work well.

kshepard:
Nice, looks good.

PR #8, opened 2015-05-07
Use npm2.1.1+ to fix npm install race conditions

CloudNiner:


kshepard:
+1, Looks like it passed!

PR #10, opened 2015-05-08
WIP resources list

moradology:
Some significant changes here:
1. There's not an implementation of the notion of a primary geography (a primary field for geographies is odd in the abstract - perhaps the applications which consume these geographies should decide how to appropriately organize them)
2. It may not always be boundaries that we're dealing with; if so, fine, but we needn't confine ourselves to the idea of a spatial extent at the outset. I've converted all 'boundaries' to 'geographies' given this.
3. Some logic surrounding uniqueness of labels on geographies has been added - this required some modifications to Ashlar as well so that namespaces wouldn't be polluted by failed uploads.


schema_editor/app/index.html
```diff
@@ -54,6 +54,8 @@
 
         <!-- build:js({.tmp,app}) scripts/scripts.js -->
         <script src="scripts/config.js"></script>
+        <script src="scripts/resources/module.js"></script>
+        <script src="scripts/resources/Boundaries.js"></script>
```
kshepard: How about: `boundaries-service.js` instead?

kshepard:
Looking good so far.

schema_editor/app/index.html
moradology: Sounds good to me

schema_editor/app/scripts/resources/boundaries-service.js
```diff
@@ -0,0 +1,29 @@
...
+
+        var stdResource = $resource(urlString, {id: '@uuid'}, {
+            query: {
+                method: 'GET',
+                transformResponse: function(data) { return angular.fromJson(data).results; },
+                isArray: true
+            }
+        });
+
+        return resource;
```
CloudNiner: Interesting. By abstracting $resource, this prevents you from doing something like:
```
function (Boundaries) {
    var newBoundary = new Boundaries({uuid: 'abcd', ... });
   newBoundary.save().then(function () { ... });
})
```

Which may or may not be a bit clearer, and would replace the `postWithFile` method (which I don't see but am assuming is just not implemented yet?).

schema_editor/app/scripts/resources/boundaries-service.js
```diff
@@ -0,0 +1,29 @@
...
+          'get': stdResource.get,
+          'post': postWithFile,
+          'remove': stdResource.remove,
+          'delete': stdResource.delete
+        };
+
+        var stdResource = $resource(urlString, {id: '@uuid'}, {
+            query: {
+                method: 'GET',
+                transformResponse: function(data) { return angular.fromJson(data).results; },
```
CloudNiner: Since this is doing something non-standard, you'll want to add a test to ensure query returns what is expected. I can point you to a few service/http test examples if you can't find one.

schema_editor/app/scripts/resources/boundaries-service.js
moradology: Yeah, this is a bit up in the air. I'm looking to maybe modify the file upload directive from OTI so that the data we want can just be patched on top of our uploaded shapefile using $resource. Angular's support for file uploading is a surprising blind spot in the framework

schema_editor/app/scripts/views/boundary/list-controller.js
```diff
@@ -2,15 +2,16 @@
     'use strict';
 
     /* ngInject */
-    function BoundaryListController() {
+    function BoundaryListController(Boundaries) {
         var ctl = this;
         initialize();
 
         function initialize() {
+          ctl.bounds = Boundaries.query();
```
CloudNiner: Consider renaming `bounds` to `boundaries`, since it's a bit more explicit. The 4 extra characters gets minified away anyways!

schema_editor/app/scripts/resources/boundaries-service.js
CloudNiner: Ah ok. That would work, although that directive is designed with async upload operations in mind and this is a synchronous one so to keep things simple you could probably just use the angular-file-upload plugin directly to start if necessary.

schema_editor/.jshintrc
```diff
@@ -3,7 +3,7 @@
   "browser": true,
   "esnext": true,
   "bitwise": true,
-  "camelcase": true,
+  "camelcase": false,
```
kshepard: This shouldn't be disabled app-wide. Where there are unavoidable non-camelcase variables you can surround the code in: `/* jshint camelcase: false */` ... `/* jshint camelcase: true */`

moradology: Got it.

schema_editor/app/scripts/views/geography/add-partial.html
```diff
@@ -0,0 +1,99 @@
...
+        </div>
+        <form name="geoAddForm" ng-submit="" novalidate>
+            <div class="form-group col-sm-12">
+                <label for="geo-label">Geography Label</label>
+                <input type="text"
+                       name="label"
+                       ng-model="geoAdd.geoFields.label"
+                       class="form-control"
+                       id="geo-label"
+                       placeholder="Manila"
```
kshepard: Maybe change these Manila defaults to something else to avoid confusion, since it's more a collection of areas rather than a single area. Perhaps something like: "Pilot Areas", "Police Districts", or something along those lines.

schema_editor/app/scripts/resources/geography-service.js
```diff
@@ -0,0 +1,74 @@
...
+                Upload.upload({
+                    url: urlString,
+                    method: 'POST',
+                    fields: {'label': label,
+                             'color': color},
+                    file: file,
+                    fileFormDataName: 'source_file'
+                }).progress(function (evt) {
+                    var progressPercentage = parseInt(100.0 * evt.loaded / evt.total);
+                    console.log('progress: ' + progressPercentage + '% ' + evt.config.file.name);
```
kshepard: Either remove or inject $log and use $log.debug.

schema_editor/app/scripts/views/geography/list-partial.html
```diff
@@ -0,0 +1,37 @@
...
+            <!-- BARANGUAYS -->
+            <div ng-repeat="geo in geoList.geographies | filter:geoList.noFailures() | orderBy:'label'" class="row">
+                <div class="type">
+                    <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 10 22" enable-background="new 0 0 10 22" xml:space="preserve">
+                        <path opacity="0.5" fill-rule="evenodd" clip-rule="evenodd" fill="#4A4A4A" d="M4.6,20.6L8,17.3H1.3L4.6,20.6z M1.3,4H8L4.6,0.7 L1.3,4z M8,10.7c0-1.8-1.5-3.3-3.3-3.3c-1.8,0-3.3,1.5-3.3,3.3c0,1.8,1.5,3.3,3.3,3.3C6.5,14,8,12.5,8,10.7z"/>
+                    </svg>
+                    <h4>
+                        <a ui-sref="geo.edit({ uuid: geo.uuid })" role="button" class="edit">{{ geo.label }}</a>
+                    </h4>
+                    <a ng-click="geoList.deactivateRecordType(recordType)" role="button" class="delete">
```
kshepard: Looks like deleting isn't implemented correctly.

schema_editor/app/scripts/resources/geography-service.js
```diff
@@ -0,0 +1,74 @@
...
+         *
+         * @param {integer} status Status code of an error response
+         * @return {string} Meaning of status code
+         */
+        res.errorMessage = function(status) {
+            var intension;
+            if (status === 409) {
+                 intension = 'Uniqueness violation - verify that your geography label is unique';
+            } else if (status === 500) {
+                intension = 'Error - check that your upload is a valid shapefile';
```
kshepard: I like the error handling! Though from my testing, it seems that when there's an error, there's still a record in the database, and it then shows up in the list.

schema_editor/app/scripts/resources/geography-service.js
```diff
@@ -0,0 +1,74 @@
...
+                var file = files[0];
+                Upload.upload({
+                    url: urlString,
+                    method: 'POST',
+                    fields: {'label': label,
+                             'color': color},
+                    file: file,
+                    fileFormDataName: 'source_file'
+                }).progress(function (evt) {
+                    var progressPercentage = parseInt(100.0 * evt.loaded / evt.total);
```
kshepard: I always pass the base 10 parameter to parseInt, since there are some weird js edge cases if it's unspecified.

schema_editor/test/spec/views/geography/list-directive.spec.js
```diff
@@ -0,0 +1,36 @@
...
+describe('ase.views.geography: geoList', function () {
+
+    beforeEach(module('ase.mock.resources'));
+    beforeEach(module('ase.templates'));
+    beforeEach(module('ase.views.geography'));
+
+    var $compile;
+    var $httpBackend;
+    var $rootScope;
+    var RecordTypes;
```
kshepard: In all of these tests, you can get rid of RecordTypes. 
Also, you should probably add a test for the geography service.

kshepard:
Looks really good. I wasn't able to successfully upload shapefiles for some reason, even after pulling down your Ashlar branch. I'll probably need a quick hand getting it up and running to test it out a little bit.

schema_editor/app/scripts/views/geography/list-partial.html
moradology: Good catch.

kshepard:
Shapefile upload works great now. +1 after addressing the comments.

PR #11, opened 2015-05-08
Add CORS headers

kshepard:


moradology:
I can barely believe how little code this required. +1

PR #13, opened 2015-05-12
Allow adding and editing record types

kshepard:
Also adds some rudimentary form validation logic. Will add tests in a followup PR.


schema_editor/app/scripts/views/recordtype/add-controller.js
```diff
@@ -2,15 +2,27 @@
...
+        }
 
+        /*
+         * Creates the record type and switches to the list view on success
+         */
+        function submitForm() {
+            RecordTypes.create(ctl.recordType, function() {
+                $state.go('rt.list');
+            }, function(error) {
+                console.error('Error while adding recordType: ', error);
```
CloudNiner: Change all `console.error` calls to `$log.debug` and inject `$log` where necessary. There is a global flag in app.js that sets the $logProvider debug mode to true when the app is provisioned for development.

CloudNiner:
+1 with the minor logging changes

schema_editor/app/scripts/views/recordtype/add-controller.js
kshepard: Cool, will do.

PR #16, opened 2015-05-12
Feature/related content

CloudNiner:
Add/Edit/Delete related content views

Currently no way to access the `/recordtype/detail/:uuid` view from the UI, pending sidebar implementation. Can be navigated to directly via url bar if you create a recordtype and paste the proper UUID.

The related controllers feel kind of messy already, since the page technically isn't ready until both the RecordTypes and RecordSchemas promises resolve and there's nested function calls. All ears if you have ideas on a relatively easy way to refactor.

### DIFF
--- a/schema_editor/app/scripts/schemas/schemas-service.js
+++ b/schema_editor/app/scripts/schemas/schemas-service.js
@@ -1,0 +1,27 @@
+
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function Schemas() {
+        var module = {
+            Object: Object
+        };
+        return module;
+
+        function Object() {
+            return {
+                type: 'object',
+                title: '',
+                plural_title: '',
+                description: '',
+                properties: {},
+                definitions: {}
+            };
+        }
+    }
+
+    angular.module('ase.schemas')
+    .service('Schemas', Schemas);
+
+})();
--- a/schema_editor/app/scripts/views/recordtype/detail-add-edit-partial.html
+++ b/schema_editor/app/scripts/views/recordtype/detail-add-edit-partial.html
@@ -1,0 +1,54 @@
+
+<div class="form-area">
+    <div class="col-sm-8 col-sm-offset-3">
+        <div class="form-area-heading">
+            <h2 ng-show="!rtDetail.schemaKey">Add New Related Content</h2>
+            <h2 ng-show="rtDetail.schemaKey">Edit Related Content</h2>
+            <div class="content-border"></div>
+        </div>
+
+        <form name="rtForm" novalidate ng-submit="rtDetail.submitForm()">
+            <div class="form-group col-sm-6">
+                <label for="single-title">Single Title</label>
+                <input required type="text" name="label" ng-model="rtDetail.definition.title"
+                       class="form-control" id="single-title" placeholder="Accident"
+                       ng-minlength="3" ng-maxlength="64" />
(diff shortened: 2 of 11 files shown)


CloudNiner:
I'll write some basic tests tomorrow am

schema_editor/app/scripts/views/recordtype/detail-add-edit-partial.html
```diff
@@ -0,0 +1,54 @@
+
```
kshepard: This looks like an almost exact copy of the add-edit-partial. Would it be worth it to try and combine them?

schema_editor/app/scripts/schemas/schemas-service.js
```diff
@@ -0,0 +1,27 @@
...
+    'use strict';
+
+    /* ngInject */
+    function Schemas() {
+        var module = {
+            Object: Object
+        };
+        return module;
+
+        function Object() {
```
kshepard: Doesn't 'Object' already exist in js? May be worth switching to something else.

schema_editor/app/scripts/schemas/schemas-service.js
```diff
@@ -0,0 +1,27 @@
...
+            return {
+                type: 'object',
+                title: '',
+                plural_title: '',
```
kshepard: Side note: I wonder if we should try and do something slick with DRF serializers to remove the need for underscores in our js.

schema_editor/app/scripts/schemas/schemas-service.js
kshepard: Yeah, looks like jshint is complaining about this and a couple other things in this file. We should probably have Travis run jshint.

kshepard:
+1. I can't think of an easy way to clean up the nested calls, though they're not too bad at the moment. I think when we add some good loading statuses/etc. it will be fine. I'm going to merge this now, since I'm currently working on sidebar integration and want to get everything here wired up.

schema_editor/app/scripts/views/recordtype/detail-add-edit-partial.html
CloudNiner: Yeah, I thought about that too. I could see these changing down the line to be more different, so I just left it as two separate views. But I don't feel strongly about it.

schema_editor/app/scripts/schemas/schemas-service.js
CloudNiner: Yeah, I'll change that name. Rookie mistake on my part.

schema_editor/app/scripts/schemas/schemas-service.js
CloudNiner: Yeah we could, theres a plugin out there and a few examples. Maybe as a separate task down the line? Not too worried about it now.

schema_editor/app/scripts/views/recordtype/detail-add-edit-partial.html
kshepard: Alright, let's just leave it for now. If time goes by and they don't diverge, and then there needs to be a structural rework for styling/etc., then we can combine them.

schema_editor/app/scripts/schemas/schemas-service.js
kshepard: Yeah, definitely a separate task. I'll add an issue.

PR #20, opened 2015-05-13
Feature/related content tests

CloudNiner:
Part 1 so that you can use the 'ase.mock.resources' module in your RecordType resources tests.


kshepard:
+1

PR #21, opened 2015-05-13
Add config mock

kshepard:
This should fix the Travis failures.


CloudNiner:
And this doesn't explode if config.js already exists elsewhere?

CloudNiner:
Also might be worth a comment in the ansible config.js template to remind devs to update this file as well when they add a key in the ansible template.

CloudNiner:
+1 on travis

kshepard:
Doesn't seem to explode with a config.js in place. Good idea on the ansible template note, I'll add it.

PR #22, opened 2015-05-13
Feature/json schema serializer

CloudNiner:
This gets us a working form on the schema edit page which will allow you to add new SelectList and TextField types and will persist user-entered data as the form schema is updated. 

On the TODO list:
  - Serialize the form values to our RecordType schema representation
  - Save the form schema and form values somewhere when the user clicks 'Save'
  - Reconstruct the form when the user navigates to this page after having saved the form
  - Add Image/Relationship form types


schema_editor/app/scripts/json-editor/json-editor-directive.js
```diff
@@ -0,0 +1,77 @@
...
+            var changeRef = null;
+
+            // Every time the editor options change, we need to destroy and recreate
+            // the form, preserving the user-entered data in the form
+            scope.$watch('options', function (newValue) {
+                if (!(newValue && newValue.schema)) {
+                    return;
+                }
+                var options = angular.extend({}, defaults, scope.options);
+                delete options.startval;
```
kshepard: I don't think this line does anything.

kshepard:
+1, looking good so far.

schema_editor/app/scripts/json-editor/json-editor-directive.js
CloudNiner: You're right, that's left over from when I was testing different ways to reload the user-entered values in the form when it gets destroyed + recreated. 

I'll remove and tweak this a bit, because as it stands now, the startval property on options will never be applied.

PR #23, opened 2015-05-14
Add basic tests for resources, recordtypes, and sidebar

kshepard:


schema_editor/test/mock/resources/resources-mocks.js
```diff
@@ -3,25 +3,70 @@
...
-                    'current_schema': '3ba9bc8f-7e4a-490e-8a14-666808711b10',
-                    'created': '2015-05-12T17:41:09.357521Z',
-                    'modified': '2015-05-12T17:41:09.357562Z',
-                    'label': 'Accident',
-                    'plural_label': 'Accidents',
-                    'description': 'An accident.',
-                    'active': true
+                    'uuid': '80c10057-2cfc-4a32-8e3c-0573e8bf8f3f',
+                    'schema': {
+                        // TODO: is this right? What are these blank properties for?
```
CloudNiner: Yeah this is ok for now. I was temporarily saving each of the related content types generated via the admin interface in the definitions block, although eventually they won't live there, but instead will need to be serialized/de-serialized from this schema object as was discussed yesterday. Not totally sure what this looks like yet.

schema_editor/test/spec/resources/recordschemas-service.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        RecordSchemas = _RecordSchemas_;
+        ResourcesMock = _ResourcesMock_;
+    }));
+
+    it('should extract record schemas from paginated response', function () {
+        var requestUrl = /\/api\/recordschemas/;
+        $httpBackend.whenGET(requestUrl).respond(ResourcesMock.RecordSchemaResponse);
+        RecordSchemas.query().$promise.then(function (data) {
+            expect(data.length).toBe(1);
+            var firearms = data[0].schema.definitions.Firearm;
```
CloudNiner: Some of these assertions would trivially break if we change the mock, without necessarily changing the substance of the test. 

Maybe just doing the data.length check, and then ensure that it has a relevant property or two at the root level, such as:
```
var recordSchema = data[0];
expect(recordSchema.schema).toEqual(jasmine.any(Object));
```

CloudNiner: Same below in recordtypes-service?

CloudNiner:
+1 good start

schema_editor/test/spec/resources/recordschemas-service.spec.js
kshepard: Sure, that's a good idea.

PR #24, opened 2015-05-19
Feature/image uploader option

CloudNiner:
Adds the two things we discussed:
- Additional dropdown for an Image
- Field options textarea for select list dropdown option


kshepard:
+1

PR #25, opened 2015-05-19
Fix form reinitialization from a destroyed editor instance

CloudNiner:
The editor var was one scope too high. If a json-editor
directive instance was destroyed, the editor object was not.

Moving the editor var inside the link function ensures
it gets destroyed with everything else when a json-editor
directive instance goes out of scope.


kshepard:
+1

PR #27, opened 2015-05-21
Rename detail to related

kshepard:


CloudNiner:
+1

kshepard:
Just snuck another rename commit into this PR: "event type" -> "record type". Everything should be nice and consistent now.

PR #28, opened 2015-05-22
Add multiple property for related content + more tests

kshepard:
This doesn't modify the schema to use an array vs. object; there is a separate task for that. This just allows setting a related type as multiple and having it reflected in the list view. Also fleshed out some tests a bit more.

### DIFF
--- a/schema_editor/test/spec/views/recordtype/related-add-controller.spec.js
+++ b/schema_editor/test/spec/views/recordtype/related-add-controller.spec.js
@@ -26,13 +26,12 @@ describe('ase.views.recordtype: RelatedAddController', function () {
     }));
 
     it('should add related content and switch view', function () {
-        /* jshint camelcase: false */
-        var recordType = {
-            label: 'Accident',
-            plural_label: 'Accidents',
-            description: 'An Accident'
-        };
+        var recordType = ResourcesMock.RecordType;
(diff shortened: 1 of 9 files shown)


schema_editor/test/spec/views/recordtype/related-add-controller.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    beforeEach(inject(function (_$controller_, _$httpBackend_, _$rootScope_,
+                                _ResourcesMock_) {
+        $controller = _$controller_;
+        $httpBackend = _$httpBackend_;
+        $rootScope = _$rootScope_;
+        $scope = $rootScope.$new();
+        ResourcesMock = _ResourcesMock_;
+    }));
+
+    it('should add related content and switch view', function () {
```
ddohler: This test (and some of the others below) could potentially be split into separate tests; it'll require a bit more boilerplate, but I think it's worth it to keep tests as granular as possible since it'll make debugging easier if they break.

ddohler:
+1

schema_editor/test/spec/views/recordtype/related-add-controller.spec.js
kshepard: That's a good idea. I'll work on splitting them out a bit.

PR #29, opened 2015-05-26
Notification Module

ddohler:
Creates a module for broadcasting global notifications, and updates the Geography page to use these notifications.

Currently, there is no notifications directive in the root partial because I had trouble with positioning and styling, so I added it into the Geography partial instead. However, we will likely want to create one so we don't have to put `<ase-notifications></ase-notifications>` every time we want to show notifications.

### DIFF
--- a/schema_editor/app/scripts/notifications/notifications-service.js
+++ b/schema_editor/app/scripts/notifications/notifications-service.js
@@ -52,9 +52,7 @@
             };
             var opts = angular.extend({}, defaults, options);
             active = opts;
-            timeoutId = $timeout(function () {
-                $rootScope.$broadcast('ase.notifications.show', opts);
-            });
+            $rootScope.$broadcast('ase.notifications.show', opts);
         }
     }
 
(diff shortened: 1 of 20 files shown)


schema_editor/app/scripts/notifications/notifications-service.js
```diff
@@ -0,0 +1,63 @@
...
+
+        /**
+         * Show a global notification, overridding the defaults defined in the function
+         * @param  {object} options Override the defaults with this config
+         * @return Broadcasts ase.notifications.show on $rootScope
+         */
+        function show(options) {
+            var defaults = {
+                timeout: 0,
+                top: '8rem',
```
kshepard: It doesn't look like this is actually being used, so it can probably be removed. If we do need something like this, I think it would be better handled in css anyway.

kshepard:
+1, pretty cool!

schema_editor/app/scripts/notifications/notifications-service.js
ddohler: Yeah that's a good point, I'll take this out.

PR #32, opened 2015-05-28
Serialize schema form

ddohler:
This sets up the Schema Entry Form for Related Content so that when the Save button is clicked, the data from the form is serialized into a JSON-Schema snippet designed to be inserted into the `definitions` of the larger JSON-Schema for the Record Type which contains the Related Content type.

Also adds some basic validation that field titles are not duplicated and improves the built-in form schema to validate the existence of field titles.

### DIFF
--- a/schema_editor/app/scripts/schemas/schemas-service.js
+++ b/schema_editor/app/scripts/schemas/schemas-service.js
@@ -23,16 +23,30 @@
         var module = {
             JsonObject: jsonObject,
             FieldTypes: {
-                'text': 'Text Field',
-                'selectlist': 'Select List',
-                'image': 'Image Uploader'
+                'text': {
+                    label: 'Text Field',
+                    jsonType: 'string',
+                    create: textField
+                },
+                'selectlist': {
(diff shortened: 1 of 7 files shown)


schema_editor/app/scripts/schemas/schemas-service.js
```diff
@@ -45,21 +61,109 @@
...
+         */
+        // TODO: This monolithic function isn't great; investigate more modular options
+        function _propertyFromSchemaFieldData(fieldData) {
+            var propertyDefinition = {};
+            propertyDefinition.type = module.FieldTypes[fieldData.fieldType].jsonType;
+
+            if (fieldData.fieldType === 'selectlist') {
+                propertyDefinition.enum = fieldData.fieldOptions;
+            }
+            if (fieldData.fieldType === 'image') {
```
kshepard: I suppose this can be an 'else if'. Though I agree that something more modular would be great -- I'll give that some thought, but I bet we can whip something up using prototypical inheritance.

kshepard:
+1

PR #36, opened 2015-05-29
Save related schema edits to db

kshepard:
Wanted to add tests for the save component, but there are no existing tests for this controller or directive -- will create an issue for that.

### DIFF
--- a/schema_editor/test/spec/schemas/schemas-service.spec.js
+++ b/schema_editor/test/spec/schemas/schemas-service.spec.js
@@ -98,6 +98,21 @@ describe('ase.schemas:Schemas', function () {
         expect(dataFormSchemaDef.required[0]).toEqual('Photo');
     });
 
+    it('should net set the "required" key when there are no required fields', function () {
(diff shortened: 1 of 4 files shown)


schema_editor/test/spec/schemas/schemas-service.spec.js
```diff
@@ -32,12 +32,35 @@ describe('ase.schemas:Schemas', function () {
...
+            isRequired: false,
+            isSearchable: true,
+            fieldTitle: 'Text',
+            textOptions: 'datetime'
+        }];
+        var dataFormSchemaDef = Schemas.definitionFromSchemaFormData(schemaFormData);
+        expect(dataFormSchemaDef.properties.Text.isSearchable)
+            .toEqual(schemaFormData[0].isSearchable);
+    });
+
```
ddohler: One more test would be good, to test that `required` is not included if `isRequired === false` for all fields.

ddohler:
+1

PR #38, opened 2015-05-29
Use array or object depending on 'multiple'

kshepard:
Working on a test for this, but it'll require some more work, since the current test for this controller is flawed in several ways, and doesn't actually test it correctly.


ddohler:
+1

PR #31, opened 2015-05-28
Use an array as the root form builder object

kshepard:
This is ready for review, but it's using feature/serialize-schema-form as a base. Once that branch is merged, I'll close this and open up a new PR targeting develop.

PR #40, opened 2015-06-02
Deserialize related type schemas to form

ddohler:
Deserializes items from the "definition" key of a schema into initial data for a Schema Entry Form. Also preserves field ordering via the "propertyOrder" attribute, and makes some incremental changes to the Schema Entry Form Schema.

@kshepard should take a look too.

### DIFF
--- a/schema_editor/app/scripts/schemas/schemas-service.js
+++ b/schema_editor/app/scripts/schemas/schemas-service.js
@@ -34,6 +34,7 @@
             addVersion4Declaration: addVersion4Declaration,
             validateSchemaFormData: validateSchemaFormData,
             definitionFromSchemaFormData: definitionFromSchemaFormData,
+            schemaFormDataFromDefinition: schemaFormDataFromDefinition,
             encodeJSONPointer: encodeJSONPointer
         };
         return module;
@@ -58,16 +59,19 @@
          * which would be used for data entry / editing. Does not perform validation.
          * @param {object} fieldData The value of one Schema Entry Form field, specifying required,
          *      searchable, possible values, etc.
+         * @param {number} index The index of this field in its form
          * @return {object} A snippet designed to be inserted into the 'properties' of a Data Form
          *      Schema
          */
         // TODO: This monolithic function isn't great; investigate more modular options
-        function _propertyFromSchemaFieldData(fieldData) {
+        // (likewise in the deserializing function below)
+        function _propertyFromSchemaFieldData(fieldData, index) {
             var propertyDefinition = {};
             propertyDefinition.type = module.FieldTypes[fieldData.fieldType].jsonType;
 
             if (fieldData.fieldType === 'selectlist') {
                 propertyDefinition.enum = fieldData.fieldOptions;
+                propertyDefinition.displayType = fieldData.displayType;
             } else if (fieldData.fieldType === 'image') {
                 propertyDefinition.media = {
                     binaryEncoding: 'base64',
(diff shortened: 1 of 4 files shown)


schema_editor/app/scripts/schemas/schemas-service.js
```diff
@@ -120,6 +136,59 @@
...
+                            break;
+                        case 'media': // Also not used
+                            break;
+                        default:
+                            fieldData[key] = value;
+                    }
+                });
+
+                // Handle required fields, which are stored outside the field info
+                if (_.findIndex(definition.required, function(item) {
```
CloudNiner: This is kinda difficult to read, I'm generally against relatively complicated ifs, although I'm not sure we have any notes about it in any of our js style guides.

Maybe something more along the lines of this:
```
var requiredIndex = _.findIndex(definition.required, function(item) { return item === title; });
fieldData.isRequired = requiredIndex >= 0;
```

schema_editor/test/spec/schemas/schemas-service.spec.js
```diff
@@ -148,4 +176,69 @@ describe('ase.schemas:Schemas', function () {
...
+            fieldType: 'text',
+            isSearchable: false,
+            isRequired: false,
+            fieldTitle: 'Text',
+            textOptions: 'datetime',
+            propertyOrder: 2
+        }];
+
+        // Transform back and forth a few times
+        var output = b(f(b(f(b(f(b(f(b(f(b(f(schemaFormData))))))))))));
```
CloudNiner: lolololol might be excessive by a few calls

schema_editor/test/spec/schemas/schemas-service.spec.js
```diff
@@ -148,4 +176,69 @@ describe('ase.schemas:Schemas', function () {
             message: jasmine.stringMatching('Diff')
         }]));
     });
+
+    // NOTE: This test must be updated when new field types are added.
+    it('should preserve all information when serializing and deserializing schemas', function () {
+        // In practice this means that the serialize and deserialize methods are inverses
```
CloudNiner: Would it be sensible to somehow test that all of the fieldTypes are represented below, and if not automatically fail the test?

CloudNiner:
+1 from me, I'm not able to check out the branch atm due to vm memory limitations, but I saw it working earlier.

schema_editor/test/spec/schemas/schemas-service.spec.js
ddohler: I was trying to think of a way to do that, but ran into the same problem identified here: https://github.com/WorldBank-Transport/DRIVER/pull/31#commitcomment-11407950 , which is that it's not easy right now to test things that are in `builder-schemas` because they're in separate files. I'd definitely like to add this test if we can find a way to do it.

schema_editor/app/scripts/schemas/schemas-service.js
ddohler: Yeah, that's a good call, will make that change.

schema_editor/test/spec/schemas/schemas-service.spec.js
ddohler: Yeah...I might have gotten a bit carried away with the nesting, but at least I'm _very_ confident they are inverses!

schema_editor/test/spec/schemas/schemas-service.spec.js
CloudNiner: Ok. I don't have any offhand ideas due to relative unfamiliarity with some of the newer stuff here. Maybe inspiration will strike once I dig in again.

kshepard:
LGTM too.

schema_editor/test/spec/schemas/schemas-service.spec.js
kshepard: @CloudNiner - Yeah, I'm really curious how to pull the json file into a test. I tried several different ways and was unsuccessful, but surely there's a way.

CloudNiner: Something like this post wouldn't work?

http://stackoverflow.com/questions/17370427/loading-a-mock-json-file-within-karmaangularjs-test

kshepard: I tried something similar to that, but I'll give it another shot this morning.

kshepard: Got it working based on some of those examples. Thanks for the pointer!

PR #42, opened 2015-06-03
Add recordtype form preview view

kshepard:


schema_editor/app/scripts/views/recordtype/preview-controller.js
```diff
@@ -0,0 +1,58 @@
...
+            return RecordSchemas.get({ id: currentSchemaId })
+                .$promise.then(function(recordSchema) {
+                    ctl.recordSchema = recordSchema;
+                });
+        }
+
+        // Called after all prerequesite data has been loaded
+        function onSchemaReady() {
+            ctl.editor = {
+                id: 'schema-editor',
```
ddohler: This should probably be different since this is a preview of a data entry form rather than a schema editing form.

ddohler:
+1, looks good, I tried this on my existing data and it worked great!

PR #45, opened 2015-07-10
Setup Angular frontend

ddohler:
This sets up a second Angular app to provide the "real" DRIVER frontend (as compared with the schema editor which is intended to be a super-admin-only tool). @CloudNiner did most of the real work.


web/Gruntfile.js
```diff
@@ -0,0 +1,497 @@
...
+
+    ngtemplates: {
+      dist: {
+        options: {
+          module: 'webApp',
+          htmlmin: '<%= htmlmin.dist.options %>',
+          usemin: 'scripts/scripts.js'
+        },
+        cwd: '<%= yeoman.app %>',
+        src: 'views/{,*/}*.html',
```
flibbertigibbet: Might want to make this less limited, to avoid infinite digest cycles like we saw in MUB.

CloudNiner: +1

CloudNiner: also change any reference to `views/{,*/}*.html` to `scripts/**/*.html`

web/Gruntfile.js
```diff
@@ -0,0 +1,497 @@
...
+          '<%= yeoman.dist %>/images',
+          '<%= yeoman.dist %>/styles'
+        ],
+        patterns: {
+          js: [[/(images\/[^''""]*\.(png|jpg|jpeg|gif|webp|svg))/g, 'Replacing references to images']]
+        }
+      }
+    },
+
+    // The following *-min tasks will produce minified files in the dist folder
```
CloudNiner: This commented out block can be deleted

kshepard:
+1, tested, looks good.

PR #47, opened 2015-07-14
Relationship field

ddohler:
Adds a relationship field that allows local references to another related content type within the same record.

Also creates a system-only `_localId` field that is added to all related types (except the initial default Details type since it doesn't really make sense there). The editor portion will need to auto-populate these fields with UUIDs.

I had to punt on auto-generating a human-readable name for the related content in the dropdown menu because it would have taken a while and I'm already a bit over time for this card, but I put in a to-do with reasonably detailed instructions, and I don't think it'll be too difficult to implement.

### DIFF
--- a/schema_editor/app/builder-schemas/related.json
+++ b/schema_editor/app/builder-schemas/related.json
@@ -8,7 +8,8 @@
         "oneOf": [
             { "$ref": "#/definitions/textField" },
             { "$ref": "#/definitions/selectList" },
-            { "$ref": "#/definitions/imageUploader" }
+            { "$ref": "#/definitions/imageUploader" },
+            { "$ref": "#/definitions/localReference" }
         ]
     },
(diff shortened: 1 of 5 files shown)


schema_editor/app/builder-schemas/related.json
```diff
@@ -150,6 +151,33 @@
...
+                "fieldTitle": {
+                    "title": "Reference Title",
+                    "type": "string",
+                    "minLength": 1
+                },
+                "referenceTarget": {
+                    "title": "Type of related info to reference",
+                    "type": "string",
+                    "format": "select",
+                    "enumSource": [["Choice one", "choice 2", "choice 3"]]
```
kshepard: Can probably get rid of these values?

kshepard:
+1, cool. I'll merge this since you're away for a few days.

PR #49, opened 2015-07-23
Fix problem with schema edits not saving

kshepard:
Also changed "Local Reference" to "Relationship"


ddohler:
+1, I've got this in my branch too but better to get the fix in quickly.

PR #50, opened 2015-07-24
Autodetect relation fields

ddohler:
This is working, but it broke two tests and needs at least one more test for the new functionality that got added, so there's a bit more work to do before it can be merged.
prsfielddetection

@CloudNiner may also want to take a look; I had an argument with JSON-Editor and I'm not sure I solved it the best way. https://github.com/WorldBank-Transport/DRIVER/pull/50/files#diff-880549e717f53f87000b28d600af6267R78

### DIFF
--- a/schema_editor/app/scripts/schemas/schemas-service.js
+++ b/schema_editor/app/scripts/schemas/schemas-service.js
@@ -61,9 +61,14 @@
                      * the reference dropdown can be auto-populated with meaningful information.
                      */
 
-                    toProperty: function(fieldData, index, allData, currentSchema, definitionName) {
+                    toProperty: function(fieldData, index, allData, currentSchema) {
                         var displayProperties = [];
                         // Switch depending on whether fieldData.referenceTarget == definitionName
+                        // Note that self-referential targets are disabled via a validation
+                        // function at the moment because they cause JSON-Editor to break. However,
+                        // I didn't realize that until after I wrote this code, so leaving it here
+                        // in a comment in case we need to re-enable this later.
+                        /*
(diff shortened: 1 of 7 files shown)


CloudNiner:
The custom validation stuff looks ok since its a limitation of the library. I took a cursory glance through the github issues for json-editor and didn't come up with any existing issues. Might be worth opening an issue there and detailing the problem.

ddohler:
I fixed the broken tests and added an extra check for the new functionality. This should be ready to go now assuming Travis is okay with it.

schema_editor/app/scripts/schemas/schemas-service.js
```diff
@@ -64,7 +96,7 @@
                             },
                             enumSource: [{
                                     source: 'target',
-                                    title: fieldData.referenceTarget + ' {{i}}',
+                                    title: displayProperties.join(' '),
```
kshepard: What do you think about a max length for this text? Maybe on a per-property basis too: in case some paragraph text (like a description), or a long street address is involved.

kshepard:
+1

schema_editor/app/scripts/schemas/schemas-service.js
ddohler: I think that would be good, but it would probably require pulling in a more powerful templating engine than the built-in one so that we have access to some more powerful features, because in this part of the code there's no visibility into what the eventual values might be. See the Swig example here: https://github.com/jdorn/json-editor#enum-values

kshepard: Oh, gotcha. This seems fine for now. We could probably even get by with just limiting the total length of the dropdown items using css.

PR #53, opened 2015-07-30
New event page

kshepard:


web/test/spec/views/home/home-directive.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+    it('should load directive', function () {
+        var requestUrl = /\/api\/recordtypes/;
+        $httpBackend.expectGET(requestUrl).respond(200, ResourcesMock.RecordTypeResponse);
+
+        var scope = $rootScope.$new();
+        var element = $compile('<driver-home></driver-home>')(scope);
+        $rootScope.$apply();
+
+        expect(element.find('.form-area-body').length).toEqual(1);
```
ddohler: Does there need to be a call to `$httpBackend.flush()` here?

web/test/spec/views/record/add-controller.spec.js
```diff
@@ -0,0 +1,71 @@
...
+            },
+            Person: [{
+                'abc': '123',
+                _localId: testUuid
+            }]
+        }, [], editor);
+
+        // Empty _localId fields should be filled in with uuids
+        expect(editorValue).toEqual(jasmine.any(Object));
+        expect(editorValue.Details._localId.length).toBe(36);
```
ddohler: There's a UUID-matching regex in `builder-schema.json` that could allow this test to be a bit stricter.

web/test/spec/views/record/add-directive.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+    it('should load directive', function () {
+        var requestUrl = /\/api\/recordtypes/;
+        $httpBackend.expectGET(requestUrl).respond(200, ResourcesMock.RecordTypeResponse);
+
+        var scope = $rootScope.$new();
+        var element = $compile('<driver-record-add></driver-record-add>')(scope);
+        $rootScope.$apply();
+
+        expect(element.find('json-editor').length).toEqual(1);
```
ddohler: Same question about `flush()` here although I think I may be missing something about when that call is and is not required.

ddohler:
+1

web/test/spec/views/record/add-controller.spec.js
kshepard: Oh, cool, good point.

web/test/spec/views/home/home-directive.spec.js
kshepard: You may be right. I'll take a look into it and give it a shot.

PR #57, opened 2015-08-03
Add record list view

kshepard:


web/app/scripts/views/record/list-controller.js
```diff
@@ -0,0 +1,58 @@
...
+                .$promise.then(function(records) {
+                    ctl.records = records.results;
+                });
+        }
+
+        function onRecordsLoaded() {
+            ctl.detailsProperty = ctl.recordType.label + ' Details';
+            ctl.propertiesKey = ctl.recordSchema.schema.definitions[ctl.detailsProperty].properties;
+            ctl.headerKeys = _.keys(ctl.propertiesKey)
+                .filter(function(key) {
```
CloudNiner: Could probably use `_.without` as a shortcut for `_.filter` here:
https://lodash.com/docs#without

web/test/spec/views/record/list-controller.spec.js
```diff
@@ -0,0 +1,66 @@
...
+        var recordSchemaId = recordSchema.uuid;
+        var recordSchemaIdUrl = new RegExp('api/recordschemas/' + recordSchemaId);
+        $httpBackend.expectGET(recordSchemaIdUrl).respond(200, recordSchema);
+
+        var recordResponse = DriverResourcesMock.RecordResponse;
+        var recordsByTypeUrl = new RegExp(recordTypeId);
+        $httpBackend.expectGET(recordsByTypeUrl).respond(200, recordResponse);
+        /*
+         TODO: can't figure out why this isn't working:
+         var recordsByTypeUrl = new RegExp('api/records/?record_type=' + recordTypeId);
```
CloudNiner: Huh, it looks ok and first glance, and I would need to play with it myself to fix it, but two things about that error message stand out.
1. The hostname is not on the 'expected' response
2. There is a trailing slash on the expected response
Based on the regex you're providing though, neither of those things should matter.

The only special characters in that regex are the slash and the question mark.

CloudNiner:
+1

web/test/spec/views/record/list-controller.spec.js
kshepard: Yeah, I have absolutely no idea. Tried a whole bunch of different things and nothing worked. Next time you're in this codebase it'd be cool if you could poke at it for a few minutes and see if anything comes to mind.

web/app/scripts/views/record/list-controller.js
kshepard: Neat, lodash has all the goodies. I'll try that out.

web/test/spec/views/record/list-controller.spec.js
CloudNiner: Sure thing.

web/app/scripts/views/record/list-controller.js
CloudNiner: Ha, yeah I only knew about it because I just used it for something else in a different project.

web/test/spec/views/record/list-controller.spec.js
kshepard: Oh, the solution just dawned on me: I was escaping it incorrectly. The question mark needs a double backslash before it, so:
`var recordsByTypeUrl = new RegExp('api/records/\\?record_type=' + recordTypeId);`

PR #59, opened 2015-08-05
Add navbar directive and skeleton layout

kshepard:


web/app/scripts/navbar/navbar-controller.js
```diff
@@ -0,0 +1,60 @@
...
+        ctl.updateState = updateState;
+
+        setRecordTypes();
+        setStates();
+        $rootScope.$on('$stateChangeSuccess', setStates);
+
+        // Sets the list of available record types, and also the active one if provided
+        function setRecordTypes() {
+            RecordTypes.query({ active: 'True' })
+                .$promise.then(function(recordTypes) {
```
ddohler: I think this could be written more concisely as `RecordTypes.query({ active: 'True'}, function(...) {})` but I'm not sure what our preferred style is; it's kind of nice making the promise semantics explicit.

web/app/scripts/navbar/navbar-controller.js
```diff
@@ -0,0 +1,60 @@
...
+                            ctl.selectedRecordType = recordTypes[0];
+                        }
+                    }
+                });
+        }
+
+        // Sets states that can be navigated to (exclude current state, since we're already there)
+        function setStates() {
+            ctl.selectedState = $state.current;
+            ctl.availableStates = _.chain(['home', 'dashboard', 'map', 'record.list','record.add'])
```
ddohler: I haven't tried this myself but it looks like there are some other functions on `$state` that could be used for figuring out the list of configured states and filtering out the current state: http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.$state

ddohler:
+1; take a look at the ui-router docs and see if there's any way to avoid hard-coding the list of allowed states into the navbar.

web/app/scripts/navbar/navbar-controller.js
kshepard: Alright, I'll take a look. We definitely don't want to grab all the configured states, because we only want to show some specific ones. A good approach may be to add a variable to the state config for 'showInNavbar' and use that for filtering.

web/app/scripts/navbar/navbar-controller.js
kshepard: Good point, I'll try that out: a more concise style is usually better.

web/app/scripts/navbar/navbar-controller.js
CloudNiner: +1 to a variable on the state config, or add a single `angular.constant()` with the array of states.

web/app/scripts/navbar/navbar-partial.html
```diff
@@ -0,0 +1,35 @@
...
+  </div>
+  <ul class="nav navbar-nav">
+    <li class="nav-dropdown">
+      <div class="btn-group" dropdown>
+        <button id="record-type" type="button" class="btn btn-primary" dropdown-toggle>
+          {{ ctl.selectedRecordType.plural_label }} <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu" role="menu" aria-labelledby="record-type">
+          <li class="nav-rt-item" role="menuitem" ng-repeat="recordType in ctl.recordTypes | orderBy:'label'">
+            <a ng-click="ctl.selectedRecordType = recordType; ctl.updateState()">
```
CloudNiner: Strongly consider moving this logic to a function on the controller. It's generally frowned upon to do multiple statements in the template because it is verbose and less clear. Bonus is that the logic in the function can be tested more easily.

kshepard: Cool, will do. I thought it looked a little ugly.

web/app/scripts/navbar/navbar-controller.js
kshepard: Looks like the variable on state config is going to work great. I didn't realize you can call $state.get() with no parameters to get all the configured states: pretty cool.

PR #60, opened 2015-08-06
Consolidate Docker volumes; relocate NPM and Bower directories

hectcastro:
The approach here is:

- Install NPM & Bower modules in separate directories (/npm and /bower)
- Create a symlinks for each project that points to those directories
- Symlinks are valid inside the container, but invalid outside

This allows us to mount the source directory inside the container without worrying about clobbering any existing NPM or Bower module installations.

Main downside is that adding or removing NPM or Bower modules requires a Docker container rebuild.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ python:
 
 cache:
   directories:
-    - schema_editor/node_modules
-    - schema_editor/bower_components
     - $HOME/.cache/pip
 
 addons:
@@ -23,6 +21,7 @@ before_install:
     #   https://github.com/npm/npm/issues/6309
   - nvm install 0.12
   - nvm use 0.12
+  - find . -type l -name "bower_components" -delete -o -name "node_modules" -delete
 
 install:
   - gem install -q sass compass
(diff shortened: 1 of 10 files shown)


ddohler:
+1 from me, this looks like an excellent refactor. Changing dependencies is exactly when I would expect to have to rebuild a container so that's not much of a downside for me.

kshepard:
+1, I had this same refactor in mind.

.travis.yml
```diff
@@ -4,8 +4,6 @@ python:
 
 cache:
   directories:
-    - schema_editor/node_modules
```
hectcastro: Any ideas for avoiding this, @ddohler @kshepard? Travis is trying to use its cache, which conflicts with the symlinks. I haven't been able to find anything that allows me to run commands before the cache gets applied to delete the symlinks.

.travis.yml
```diff
@@ -23,6 +21,7 @@ before_install:
     #   https://github.com/npm/npm/issues/6309
   - nvm install 0.12
   - nvm use 0.12
+  - find . -type l -name "bower_components" -delete -o -name "node_modules" -delete
```
hectcastro: This helps by getting rid of the symlinks before NPM and Bower try to do their thing.

.travis.yml
ddohler: That's certainly annoying but this seems like the only way to address it. What's the effect of not caching here?

kshepard: Sorry, I can't think of a better way either.

PR #61, opened 2015-08-06
Load accident data

ddohler:
Creates a script that will load in historical accident data; the data is available on the fileshare at /data/accidents_and_sites.csv.

I had to add a separate settings file for development; with the development server, `CONN_MAX_AGE: 3600` causes the database to run out of connection slots because the development server uses a new connection for each request but never tears them down.

### DIFF
--- a/deployment/ansible/roles/driver.app/templates/upstart-app.conf.j2
+++ b/deployment/ansible/roles/driver.app/templates/upstart-app.conf.j2
@@ -21,7 +21,7 @@ script
     {% if developing -%}
     --volume {{ root_app_dir }}:{{ root_app_dir }} \
     --volume /opt/ashlar:/opt/ashlar \
-    -e "DJANGO_SETTINGS_MODULE='driver.settings_dev'" \
+    --env "DJANGO_SETTINGS_MODULE=driver.settings_dev" \
     {% endif -%}
     {% for k,v in driver_conf.items() -%}
     --env {{ k }}={{ v }} \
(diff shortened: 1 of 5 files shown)


deployment/ansible/roles/driver.app/templates/upstart-app.conf.j2
```diff
@@ -21,6 +21,7 @@ script
     {% if developing -%}
     --volume {{ root_app_dir }}:{{ root_app_dir }} \
     --volume /opt/ashlar:/opt/ashlar \
+    -e "DJANGO_SETTINGS_MODULE='driver.settings_dev'" \
```
kshepard: This should probably be changed to "--env" to match the others in the file.

kshepard:
This looks good to me, but for some reason I'm having trouble while provisioning. I get this during the collectstatic step:
ImportError: No module named 'driver.settings_dev'

+1 when that problem is worked out.

ddohler:
Oops, I had already solved that problem but somehow the fix didn't make it into the PR. I pushed a fix and was able to provision successfully; let me know if it works for you.

kshepard:
Provisioning and running the script works for me now. :+1: 
There is currently a convention in the app (which possibly needs to be made more explicit), that the details property needs to be named "<title> Details" -- this assumption is made in a couple places. So in order for these records to work correctly as-is, either the details property name needs to be changed, or the record_type label needs to change.

ddohler:
I went ahead and changed the record type that it creates to "Accident" so that the details match up.

PR #62, opened 2015-08-06
Add next/previous buttons to list view

kshepard:


web/app/scripts/views/record/list-controller.js
```diff
@@ -33,12 +37,27 @@
...
             /* jshint camelcase: true */
+
+            if (offset) {
+                ctl.currentOffset += offset;
+                if (ctl.currentOffset) {
+                    params.offset = ctl.currentOffset;
+                }
+            }
+
+            return Records.get(params)
```
ddohler: It seems a bit strange to me that the Angular $resource is causing us to avoid using the API next/previous URLs here. I'm not sure there's anything we can do about it, but I wanted to bring it up, maybe there's a solution I don't know about.

ddohler:
+1

web/app/scripts/views/record/list-controller.js
kshepard: Yeah, it is an interesting conundrum: we have the next/previous URLs readily available, but it isn't obvious how to easily use them. I searched a bit, but haven't been able to find a good solution.

PR #64, opened 2015-08-07
Fix Driver application startup when developing

hectcastro:
If in development, wait for `vagrant-mounted` and `docker`. Otherwise, just wait for the `filesystem` and `docker.


kshepard:
+1

PR #65, opened 2015-08-07
Fix Docker volumes in Grunt wrapper script

hectcastro:
This should have been addressed in #60, but was missed. All `node_modules` are now part of the container image, so volume mounts of the source code directory will not clobber them.


kshepard:
+1

PR #66, opened 2015-08-10
Ensure that development version of Ashlar is used on restart

hectcastro:
A version of Ashlar is installed during the Docker image building process, but that version comes from the `develop` branch on GitHub. In order to support local development, an Upstart task definition runs after `driver-app` enters the `started` state to ensure that Ashlar is re-installed inside of the container on each service restart using the contents of `/opt/ashlar`.


kshepard:
Great, testing this out now.

kshepard:
+1

PR #67, opened 2015-08-11
Add geography and polygon choosers

kshepard:
State is currently propagated via stateParams in the URL. Once we have user account modules in place, it may make more sense to store these parameters there.


web/app/scripts/navbar/navbar-controller.js
```diff
@@ -2,45 +2,24 @@
     'use strict';
 
     /* ngInject */
-    function NavbarController($log, $rootScope, $state, $stateParams, RecordTypes) {
+    function NavbarController($log, $rootScope, $state, $stateParams,
```
ddohler: I wonder if there might be benefit in making a service to handle the responsibility for tracking the selected and available Geographies, Polygons, and RecordTypes. That would potentially simplify `setResourceObjects` because you would be able to remove the logic that sets the controller values via the `$stateParams` since a service would be less transient. It would also make things easier for other parts of the app that might need to respond to changes in the selected types, which several of them might.

ddohler:
Consider making a service to track Geographies / Polygons / RecordTypes. Looks good otherwise. +1

web/app/scripts/navbar/navbar-controller.js
kshepard: I definitely agree that some abstraction here would be beneficial. I had a couple different ideas in mind that would depend on how we'd like to interface with these values in other parts of the app. Since it's not trivial, I'm going to make an issue for it, so we can discuss the best approach and then implement it in the next Sprint.

PR #69, opened 2015-08-11
Add Leaflet wrapping directive

ddohler:
This creates a directive to wrap Leaflet; largely based off of @CloudNiner 's work here: https://github.com/azavea/mub-monitor/pull/27/files

There are a couple differences, however:
- I'm using an Attribute binding instead of an Element binding to avoid the use of `replace` which is apparently deprecated.
- I decided to forgo the DataService portion in favor of writing some tests.


web/app/index.html
```diff
@@ -39,6 +40,8 @@
     <script src="bower_components/angular-bootstrap/ui-bootstrap-tpls.js"></script>
     <script src="bower_components/lodash/lodash.js"></script>
     <script src="bower_components/angular-uuid/uuid.min.js"></script>
+    <script src="bower_components/leaflet/dist/leaflet.js"></script>
+    <script src="bower_components/leaflet/dist/leaflet-src.js"></script>
```
kshepard: It seems like only one of these scripts would be needed. Any reason for having both?

web/app/scripts/leaflet/leaflet-defaults-provider.js
```diff
@@ -0,0 +1,47 @@
...
+            center: [0,0],
+            zoom: 1,
+            crs: L.CRS.EPSG4326
+        };
+
+        /**
+         * Update defaults by merging user-set defaults into defaults object
+         * @param {LeafletDefaults} newDefaults
+         */
+        this.setDefaults = function (newDefaults) {
```
kshepard: I like the pattern we use in all the other modules, where `this` is set to a variable at the top of the function, and then the variable is used instead of `this`. It gets us away from needing to worry about the context-aware intricacies of `this`.

kshepard:
+1

web/app/index.html
ddohler: Hmm, good point -- this is just what happened when I did `bower install` but it's a bit strange to have both. I'll try taking out the minified version and see if that breaks anything.

CloudNiner: These scripts are added automatically by the grunt wiredep task. You'll need to add an entry to the wiredep task to exclude one of the files.

Example: https://github.com/azavea/lgbs-dtax-mapping/blob/develop/web/Gruntfile.js#L182

web/test/karma.conf.js
```diff
@@ -40,6 +40,8 @@ module.exports = function(config) {
       'bower_components/angular-bootstrap/ui-bootstrap-tpls.js',
       'bower_components/lodash/lodash.js',
       'bower_components/angular-uuid/uuid.min.js',
+      'bower_components/leaflet/dist/leaflet.js',
```
kshepard: Should remove this line as well. I'll take care of it in my current branch.

ddohler: Huh. I swear I had gotten rid of that. Thanks.

kshepard: Haha, I just had the same thing happen to me. The exclude was also needed on the 'test' section. I've added that and the line has stopped resurrecting itself.

PR #70, opened 2015-08-11
Feature/toddow directive

moradology:
untitled
(The data in this pic is accurate upon double check)


moradology:
Trying to figure these tests out...

moradology:
I know you're busy with other projects, @CloudNiner but I figured it might be good for you to take a look at what I've done directive-wise.

ddohler:
Just as a note, there was a problem with the times in the database dump that we got, so I manually set all the times to noon, that's why this looks funky.

schema_editor/bower.json
```diff
@@ -11,7 +11,7 @@
     "angular-spinkit": "^0.3.3",
     "angular-resource": "^1.3.0",
     "angular-ui-router": "~0.2.14",
-    "json-editor": "0.7.18",
+    "json-editor": "0.7.21",
```
kshepard: This needs to be kept at 0.7.18, since versions past that don't support spaces in template variable names (which we rely on). I have an open PR with json-editor to address this, but it hasn't been merged.

CloudNiner:
LGTM

web/app/scripts/toddow/toddow-directive.js
```diff
@@ -0,0 +1,164 @@
...
+                    return moment(time).format('e:H') + '-' + (+moment(time).format('H') + 1);
+                }
+
+                /**
+                 * Helper function to gather data together into a format more friendly to D3
+                 */
+                function formatData(events) {
+                    /* jshint camelcase: false */
+                    return d3.nest()
+                        .key(function(d) { return formatHourRange(d.occurred_from); })
```
ddohler: It'd be great if the field name could be another bound attribute that would allow setting the field to use as the date (i.e. `d[datetime_field]` instead of `d.occurred_from`). Then this could be used for arbitrary data structures as long as they contained datetime information at the root level.

web/app/scripts/toddow/toddow-directive.js
```diff
@@ -0,0 +1,164 @@
...
+                    width = 752;
+                var rect, color, svg, tooltip;  // GLOBAL
+                init();
+
+                /**
+                 * Watch for changes to chartData and redraw and redraw and redraw
+                 */
+                scope.$watch('chartData', function(val) {
+                    if (val) {
+                        color = d3.scale.linear()
```
ddohler: Is there a reason this needs to be in the `$watch` function? It doesn't look like it depends on `val` so maybe it could be moved into `init()`?

ddohler:
It looks like bower on Travis is complaining about not being able to find the correct d3 version; I'm not sure what that's about but it should be fixed before merging this. +1 otherwise.

web/app/scripts/toddow/toddow-directive.js
moradology: That's a great idea

web/app/scripts/toddow/toddow-directive.js
moradology: Ah, there appears to be some code I meant to change - this piece should be in the $watch block so that we can update the domain to match the dataset we're dealing with (0 to max rather than 0 to 4)

schema_editor/bower.json
kshepard: Looks like this wasn't reverted before merging.

web/app/scripts/views/home/home-partial.html
```diff
@@ -1,9 +1,28 @@
-<div class="form-area">
-    <div class="col-sm-8 col-sm-offset-3">
-        <div class="form-area-heading">
-            <h2>Home page placeholder</h2>
-        </div>
-        <div class="form-area-body">
+<div class="row">
```
kshepard: Looks like this partial was accidentally modified.

PR #77, opened 2015-08-13
State control broken out into services; broadcasted over rootscope

moradology:
This PR breaks state management out into a service and broadcasts changes over rootscope. Maybe of interest to @ddohler and perhaps could use an eye by @afink (I tried the slightly less common service pattern for this rather than factory, since ideally all views are sharing a single stateful thing and the semantics of angular.service lend themselves to this)


web/app/scripts/navbar/module.js
```diff
@@ -2,6 +2,9 @@
     'use strict';
 
     angular.module('driver.navbar', [
+        'driver.resources.recordstate',
+        'driver.resources.polygonstate',
+        'driver.resources.geographystate',
```
CloudNiner: It might be overkill to have all of these in separate modules. Maybe something like `driver.state` or `driver.storage` would be better?

web/app/scripts/resources/polygonstate/polygonstate-service.js
```diff
@@ -0,0 +1,66 @@
...
+        init();
+
+        /**
+         * initialization
+         */
+        function init() {
+          selected = null;
+          options = [];
+          defaultParams = {'active': true};
+          that.updateOptions();
```
CloudNiner: Should be able to just use `updateOptions()` here.

web/app/scripts/navbar/module.js
ddohler: I like this idea too; there seems to be a fair amount of duplication across the three modules the way it's currently structured.

web/app/scripts/resources/polygonstate/polygonstate-service.js
```diff
@@ -0,0 +1,66 @@
...
+ *  are broadcast to rootscope for use by controllers
+ */
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function PolygonState($window, $log, $rootScope, Polygons) {
+        var defaultParams, selected, options;
+        var _ = $window._;
+        this.updateOptions = updateOptions;
```
CloudNiner: For consistency with our typical angular style, consider using the format:
```
var module = {
updateOptions: updateOptions,
// ...
};
return module;
```
Instead of this + that.

kshepard:
Looks like a good start. I agree with @CloudNiner's suggestions, and will also add that the new services need some test cases. One thing this PR doesn't address that I think it should, is to modify the current views to use these services for obtaining the already-retrieved selected objects. For instance, when you're on, say, the dashboard, and then you switch to the list view, the current record type should already be available, so it shouldn't need to be re-requested from the API.

web/app/scripts/state/geographystate-service.js
```diff
@@ -0,0 +1,79 @@
...
+ * Geography state control - changes to the private vars which define this state
+ *  are broadcast to rootscope for use by controllers
+ */
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function GeographyState($window, $log, $rootScope, Geography) {
+        var defaultParams, selected, options;
+        var _ = $window._;
```
ddohler: Is this for `jshint` purposes? I think that can be solved by setting it in the jshint config file.

ddohler:
+1

PR #78, opened 2015-08-19
Make the details definition more explicit

### LINKED ISSUE #63: Make the details definition more explicit
The details object is currently added to the object with the convention `<title> Details`, and that convention is used in a couple places. It'd be better to have an explicit flag (or something of the sort) to make it less brittle (e.g. so it doesn't break if someone changes the title).

kshepard:
Closes #63 
Note: it didn't seem worth writing a data migration, so recreating record types will be necessary after this is merged.

### DIFF
--- a/schema_editor/app/scripts/views/recordtype/add-controller.js
+++ b/schema_editor/app/scripts/views/recordtype/add-controller.js
@@ -31,6 +31,7 @@
             var schema = Schemas.JsonObject();
             schema = Schemas.addVersion4Declaration(schema); // Make root object a "real" JSON-Schema
             var definition = Schemas.JsonObject();
+            definition.details = true; // Flag denoting that this is the 'details' definition
             definition.description = 'Details for ' + recordType.label;
             definition.multiple = false;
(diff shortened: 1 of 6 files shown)


schema_editor/app/scripts/views/recordtype/add-controller.js
```diff
@@ -31,6 +31,7 @@
             var schema = Schemas.JsonObject();
             schema = Schemas.addVersion4Declaration(schema); // Make root object a "real" JSON-Schema
             var definition = Schemas.JsonObject();
+            definition.details = true; // Flag denoting that this is the 'details' definition
```
ddohler: Did you confirm that this survives when the Details schema is edited using the schema builder?

ddohler:
+1

schema_editor/app/scripts/views/recordtype/add-controller.js
kshepard: Confirmed, works fine.

PR #80, opened 2015-08-21
Feature/windshaft docker

flibbertigibbet:
Initial Windshaft setup.

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -18,6 +18,9 @@ postgresql_host: "database.service.driver.internal"
 postgresql_port: 5432
 postgresql_version: "9.4"
 
+redis_bind_address: "redis.service.driver.internal"
+redis_port: 6379
+
 ## WEB SETTINGS
 web_js_html5mode: "true"
 web_js_html5mode_prefix: "!"
--- a/windshaft/Dockerfile
+++ b/windshaft/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:vivid
+
+MAINTAINER Azavea
+
+RUN apt-get update
+
+# Windshaft requires node < 11; vivid has 10.25 in nodejs package
+RUN apt-get -q install -y nodejs npm git-core redis-server \
+    libcairo2-dev libpango1.0-dev libjpeg8-dev libgif-dev libpq-dev
+
+# link nodejs command to node
+RUN update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
+
+RUN mkdir /windshaft
+WORKDIR /windshaft
+COPY server.js /windshaft/
+COPY healthCheck.js /windshaft/
+
+# NB: the most recent versions do not work with the provided configurations
+RUN npm install windshaft@0.36.0
+
+EXPOSE 5000
+
+CMD ["node", "server.js"]
--- a/windshaft/Dockerfile
+++ b/windshaft/Dockerfile
@@ -1,24 +1,31 @@
+# Windshaft dependencies are for Ubuntu
 FROM ubuntu:vivid
 
 MAINTAINER Azavea
 
-RUN apt-get update
-
-# Windshaft requires node < 11; vivid has 10.25 in nodejs package
-RUN apt-get -q install -y nodejs npm git-core redis-server \
-    libcairo2-dev libpango1.0-dev libjpeg8-dev libgif-dev libpq-dev
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    nodejs \
+    npm \
+    git-core \
+    libcairo2-dev \
+    libpango1.0-dev \
+    libjpeg8-dev \
+    libgif-dev \
+    libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
 
 # link nodejs command to node
 RUN update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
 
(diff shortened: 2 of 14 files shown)


deployment/ansible/group_vars/all
```diff
@@ -18,6 +18,9 @@ postgresql_host: "database.service.driver.internal"
 postgresql_port: 5432
 postgresql_version: "9.4"
 
+redis_bind_address: "database.service.driver.internal"
```
ddohler: It's not a big deal since this will likely get overridden in most cases anyway, but it'd probably be cleaner to have this be `cache.service...` or `redis.service...` since there's no guarantee the redis host will be the same as the database host.

windshaft/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:vivid
```
ddohler: I've read in a couple places that the ubuntu base images are heavier than they need to be and that using debian base images is generally preferable, but I don't know if we've come to a company consensus on this. @hectcastro ?

ddohler:
+1

windshaft/server.js
```diff
@@ -0,0 +1,57 @@
...
+        },
+        redis: {
+            host: redisHost,
+            port: redisPort
+        }
+    };
+
+// Initialize tile server
+var ws = new Windshaft.Server(config);
+ws.get('/health-check', healthCheck(config));
```
hectcastro: Nice.

windshaft/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+MAINTAINER Azavea
+
+RUN apt-get update
+
+# Windshaft requires node < 11; vivid has 10.25 in nodejs package
+RUN apt-get -q install -y nodejs npm git-core \
```
hectcastro: One of the Docker best practices is to combine similar steps into a single `RUN` instruction. If you switch to the `node` base image, these can probably be combined as:

```dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends \
    libcairo2-dev \
    libpango1.0-dev \
    libjpeg8-dev \
    libgif-dev \
    libpq-dev \
    rm -rf /var/lib/apt/lists/*
```

windshaft/Dockerfile
hectcastro: We have been trying to standardize on `debian:jessie` as the base image. This gives us many of the conveniences of Ubuntu, but with a overall lighter weight image.

That said, in this specific case, the majority of the container's purpose is to run Node, so we can try to reuse the `node:0.10-slim` container image as a base (this uses `debian:jessie` under the covers).

See also: https://github.com/azavea/docker-static-core/blob/7cba6b0eb3379068c33ff632508214c951d59266/Dockerfile

windshaft/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN mkdir -p /opt/windshaft
+WORKDIR /opt/windshaft
+COPY . /opt/windshaft
+
+# NB: the most recent versions do not work with the provided configurations
+RUN npm install windshaft@0.36.0
+
+EXPOSE 5000
+
+CMD ["node", "server.js"]
```
hectcastro: Not clear whether we expect to ever run anything besides `node server.js`, but if we do (maybe tests) we can make either `node` or `npm` (and use the `scripts` option in `package.json`) the `ENTRYPOINT`. That makes it so that arguments at the end of `docker run` get used as arguments for the `ENTRYPOINT` command.

When `ENTRYPOINT` exists, `CMD` becomes the arguments to `ENTRYPOINT`. So to keep the current approach as the default, it could be:

```dockerfile
ENTRYPOINT ["/usr/bin/node"]
CMD ["server.js"]
```

A potentially clearer example of this approach is [here](https://github.com/azavea/docker-aws-cli).

windshaft/Dockerfile
flibbertigibbet: Ok. I used ubuntu instead of node-slim here to ease installation of the dependency imaging libraries. Hopefully they're interchangeable then.

windshaft/Dockerfile
flibbertigibbet: Ok, will do. I'd doubt we'd want to bother with the Windshaft tests, as running them requires extra ImageMagick dependencies.

windshaft/Dockerfile
flibbertigibbet: Apparently the dependency library packages do not match those of Ubuntu (the Windshaft docs seem to assume Ubuntu.) How much do we care about making this work on Debian?

windshaft/Dockerfile
flibbertigibbet: Getting an error that it cannot stat `/usr/bin/node`.

windshaft/Dockerfile
flibbertigibbet: Fixed a changed dependency and added `build-essential` to get install to complete, but the resulting Windshaft build is broken and will not start. Going to stick with Ubuntu image.

ddohler: From my point of view it's probably not that important; I'd just leave a comment so that someone coming along later doesn't assume that we use Ubuntu for everything.

windshaft/Dockerfile
hectcastro: Can probably just leave it at `node` then.

flibbertigibbet: The entrypoint does work with the Ubuntu image.

PR #81, opened 2015-08-24
Use node:0.10-slim base image for Windshaft

hectcastro:
Make use of the `node:0.10-slim` base image to avoid having to install Node.js and NPM properly. Also, add image creation command to `Makefile` for faster rebuilding.

At some point it may make sense to break this image definition out into its own Azavea repository.

### DIFF
--- a/windshaft/Dockerfile
+++ b/windshaft/Dockerfile
@@ -1,31 +1,25 @@
-# Windshaft dependencies are for Ubuntu
-FROM ubuntu:vivid
+FROM node:0.10-slim
 
-MAINTAINER Azavea
+MAINTAINER Azavea <systems@azavea.com>
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
-    nodejs \
-    npm \
     git-core \
     libcairo2-dev \
     libpango1.0-dev \
-    libjpeg8-dev \
+    libjpeg62-turbo-dev \
     libgif-dev \
     libpq-dev \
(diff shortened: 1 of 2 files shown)


windshaft/Dockerfile
```diff
@@ -1,31 +1,25 @@
...
-# link nodejs command to node
-RUN update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
-
 RUN mkdir -p /opt/windshaft
 WORKDIR /opt/windshaft
 COPY . /opt/windshaft
 
 # NB: the most recent versions do not work with the provided configurations
-RUN npm install windshaft@0.36.0
+RUN npm install --unsafe-perm windshaft@0.36.0
```
flibbertigibbet: Any idea why this is necessary on Debian but not Ubuntu?

flibbertigibbet:
Does the windshaft service start successfully with this? Apart from `unsafe-perm`, I tried these same changes. The install would complete, but the build was bad, and windshaft would not start.

windshaft/Dockerfile
hectcastro: Unclear. My suspicion is that it's related to installing Node via Ubuntu's repositories. Given what `unsafe-perm` is supposed to prevent, seems like it should be happening with both base images.

hectcastro:
The test HTML file you supplied appears to work after a fresh provision:

screen shot 2015-08-24 at 15 31 24

flibbertigibbet:
Ok, neat. +1

PR #87, opened 2015-08-27
Feature/edit constant fields

flibbertigibbet:
Adds area to the head of the record edit form for the constant record fields: slug, label, geom, and occurred from/to.

image


flibbertigibbet:
We will probably want to use something else for a date/time picker. This is one of only two Bootstrap date pickers with a time components, which is why I'm using it here. (The other option is less suitable.)

The Angular adapter for it does not expose a way to dynamically update the options, so it's not currently possible to change the valid min/max date based on the other picker's value. The Angular adapter also does not recognize click events on the calendar icon, though the base library does. Also, @jbranigan does not like the UI of its time component (it has been used in another project).

web/app/scripts/views/record/add-edit-controller.js
```diff
@@ -99,8 +152,32 @@
...
+            }
+            /* jshint camelcase: true */
+
+            return true;
+        }
+
         function onSaveClicked() {
-            if (ctl.editor.errors.length > 0) {
+            // set geom array back on record
+            ctl.record.geom.coordinates = [ctl.geom.lat, ctl.geom.lng];
```
kshepard: This won't work when adding a new record, because the geom object won't yet be defined.

web/app/scripts/views/record/add-edit-controller.js
```diff
@@ -123,13 +200,12 @@
...
 
-                    // TODO: the following fields are external to the schema and need implementation
-                    // Generating bogus values for now -- to be revisited in a future task.
-                    slug: 'testslug', // Note: don't think we need a slug for a record, uuid is fine
-                    label: 'testlabel', // Note: label also seems unnecessarry for a record
-                    geom: 'POINT (0 0)', // TODO: we'll need a map with ability to search/drop point
-                    occurred_from: new Date(), // TODO: we'll need date pickers for occured from/to
-                    occurred_to: new Date()
+                    // constant fields
+                    slug: ctl.slug,
```
kshepard: I vote for removing the slug and the label from both this form and the backend (ashlar). If we do want to use a slug, it should be generated on the backend and would need to be unique: but I think the uuid works fine for this. I don't think the label will be useful in practice, since all of the information you'd want to put in a label will already be available in all of the other fields: it'll just make adding a new record more cumbersome.

web/app/scripts/views/record/add-edit-partial.html
```diff
@@ -5,6 +5,83 @@
...
+                                    <span class="input-group-addon">
+                                        <span class="glyphicon glyphicon-calendar"></span>
+                                    </span>
+                                </div>
+                            </div>
+                        </div>
+                        <div class="col-md-6 pull-right">
+                            <div class="form-group">
+                                <label class="control-label">occurred To</label>
+                                <div class="input-group" id="datetimepicker-To">
```
kshepard: The last I talked with @jbranigan, we wanted to not include occurred_to in the interface, as we're not planning on handling aoristic data in DRIVER. We do want to support it in ashlar though, so we'd only have a single datetime input and set both occurred_to and occurred_from to be equal. I say we either do this, or, if we decide we do want to support it, add a checkbox for "Occurred to is the same as Occurred from" and have it checked by default, since the vast majority of records will not have different from/to datetimes.

web/app/scripts/views/record/embed-map-directive.js
```diff
@@ -0,0 +1,70 @@
+(function () {
+    'use strict';
+
+    var stamenTonerAttribution = 'Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.';
+
+    function driverEmbedMap() {
```
kshepard: This directive could use a test.

kshepard:
+1 after addressing my comments. Looks good.

flibbertigibbet:
The task this addresses is for adding the interface to edit the existing constant fields, not modifying what those are. Perhaps it would make more sense to address those changes here and to Ashlar separately.

kshepard:
Sure, that works. I'll add a couple issues.

web/app/scripts/views/record/add-edit-controller.js
flibbertigibbet: Good call.

PR #90, opened 2015-08-27
Feature/map geo boundaries

flibbertigibbet:
Refactors Windshaft parameter-building to separate file, and adds option to get uploaded bounding polygon. Displays boundary polygon on map page.

Due to azavea/ashlar#44, bounding polygon may not be in correct projection.

### DIFF
--- a/windshaft/server.js
+++ b/windshaft/server.js
@@ -11,11 +11,17 @@ var redisPort = process.env.DRIVER_REDIS_PORT;
 // RFC4122. See: http://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid
 var uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
-var baseQuery = ["(SELECT * FROM ashlar_record WHERE schema_id = ",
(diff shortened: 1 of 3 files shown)


windshaft/server.js
```diff
@@ -8,45 +9,17 @@ var dbPassword = process.env.DRIVER_DB_PASSWORD;
...
-
 var config = {
         useProfiler: true,
-        base_url: '/tiles/recordtype/:recordtype',
-        base_url_notable: '/tiles/recordtype/:recordtype',
+        // :recordtype parameter can be:
+        // GEO to get the user-uploaded boundary polygon;
+        // ALL to get record points of all record types; or,
+        // a UUID of a particular record type (returns record points of that type)
+        base_url: '/tiles/:recordtype',
```
ddohler: Now that we're serving more than just Record dots, it might be good to generalize this a bit, perhaps by doing something like `/tiles/:type/:id` and then accessing the boundaries by e.g. `/tiles/boundary/ALL` and likewise for records.

ddohler:
+1, consider generalizing the URL structure.

windshaft/server.js
flibbertigibbet: Ok. I'd considered using `type` for the parameter name, but decided not to, as it is a reserved keyword in JavaScript. I'll figure something else out that's more general.

PR #92, opened 2015-08-28
Feature/event record interactivity

flibbertigibbet:
image

### DIFF
--- a/web/app/scripts/views/map/layers-directive.js
+++ b/web/app/scripts/views/map/layers-directive.js
@@ -34,20 +34,19 @@
                                                     {useJsonP: false});
             map.addLayer(utfGridRecordsLayer, {detectRetina: true});
 
-            utfGridRecordsLayer.on('mouseover', function(e) {
+            utfGridRecordsLayer.on('click', function(e) {
+                var popupOptions = {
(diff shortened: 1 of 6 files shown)


web/app/scripts/views/map/layers-directive.js
```diff
@@ -28,12 +28,69 @@
...
             map.addLayer(boundaryLayer, {detectRetina: true});
         }
+
+        /**
+         * Build popup content from arbitrary record data.
+         *
+         * @param {Object} UTFGrid interactivity data from interaction event object
+         * @returns {String} HTML snippet for a Leaflet popup.
+         */
+        function buildRecordPopup(record) {
```
ddohler: This function could use a simple test passing in an object and checking to ensure that the correct HTML string is generated.

ddohler:
+1

web/app/scripts/views/map/layers-directive.js
flibbertigibbet: There's not really a way to test this directly that I know of, as it is a private function, but I'll see if I can find a way to test it indirectly by inspecting the generated element.

PR #94, opened 2015-08-31
Initial, non-interactive stepwise graph directive created

moradology:
zz015-08-31 at 9 59 13 am

### DIFF
--- a/web/app/scripts/stepwise/stepwise-directive.js
+++ b/web/app/scripts/stepwise/stepwise-directive.js
@@ -8,7 +8,7 @@
     'use strict';
 
     /* ngInject */
-    function Stepwise($window) {
+    function Stepwise() {
         var module = {
             restrict: 'A',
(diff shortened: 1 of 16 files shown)


web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -0,0 +1,182 @@
...
+    function Stepwise($window) {
+        var module = {
+            restrict: 'A',
+            scope: {
+              chartData: '='
+            },
+            template: '<svg></svg>',
+            link: function(scope, elem) {
+                var d3 = $window.d3;
+                var _ = $window._;
```
ddohler: Let's move these into the .jshintrc to minimize non-functional code.

ddohler:
It looks like the Y-axis positioning is a little bit off; it's overlapping the `0` at the bottom.

+1

PR #97, opened 2015-08-31
Increase max upload size to 10MB

ddohler:
This is a more reasonable size for most shapefiles. This allows importing the Philippines administrative regions shapefile.


flibbertigibbet:
+1

PR #100, opened 2015-09-01
Tabbed details view

kshepard:
Will add remaining tests for controllers/directives before merging.


moradology:
Looks like this is maybe a good idea for the textarea `<pre>` elements: https://css-tricks.com/snippets/css/make-pre-text-wrap/

moradology:
This looks really good; +1 after trying to get that text to wrap

kshepard:
Cool, good call on the text wrap. Didn't have any text long enough in my tests to notice it didn't wrap.

moradology:
Yeah, and `pre` is a really obscure tag

PR #102, opened 2015-09-01
Filterbar stub/numeric

moradology:


web/app/scripts/filterbar/numeric-range-directive.js
```diff
@@ -0,0 +1,66 @@
...
+                function init() {
+                    scope.filter = {};
+                    scope.error = {};
+                }
+
+                /**
+                 * A simple wrapper around driver-filterbar's updateFilter function;
+                 *  filters should only be updated when data validates
+                 *
+                 * @param filterLabel {string} label of which field to filter
```
kshepard: These param names don't match up with the comment.

web/app/scripts/filterbar/numeric-range-directive.js
```diff
@@ -0,0 +1,66 @@
...
+                 * @param filterObj {object} filter data
+                 */
+                scope.updateFilter = function(fLabel, fObj) {
+                    if (scope.isMinMaxValid()) {
+                        filterCtl.updateFilter(fLabel, fObj);
+                    }
+                };
+
+                $('div.dropdown.numrange ul.dropdown-menu').on('click', function (event) {
```
kshepard: I wonder if it'd be cleaner to just put this in an ng-click on the template.

web/app/scripts/filterbar/numeric-range-directive.js
```diff
@@ -0,0 +1,66 @@
...
+                    event.stopPropagation();
+                });
+
+                /**
+                 * When called, evaluate filter.min and filter.max to ensure they're valid
+                 */
+                scope.isMinMaxValid = function() {
+                    if (typeof scope.filter.min === 'number' &&  typeof scope.filter.max === 'number') {
```
kshepard: Super minor: but there's an extra space after the &&.

web/app/scripts/filterbar/filterbar-controller.js
```diff
@@ -0,0 +1,57 @@
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function FilterbarController($log, $window, $rootScope, $scope, RecordSchemas) {
```
kshepard: There should be a test added for this controller.

web/test/spec/filterbar/numeric-range-directive.spec.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+describe('driver.filterbar: Numeric Range', function () {
```
kshepard: I'd like to see a test here that checks the validation logic. Also, I noticed this directive doesn't have an associated controller: it may be easier to test this if the controller is broken out.

kshepard:
+1 after adding a couple more functional tests. Cool!

web/app/scripts/filterbar/numeric-range-directive.js
moradology: I had no idea about the '$event' object that angular creates for ng-click - ng-click is way better than setting up (potentially redundant) listeners

web/test/spec/filterbar/numeric-range-directive.spec.js
moradology: Breaking this controller out will be more of a headache than it is worth since I'm requiring filterCtl from parent as the shared controller. I can probably test this logic with some jquery selections instead

Issue #84, opened 2015-08-25
Web app does not run on root without manual start


flibbertigibbet:
After provisioning, the editor app is running, but the web app does not run without executing the `grunt.sh` script from the host to serve it.

ddohler:
I did some digging with some help from @CloudNiner and this is a two-part problem:
- nginx is returning `/index.html` when it should be returning a 404 (Looks like the fix here is for the `try_files` stanza to end with `=404;` rather than `index.html`).
- The minification process isn't handling the view partials correctly, so nginx falls back to returning `index.html`, which tries to load the view partials again, ad infinitum.

PR #103, opened 2015-09-01
Feature/fix minification

flibbertigibbet:
Fixes grunt production task, so web app will run in production from `build` task.
Closes #55.

Although this makes partials available again, they are not minified or concatenated. It might be better to reinstate the `ngtemplates` task removed in #55 to serve the partials more efficiently.

Also fixes issues with Leaflet image references.

### DIFF
--- a/deployment/ansible/roles/driver.app/templates/nginx-app.conf.j2
+++ b/deployment/ansible/roles/driver.app/templates/nginx-app.conf.j2
@@ -13,7 +13,7 @@ server {
         proxy_set_header X-Forwarded-Server $host;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
-        try_files $uri $uri/ /index.html;
+        try_files $uri $uri/ =404;
     }
 
     location /editor/ {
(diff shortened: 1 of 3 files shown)


deployment/ansible/roles/driver.app/templates/nginx-app.conf.j2
```diff
@@ -13,7 +13,7 @@ server {
         proxy_set_header X-Forwarded-Server $host;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
-        try_files $uri $uri/ /index.html;
```
ddohler: Do all the app pages still work with this change? After replying to #84 I was told that a fallback to `/index.html` is required when `html5mode = true` in order to allow Angular to handle routing paths that nginx can't find. I'd love it if that weren't true though since I think that recursively redirecting forever is a poor substitute for 404s.

ddohler:
+1 -- should probably also open an issue for minifying / concatenating partials.

deployment/ansible/roles/driver.app/templates/nginx-app.conf.j2
flibbertigibbet: I've tried clicking around everything in the web app, and it seems fine. Not sure if there's maybe some other browsers/configurations that might not work with this.

flibbertigibbet: It it curious that `/#!/` gets inserted in all the routes.

PR #109, opened 2015-09-03
Move comment to fix travis

moradology:


kshepard:
+1 (assuming it passes).

PR #116, opened 2015-09-04
Feature/embed map controller

flibbertigibbet:
Refactor embedded map directive to have a controller, and make it configurable to be editable or not.

### DIFF
--- a/web/app/scripts/views/record/add-edit-partial.html
+++ b/web/app/scripts/views/record/add-edit-partial.html
@@ -42,7 +42,7 @@
                             </div>
                         </div>
                         <div class="col-md-12">
-                            <div class="map" leaflet-map driver-embed-map style="height:300px;"></div>
+                            <div class="map" leaflet-map driver-embed-map editable="true"></div>
                         </div>
                     </div>
                     <div class="row">
--- a/web/app/index.html
+++ b/web/app/index.html
@@ -132,6 +132,7 @@
     <script src="scripts/toddow/module.js"></script>
     <script src="scripts/toddow/toddow-directive.js"></script>
     <script src="scripts/views/record/module.js"></script>
+    <script src="scripts/views/record/embed-map-controller.js"></script>
     <script src="scripts/views/record/embed-map-directive.js"></script>
     <script src="scripts/views/record/add-edit-controller.js"></script>
     <script src="scripts/views/record/add-edit-directive.js"></script>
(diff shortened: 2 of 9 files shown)


web/app/scripts/views/record/add-edit-partial.html
```diff
@@ -42,7 +42,7 @@
                             </div>
                         </div>
                         <div class="col-md-12">
-                            <div class="map" leaflet-map driver-embed-map style="height:300px;"></div>
+                            <div class="map" leaflet-map driver-embed-map editable="true"></div>
```
kshepard: Would it be hard to add attributes for latitude and longitude? It would make it super easy to add an embedded map without needing to fire an event to update the coords.

kshepard:
+1

web/app/scripts/views/record/add-edit-partial.html
flibbertigibbet: If you have the initial coordinates before the map loads, that might be a great idea.  But I imagine there might be a timing issue with the record and map both loading asynchronously, and the record coordinates not necessarily populating before the map gets set up.

For the editable map, it will still need the events. Because there are multiple directives involved, I don't think this directive can go put/read things on the scope.

kshepard: Yeah, I was mostly thinking of it from the perspective of the embedded map I'm adding on the details view. There I already have the latitude and longitude on the controller at the time this directive is rendered, so it'd be great to be able to just pass them in.

kshepard: Also, it is possible to use something like: ng-if="ctl.record" to make sure the directive doesn't render before the data is available. Not suggesting you should change that here, just letting you know since I've found it useful.

flibbertigibbet: I think it might be a slightly better experience for the map to go ahead and load, then animate the pan/zoom to the marker when it comes in, rather than have the page height suddenly expand to show the fully loaded map after other things have rendered.

kshepard: Yeah, that definitely makes sense for your use case. Mine is a bit different since the details tabs are being rendered dynamically based on record content and the map data is already available at that time. I think it's fine as-is though, we can add that ability later on if we want.

flibbertigibbet:
Okay, added `lat` and `lng` attributes.

kshepard:
Awesome, thanks!

PR #117, opened 2015-09-04
Add record constants to detail tabbed view

kshepard:
@flibbertigibbet: for some reason the map is behaving a little weird here. See below. It has a bunch of gray area and the marker is loaded off to the upper-left. When I drag the map down-right it comes into view. Was wondering if you've seen this before or have any insight into how to fix it.

details_constants

### DIFF
--- a/web/app/scripts/details/details-constants-partial.html
+++ b/web/app/scripts/details/details-constants-partial.html
@@ -1,0 +1,38 @@
+<div class="col-sm-6">
+    <label>
+        Occurred
+    </label>
+    <div class="value constant date occurred">
+        {{ ::ctl.record.occurred_from | date : ctl.dateFormat }}
+    </div>
+</div>
+<div class="col-sm-6">
+    <label>
+        Created
+    </label>
+    <div class="value constant date created">
+        {{ ::ctl.record.created | date: ctl.dateFormat }}
+    </div>
+</div>
+<div class="col-sm-6">
+    <label>
+        Latitude
+    </label>
+    <div class="value constant float latitude">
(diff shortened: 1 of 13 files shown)


flibbertigibbet:
That's a new one to me. Maybe try disabling animation on `setView` for the marker?

kshepard:
Alright, I'll try a few things. I think it may just a styling problem. Even before I added the marker it was behaving like this.

flibbertigibbet:
I did have to explicitly set the height when I added it to the records view, though in that case the map wasn't visible at all.

web/app/scripts/details/details-constants-partial.html
```diff
@@ -0,0 +1,38 @@
...
+    <label>
+        Longitude
+    </label>
+    <div class="value constant float longitude">
+        {{ ::ctl.record.geom.coordinates[0] }}
+    </div>
+</div>
+<div class="col-md-12">
+    <div class="map" leaflet-map driver-embed-map
+        lat="{{ ::ctl.record.geom.coordinates[1] }}"
```
flibbertigibbet: These should probably be sent as numbers and not strings.

flibbertigibbet:
+1 once the map view is sorted out

web/app/scripts/details/details-constants-partial.html
kshepard: It doesn't appear to be possible to have them passed as numbers. I tried removing the quotes and removing the colons, but it always comes across as a string. It does still work though as-is, so probably not a problem. If you know of an alternate way, let me know.

PR #118, opened 2015-09-04
Feature/map updates

flibbertigibbet:
* Refactor map layer directive logic into controller
* Add styling for events and boundaries layers
* Add event interactivity to the layer switcher
* Only show events and interactivity by default

Will add some tests.


flibbertigibbet:
Events + boundaries:
image

Heatmap shows on top of events now that layer zIndex is set explicitly.

kshepard:
+1, nice refactor.

jbranigan:
Yay

flibbertigibbet:
Can't get tests to run for the layers controller, seems to be related to UTFGrid. Merging.

PR #119, opened 2015-09-04
Feature/fix urls - WIP

moradology:
A bit of a misnomer here - included is also some styling for John's demo of interest, perhaps, to @flibbertigibbet as well


moradology:
Ah, yes. I usually use 'asdf' before I squash it - must have forgotten somehow.

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -75,7 +75,7 @@
 
                     svg.append('g')
                       .attr('class', 'x axis')
-                      .call(xAxis)
+                      //.call(xAxis)
```
kshepard: Remove

web/app/scripts/views/dashboard/dashboard-partial.html
```diff
@@ -1,13 +1,136 @@
...
-            <div class="form-area-heading">
-                <h2>Dashboard page placeholder</h2>
-            </div>
-            <div driver-stepwise chart-data="ctl.records.results"></div>
-            <div driver-toddow chart-data="ctl.records.results"></div>
-            <div class="form-area-body">
-            </div>
-        </div>
-    </div>
+<main class="dashboard">
```
kshepard: This file can use some formatting.

moradology: This HTML is purely for the demo on tues; i'm not sure how important the text you see here actually is for readability/reuse/maintenance

kshepard: I imagine we'll probably end up using some of it afterwards though when fleshing out the dashboard. I wasn't thinking anything excessive, just having your editor auto-format the tags.

kshepard:
I took this branch for a spin and there are a few things that aren't working correctly. Switching record types doesn't seem to do anything: for example, when you're on the list view and change the record type, the new set of records aren't pulled down. Also, nothing is being persisted on page refresh -- it's always choosing the default options in the navbar dropdowns. If you're on the list view and change the the record type and geometry and then refresh the page, it should maintain those selected values and also apply them to the page. Also, the dashboard styling looks pretty weird: there are a lot of squashed-together elements.

moradology:
I'll be adding commits to this tomorrow addressing the tests and missing functionality. This should have been marked WIP.

moradology:
Functionality is nicely in place and I'm currently working on testing - angular-local-storage introduces some new difficulties in testing that I think I'm going to have to use jasmine's spy for.

web/app/scripts/state/recordstate-service.js
```diff
@@ -20,6 +27,8 @@
          */
         function init() {
           selected = null;
+          gettingSelected = false;
+          gettingOptions = false;
```
kshepard: Just curious: why are these variables needed in this service, but not the other services?

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -10,21 +10,31 @@
...
-                var margin = {top: 20, right: 20, bottom: 30, left: 50},
-                    width = 752 - margin.left - margin.right,
-                    height = 475 - margin.top - margin.bottom;
+            link: function(scope, elem, attrs, elemStat) {
+                scope.$watchGroup([
+                    function() { return elemStat.height; },
+                    function() { return elemStat.width; }],
+                    function(oldVal, newVal) {
+                        console.log('old', oldVal, 'new', newVal);
+                        console.log(elemStat.height, elemStat.width);
```
kshepard: These can either be removed, or if needed, switched to use $log.

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -52,52 +62,64 @@
...
                         .scale(y)
                         .orient('left')
-                        .ticks(Math.min(10, _.max(_.map(data, function(d) { return d.count; }))))
                         .tickSize(1);
 
                     svg.append('g')
-                      .attr('class', 'x axis')
-                      .call(xAxis)
+                      .attr('class', 'xAxis')
+                      //.call(xAxis)  // TODO: uncomment after worldbank demo
```
kshepard: What happens when we eventually uncomment this? Does it break anything in particular?

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -52,52 +62,64 @@
...
+                 * @param data {array} the data to change our chart with
+                 */
+                function updateChart(data) {
+                    y.domain([0, d3.max(data, function(d) { return d.count; })]);
+                    yAxis.scale(y)
+                        .ticks(Math.min(10, _.max(_.map(data, function(d) { return d.count; }))));
+                    xAxis.scale(x)
+                        .tickFormat(xAxisTextFormat);
+
+                    svg.select('.xAxis').transition().call(xAxis);  // TODO: uncomment
```
kshepard: Should this be commented, or the comment removed?

kshepard:
This is looking good. The only weird thing I noticed is that clicking on the state dropdowns and not making a selection changes the url to the root state. It'd be good to fix that if possible. +1 after fixing linting/tests.

web/app/scripts/state/recordstate-service.js
moradology: They should definitely be extended to the other scripts - the main purpose is just to reduce the number of outgoing requests.

PR #122, opened 2015-09-09
Ensure Windshaft starts on boot

### LINKED ISSUE #115: Windshaft service on vagrant requires manual start
after `vagrant up` with existing VMs.

hectcastro:
Now, when configuration files associated with Windshaft change, the Windshaft service is restarted.

In addition to this fix, it was discovered that there is a Vagrant regression in emitting the `vagrant-mounted` Upstart event. To work around that, the `pre-start` stanza of both the Windshaft and Driver applications include a wait for their respective source code mount points (in development).

Attempts to resolve #115.

See also: https://github.com/mitchellh/vagrant/issues/6074


kshepard:
+1

PR #128, opened 2015-09-09
Feature/remove occurred to

flibbertigibbet:
Remove `occurred_to` from record edit interface. Set field to `occurred_from` on save.


flibbertigibbet:
Closing for #129, due to errant commits showing up here.

PR #130, opened 2015-09-10
Minify Angular templates

flibbertigibbet:
Minify Angular templates and concatenate to `scripts.js`. Also adds back `cdnify` task.


kshepard:
+1

PR #131, opened 2015-09-10
Remove slug and label from record editor.

### LINKED ISSUE #88: Remove record slug and label in both DRIVER and ashlar
See: https://github.com/WorldBank-Transport/DRIVER/pull/87#discussion-diff-38119935

flibbertigibbet:
Closes #88.

Will make companion PR to Ashlar.


kshepard:
I think there are other places that the label/slug need to be removed from. Off the top of my head: the map popups, the pnp data loading scripts, and the data mocks. +1 after those are removed.

Issue #133, opened 2015-09-10
Wire up record type dropdown to filter via Windshaft query


flibbertigibbet:
The Windshaft config currently accepts an argument for the record type UUID. It's not yet wired up to respond to the record type dropdown.

flibbertigibbet:
Closed in #137.

PR #137, opened 2015-09-10
Filter record types on map using dropdown

flibbertigibbet:


flibbertigibbet:
Closes #133.

windshaft/driver.js
```diff
@@ -10,10 +10,11 @@ var uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-
 // NB: interactivity only works with string columns
 var baseRecordQuery = ["(SELECT geom, uuid::varchar, data::varchar, occurred_from::varchar, ",
                        "occurred_to::varchar ",
-                       "FROM ashlar_record"
+                       "FROM ashlar_record WHERE schema_id IN ",
+                       "(SELECT uuid FROM ashlar_recordschema "
                        ].join("");
-var filterRecordQuery = " AND record_type_id = '";
-var endRecordQuery = ") AS ashlar_record";
+var filterRecordQuery = " WHERE record_type_id = '";
```
kshepard: Might as well also add in a check for active being true in here, since we don't want to display inactive record types. This would prevent those old records in staging from being displayed.

kshepard:
+1

windshaft/driver.js
flibbertigibbet: If the record type is inactive, then it shouldn't show as an option in the UI, right? I'm not seeing an `active` column on the schema.

kshepard: That's correct. There's an active column on the record type. I'm just referring to the 'ALL' scenario where every record type is pulled: those should only be the active record types.

flibbertigibbet: I suppose that might make sense for when `ALL` is selected. There's no way to see `ALL` in the UI currently, since there's not an option for that in the dropdown.

PR #138, opened 2015-09-11
Closes #110: Add CSRF header to angular apps

### LINKED ISSUE #110: CSRF token not set properly for production grunt build
Attempting to save a record on vagrant-hosted port 7000 returns 403 with error:
```
CSRF Failed: CSRF token missing or incorrect.
```

Saving records works when hosting via grunt on port 7002.

sharph:
`X-CSRFToken` header is now sent from angular apps on port 7000 when accessing django.

@kshepard The reason we didn't see the angular apps served on 7001 and 7002 sending `X-CSRFToken` is because they were requesting resources from port 7000; effectively across domains.

### DIFF
--- a/schema_editor/app/scripts/app.js
+++ b/schema_editor/app/scripts/app.js
@@ -14,6 +14,12 @@
         $logProvider.debugEnabled(Config.debug);
     }
 
+    /* ngInject */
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
+1

PR #141, opened 2015-09-14
Feature/namespace configs

### LINKED ISSUE #83: Schema editor config.js overrides web app config.js
Modifying the web app `config.js` template has no effect without making a matching edit to the schema editor `config.js`. On provisioning, the web app `config.js` template file does get copied to the correct directory (`/opt/web/`) and is correct in the web docker container.

Apparently both `config.js`files load on a request to the web app, the editor config loading second.

flibbertigibbet:
Closes #83.

Namespace the two `Config` Angular constant objects loaded by the web app (which also loads the schema editor's resources).

### DIFF
--- a/deployment/ansible/roles/driver.web/templates/editor-config-js.j2
+++ b/deployment/ansible/roles/driver.web/templates/editor-config-js.j2
@@ -6,6 +6,9 @@
     //
     // Note: This module is also mocked for Travis. If any changes are made to it, they should also
     // be made in: `schema_editor/test/mock/config.js`
+    //
+    // The web project loads the resources for this project, which includes this config.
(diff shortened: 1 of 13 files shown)


ddohler:
Why are Schema Editor resources getting loaded in the web app in the first place? My initial thought is that that would itself be something we should fix, but there might be a reason the web app needs them.

flibbertigibbet:
The editor is a dependency of the web app. There's several places where editor things are imported and referenced, including in tests, so the relationship seems very intentional. I don't know why the decision was made to do things that way.

moradology:
Ditto on what @flibbertigibbet said. It makes sense as things are now and I am unclear on why they are this way.

moradology:
+1

PR #142, opened 2015-09-16
Feature/session storage

flibbertigibbet:
Save and restore filters using session storage; fetch filters before loading record list.

Also fixes an issue with console errors and data loading twice on initial app load.


web/app/scripts/views/record/list-controller.js
```diff
@@ -18,7 +19,30 @@
         function initialize() {
             RecordState.getSelected().then(function(selected) { ctl.recordType = selected; })
                 .then(loadRecordSchema)
-                .then(loadRecords)
```
moradology: AH!

flibbertigibbet: That wasn't the source of the double-loading; that was [here](https://github.com/WorldBank-Transport/DRIVER/pull/142/files#diff-c2eba903e8631b2cfe3f6bef2b095e34R65).

moradology:
Solid PR. Took me a while to review this due to a little power struggle with my VM... +1

Issue #144, opened 2015-09-17
Create DB user for Windshaft


flibbertigibbet:
Given read-only permissions, and limit access to tables needed by Windshaft (probably doesn't need access to user info).

PR #145, opened 2015-09-17
Feature/user boundary filter

flibbertigibbet:
Filter map view by user drawn boundary.

Record list does not yet filter by user boundary, and the map view still does not filter by anything else, other than record type. Those things can go in separate PRs.

Companion PR on ashlar for the boundary filter back-end and endpoint used to get SQL to pass along to Windshaft: azavea/ashlar/pull/47

Also bumps memory for app Vagrant box.

### DIFF
--- a/windshaft/driver.js
+++ b/windshaft/driver.js
@@ -86,18 +86,20 @@ function getRequestParameters(request) {
         params.interactivity = 'uuid,occurred_from,data';
         params.style = eventsStyle;
 
-        // build query for record points
-        params.sql = baseRecordQuery;
-        if (params.id !== 'ALL') {
-            params.sql += filterRecordQuery + params.id + "'";
+        // build query for record points if do not already have a query
+        if (request.query.sql) {
+            params.sql = request.query.sql;
+        } else {
+            params.sql = baseRecordQuery;
+            if (params.id !== 'ALL') {
+                params.sql += filterRecordQuery + params.id + "'" + endRecordQuery;
+            }
         }
 
         if (request.query.heatmap) {
             // make a heatmap if optional parameter for that was sent in
             params.style = heatmapStyle;
         }
-
-        params.sql += endRecordQuery;
     } else {
         params.interactivity = 'label';
         params.style = boundaryStyle;
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -105,7 +105,7 @@ Vagrant.configure("2") do |config|
     app.ssh.forward_x11 = true
 
     app.vm.provider :virtualbox do |v|
-      v.memory = ENV.fetch("DRIVER_APP_MEM", 2094)
+      v.memory = ENV.fetch("DRIVER_APP_MEM", 3072)
       v.cpus = ENV.fetch("DRIVER_APP_CPUS", 2)
     end
   end
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -2,7 +2,7 @@
     'use strict';
 
     /* ngInject */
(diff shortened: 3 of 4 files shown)


flibbertigibbet:
image

windshaft/driver.js
```diff
@@ -86,18 +86,20 @@ function getRequestParameters(request) {
         params.interactivity = 'uuid,occurred_from,data';
         params.style = eventsStyle;
 
-        // build query for record points
-        params.sql = baseRecordQuery;
-        if (params.id !== 'ALL') {
-            params.sql += filterRecordQuery + params.id + "'";
+        // build query for record points if do not already have a query
+        if (request.query.sql) {
+            params.sql = request.query.sql;
```
sharph: Won't this allow an attacker (who has authenticated) to execute arbitrary SQL against the database?

sharph: Maybe sign the SQL with a combination of [itsdangerous](http://pythonhosted.org/itsdangerous/) and [node-jsjws](https://github.com/davedoesdev/node-jsjws)?

flibbertigibbet: Windshaft wraps all queries in a `SELECT`, which presumably would prevent actions other than reading. See #144, which should also add some controls via database roles. Windshaft is designed to allow arbitrary SQL queries.

sharph:
You can escape the `SELECT` wrapper with semicolons as demonstrated in https://github.com/WorldBank-Transport/open-transit-indicators/issues/360

I think #144 would be a prerequisite for accepting this PR if we're to go that route. Would there be any concerns about different orgs accessing each others data or is that out of scope at this point?

flibbertigibbet:
And WorldBank-Transport/open-transit-indicators#360 was closed without a fix. SQL vuln, inherent to Windshaft, is a larger issue that we'll have to address separately.

sharph:
https://github.com/WorldBank-Transport/open-transit-indicators/issues/360 was closed in https://github.com/WorldBank-Transport/open-transit-indicators/pull/413 though it wasn't properly linked.

I strongly disagree that there is a SQL vulnerability or problem in windshaft. winshaft assumes it is being given trusted data and we are breaking that assumption by passing it untrusted user submitted data.

flibbertigibbet:
CartoDB takes arbitrary, user-submitted SQL without any filtering or testing, which is passed straight to the SQL API and Windshaft. The only checks are via user/role permissions.

We need to implement custom, complex filters here, involving unknown fields, so limiting the options by field as we've done in the past won't work in this case. The SQL here is *not* from the user, it is from Django, via the app frontend. We don't have anything in place yet to prevent requests that aren't from the app going to Windshaft.

sharph:
Whether the request came *from the user* is an issue of semantics. The request was sent from the user's machine and should be considered untrusted. CartoDB implements access control at the database level with user permissions as you say, and *at the very least* we should lock down windshaft's user to only `SELECT`. Preferably limiting by field could be implemented, perhaps by signing the query as I've suggested earlier.

I'm afraid that implementing the feature without the relevant access controls in this case will open a security hole that will go forgotten.

flibbertigibbet:
There's an open issue for it I've created, so it will not be forgotten. We have not done the cards yet for user permissions, so what should have access is an open question. There is no sensitive data on staging, nor any significant negative consequences for data loss or tampering there. This PR is however needed for further work to be done, and security is again a larger issue best addressed separately, for completeness.

sharph:
I'll +1 with lots of hesitation and alarm flags raised to keep things moving along. Please consider adding a user and role with `SELECT` access to the relevant tables for windshaft -- it is not blocked by the user permission card and will at least keep someone from dropping tables or reading/`UPDATE`ing the users table.

PR #147, opened 2015-09-18
Create separate service for fetching boundary info.

### LINKED ISSUE #146: Long initial app load
The initial app load takes ~11.5s, running locally, before the schema populates and the filter bar and records can show up. The `/api/boundarypolygons/` fetch seems to taking up most of that time.

flibbertigibbet:
Fixes #146.

The app has been fetching the GeoJSON for the currently selected user boundary before loading the rest of the app, causing long load times. The bounds themselves are not needed in the UI, and will only be used in filtering for some reports. This fetches the boundary metadata instead, which is sufficient for the dropdown.


moradology:
LGTM

PR #148, opened 2015-09-21
Add datefilter directive and associated components

moradology:
This PR includes styling that should bring us much closer to the filterbar prototype

### DIFF
--- a/web/app/scripts/app.js
+++ b/web/app/scripts/app.js
@@ -62,6 +62,5 @@
     .config(LogConfig)
     .config(LeafletDefaultsConfig)
     .config(LocalStorageConfig)
-    .config(HttpConfig)
-    .run(function(FilterState) { FilterState.restoreFilters(); });
+    .config(HttpConfig);
 })();
--- a/web/app/scripts/app.js
+++ b/web/app/scripts/app.js
@@ -62,5 +62,6 @@
     .config(LogConfig)
     .config(LeafletDefaultsConfig)
     .config(LocalStorageConfig)
-    .config(HttpConfig);
+    .config(HttpConfig)
+    .run(function(FilterState) { FilterState.restoreFilters(); });
 })();
--- a/web/app/scripts/filterbar/date-range-directive.js
+++ b/web/app/scripts/filterbar/date-range-directive.js
@@ -2,7 +2,7 @@
     'use strict';
 
     /* ngInject */
-    function DateRangeField(FilterState, $timeout) {
+    function DateRangeField(FilterState) {
         var module = {
             restrict: 'A',
             require: ['^driver-filterbar', 'date-range-field'],
@@ -15,6 +15,7 @@
 
                 // restore previously set filter selection on page reload
                 scope.$on('driver.filterbar:restored', function(event, filter) {
+                    console.log(filter);
                     if (filter.label === filterLabel) {
                         scope.dtMin = filter.value.min;
                         scope.dtMax = filter.value.max;
--- a/web/app/scripts/filterbar/date-range-directive.js
+++ b/web/app/scripts/filterbar/date-range-directive.js
@@ -10,8 +10,10 @@
             controller: 'dateRangeController',
             link: function(scope, elem, attrs, ctlArray) {
                 var filterLabel = '__dateRange';
+                var filterBarCtl = ctlArray[0];
+                var dateRangeCtl = ctlArray[1];
                 init();
-                scope.calendarOptions = {'format': ctlArray[1].dateTimeFormat};
+                scope.calendarOptions = {'format': filterBarCtl.dateTimeFormat};
 
                 // restore previously set filter selection on page reload
                 scope.$on('driver.filterbar:restored', function(event, filter) {
@@ -40,7 +42,7 @@
                  */
                 scope.updateFilter = function(filterLabel, filterObj) {
                     if (scope.isMinMaxValid()) {
-                        ctlArray[0].updateFilter(filterLabel, filterObj);
+                        filterBarCtl.updateFilter(filterLabel, filterObj);
                     }
                 };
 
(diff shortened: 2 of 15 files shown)


web/test/spec/filterbar/date-range-directive.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+        $rootScope.$apply();
+    }));
+
+    it('should handle restoring filter selection', function () {
+        // should have no maximum set yet
+        expect(Element.find("input[type='text'][name='maximum']").val()).toEqual('');
+
+        // mine goes to eleven
+        $rootScope.$broadcast('driver.filterbar:restore', {'__dateRange': {min: 0, max: 11}});
```
flibbertigibbet: Maybe test this with an actual date object instead of a number?

web/app/scripts/state/filters-service.js
```diff
@@ -16,29 +23,45 @@
...
+
 
         /**
          * Broadcast event to trigger setting previously stored filters back on the filter bar.
          */
-        svc.restoreFilters = function() {
+        function restoreFilters() {
             var filterString = localStorageService.get(storageName);
-            var filterObj = !!filterString ? JSON.parse(filterString) : {};
+            var filterObj = !!filterString ? filterString : {};
```
flibbertigibbet: Why remove the `parse` here? Storage contains stringified objects.

web/app/scripts/state/filters-service.js
```diff
@@ -16,29 +23,45 @@
          *
          * @param {Object} filters FilterBar's filters object to restore on load.
          */
-        svc.saveFilters = function(filters) {
-            localStorageService.set(storageName, JSON.stringify(filters));
-        };
+        function saveFilters(filters) {
+            localStorageService.set(storageName, filters);
```
flibbertigibbet: Objects must be stringified to go into storage.

web/app/scripts/filterbar/numeric-range-directive.js
```diff
@@ -15,17 +15,18 @@
...
                     scope.error = {};
-
-                    // restore previously set filter selection on page reload
-                    scope.$on('driver.filterbar:restored', function(event, filter) {
-                        if (filter.label === scope.label) {
-                            scope.filter = filter.value;
-                            scope.isMinMaxValid();
-                        }
-                    });
+                    scope.filter = FilterState.get(scope.label);
```
flibbertigibbet: I don't see the reason for the changes to this file. `FilterState` should not be referenced in the specific filter directives; that's managed at a higher level. The numeric range filters were working fine.

web/app/scripts/state/filters-service.js
moradology: The localStorageService already parses in its get method, so I think this was noop at best

web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -0,0 +1,68 @@
...
+            templateUrl: 'scripts/filterbar/date-range.html',
+            controller: 'dateRangeController',
+            link: function(scope, elem, attrs, ctlArray) {
+                var filterLabel = '__dateRange';
+                init();
+                scope.calendarOptions = {'format': ctlArray[1].dateTimeFormat};
+
+                // restore previously set filter selection on page reload
+                scope.$on('driver.filterbar:restored', function(event, filter) {
+                    console.log(filter);
```
flibbertigibbet: Remove console log here, or use `$log`.

web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -0,0 +1,68 @@
...
+    function DateRangeField(FilterState, $timeout) {
+        var module = {
+            restrict: 'A',
+            require: ['^driver-filterbar', 'date-range-field'],
+            templateUrl: 'scripts/filterbar/date-range.html',
+            controller: 'dateRangeController',
+            link: function(scope, elem, attrs, ctlArray) {
+                var filterLabel = '__dateRange';
+                init();
+                scope.calendarOptions = {'format': ctlArray[1].dateTimeFormat};
```
flibbertigibbet: It would be nice to locally define controller vars instead of referencing them by array offset, for clarity.

web/app/scripts/app.js
```diff
@@ -62,5 +62,6 @@
     .config(LogConfig)
     .config(LeafletDefaultsConfig)
     .config(LocalStorageConfig)
-    .config(HttpConfig);
+    .config(HttpConfig)
+    .run(function(FilterState) { FilterState.restoreFilters(); });
```
flibbertigibbet: Why is this here?

web/app/scripts/state/filters-service.js
moradology: I think it automatically serializes as well

flibbertigibbet: Ok, neat

web/app/scripts/app.js
moradology: An oversight left from something I was testing out - removing

web/app/scripts/filterbar/date-range-directive.js
moradology: That's a good idea

web/app/scripts/state/filters-service.js
```diff
@@ -16,29 +23,45 @@
...
 
             $log.debug('Restoring filters:');
             $log.debug(filterObj);
 
             // tell the filterbar to set the filters back
             $rootScope.$broadcast('driver.filterbar:restore', filterObj);
-        };
+        }
 
+        svc.restoreFilters();
```
flibbertigibbet: Why is this here?

web/bower.json
```diff
@@ -24,7 +24,8 @@
     "Leaflet.utfgrid": "danzel/Leaflet.utfgrid",
     "d3": "3.5.5",
     "d3-tip": "^0.6.7",
-    "moment": "^2.10.6"
+    "moment": "^2.10.6",
+    "jquery": "^2.1.4"
```
flibbertigibbet: What requires proper jQuery? This won't conflict with the Angular jQuery lite?

moradology: I think angular can use jQuery rather than jqLite, but you're right that it is not necessary - angular-ui-bootstrap's implementation doesn't require jQuery like the datepicker otherwise would

flibbertigibbet:
+1

PR #152, opened 2015-09-22
Feature/geo filter records list

flibbertigibbet:
Assorted filter-related changes:

* Use all other, non-geo filters on map
* Use geo filter on records list
* Fix loading filters from storage on map view
* Save and load geo filter in storage
* Fix load events triggering multiple times after state changes

### DIFF
--- a/web/app/scripts/app.js
+++ b/web/app/scripts/app.js
@@ -35,12 +35,6 @@
         $httpProvider.defaults.xsrfCookieName = 'csrftoken';
     }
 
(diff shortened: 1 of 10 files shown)


moradology:
I'm almost certainly going to +1 this - it looks good. Let me make a couple changes that fix the way the date filter is stored on the filter service so that I can test things on my own box really quick (I'm a bit paranoid when events are involved).

moradology:
+1!

PR #154, opened 2015-09-23
Cast query fields to strings, for interactivity.

### LINKED ISSUE #153: Interactivity broken by custom filters
Map popups no longer display, due to the `SELECT` clause of the passed query not casting its fields to strings. One solution might be to just return the `WHERE` clause from the Django query endpoint instead of the full query, and define the fields in Windshaft.

flibbertigibbet:
Fixes #153.

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -150,6 +150,31 @@
         }
 
         /**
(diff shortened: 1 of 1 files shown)


kshepard:
+1

kshepard:
This has been manually merged.

PR #156, opened 2015-09-24
Grab polygons rather than geographies

moradology:
It looks like this was errantly changed to query the same endpoint as geographies. Reverting to the previous behavior - if performance is an issue, we'll have to do something on the backend.

### DIFF
--- a/web/app/scripts/navbar/navbar-controller.js
+++ b/web/app/scripts/navbar/navbar-controller.js
@@ -18,7 +18,6 @@
         function init() {
             GeographyState.getOptions().then(function(opts) { ctl.geographyResults = opts; });
             RecordState.getOptions().then(function(opts) { ctl.recordTypeResults = opts; });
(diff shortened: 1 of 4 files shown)


kshepard:
+1, Looks like the original change was made in: https://github.com/WorldBank-Transport/DRIVER/pull/147
in an attempt to speed up the page load. I think it'd be a pretty quick change to add a parameter to the polygon request that makes it so the actual geometry field itself isn't returned. Would you be able to spend a little time on trying to implement that before merging? It'd be good to not go back to the slow page load if it's not too much extra work.

moradology:
I'm with you and I think the concern raised was valid, but that card was 0 points. I'm probably a bit beyond 15 minutes as it is.. I could add a TODO?

kshepard:
That's fair, didn't realize it was only 0 points. If it's not trivial, re-open #146.

PR #157, opened 2015-09-24
Remove filterbar where its beauty isn't appreciated

moradology:


web/app/scripts/navbar/navbar-controller.js
```diff
@@ -52,6 +57,13 @@
             });
         });
 
+        // A function to set properties related to whether or not the filterbar should be instantiated for a given page
+        function setFilters(state) {
+            var filterPages = ['Map', 'Record List'];
```
kshepard: My initial thought is that this would probably work better as flags in the route states themselves, similar to how the 'showInNavbar' flag is used. It would be nice to have this type of configuration logic in one place rather than depending on the navbar to have knowledge about the individual pages. Would you be able to explore that a bit?

kshepard:
+1, consider moving isFilterPage logic to a flag on route states.

jbranigan:
Also, remove from add/edit/view a record. It's not going to show up on any other pages - just the map and the records list.

PR #164, opened 2015-09-30
Feature/generic record filter

moradology:
This task really spans multiple cards - we need a generic filter construction service for the later card to hook up to the JSONB filters but we also require the ability to filter on the frontend for the date range functionality. I've implemented the structure of the QueryBuilder (it is pretty simple, source should be explanatory enough) - it currently uses only date ranges.

Test currently ensures that the function actually uses the selected recordtype in its query construction


web/app/scripts/resources/query-builder-service.js
```diff
@@ -0,0 +1,76 @@
...
+            doFilter = doFilter !== undefined ? doFilter : true;
+            assembleParams(doFilter, offset).then(function(params) {
+                Records.get(params).$promise.then(function(records) {
+                    deferred.resolve(records);
+                });
+            });
+            return deferred.promise;
+        }
+
+        function windshaftQuery(doFilter) {
```
kshepard: Should this return a promise too?

web/app/scripts/resources/query-builder-service.js
```diff
@@ -0,0 +1,76 @@
...
+                Records.get(params);
+            });
+        }
+
+        function assembleParams(doFilter, offset) {
+            var deferred = $q.defer();
+            var paramObj = {};
+            /* jshint camelcase: false */
+            if (doFilter) {
+                if ('__dateRange' in FilterState.filters) {
```
kshepard: I think this may get messy when logic for all the filters are added. It seems like it should be the responsibility of the individual filters to constructor the parameters that they need.

web/test/spec/resources/query-builder-service-spec.js
```diff
@@ -0,0 +1,38 @@
...
+    beforeEach(inject(function (_$rootScope_, _$httpBackend_,
+                                _QueryBuilder_, _DriverResourcesMock_, _ResourcesMock_) {
+        $httpBackend = _$httpBackend_;
+        QueryBuilder = _QueryBuilder_;
+        $rootScope = _$rootScope_;
+        DriverResourcesMock = _DriverResourcesMock_;
+        ResourcesMock = _ResourcesMock_;
+    }));
+
+    it('should result in a call out to determine the selected RecordType', function () {
```
kshepard: The date range can use a test too.

web/app/scripts/resources/query-builder-service.js
```diff
@@ -0,0 +1,76 @@
...
+                    }
+                }
+            }
+
+            // Pagination offset
+            if (offset) {
+                paramObj.offset = offset;
+            }
+
+            // Record Type and offset
```
kshepard: Seems to be an out-of-date comment (mention of offset).

kshepard:
+1, consider decoupling individual filter logic from query builder service.

web/app/scripts/resources/query-builder-service.js
moradology: I think so. Some of this is very skeletal at the moment, but I'll go ahead and make that piece consistent.

web/app/scripts/resources/query-builder-service.js
moradology: Because dates are different (they don't exist as part of the partial ordering that the rest of our filters do), I handled this case a bit differently. The prefixed '__' is being used because this filter is a bit like a reserved word.

moradology:
The only difficulty in decoupling is that we've got a couple of things they need to handle:
1. Actually generating the URL we'll use to get the data (what we've got here)
2. Being filled back out on refresh and being restored from saved JSON

The approach I've taken to solving these two problems is to treat the filters as nothing more than views over data which at the last moment is transformed in whatever fashion is of interest to the query in question. There's also some difficulty, it seems to me, in reasoning about what the filters could provide that is generic for both windshaft and django that they don't already. I'm going to merge this, but I think that your worry/thought here is worth more consideration

PR #167, opened 2015-10-02
Add boundary layer picker

ddohler:
Adds a layer picker so that users can display individual boundaries. Also displays each boundary using the color pulled from that boundary's database record.
prsboundaries

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -2,8 +2,9 @@
     'use strict';
 
     /* ngInject */
-    function DriverLayersController($log, $scope, $rootScope, $timeout,
-                                    WebConfig, FilterState, RecordState, Records, QueryBuilder) {
+    function DriverLayersController($q, $log, $scope, $rootScope, $timeout,
+                                    WebConfig, FilterState, RecordState, GeographyState,
+                                    Records, QueryBuilder) {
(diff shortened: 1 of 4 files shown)


web/app/scripts/views/map/layers-controller.js
```diff
@@ -203,8 +204,19 @@
             });
 
             // user-uploaded boundary layer(s)
-            var boundaryLayerOptions = angular.extend(defaultLayerOptions, {zIndex: 2});
-            var boundaryLayer = new L.tileLayer(boundaryUrl, boundaryLayerOptions);
+            var availableBoundaries = $q.defer();
+            GeographyState.getOptions().then(function(boundaries) {
+                var _overlays = {};
+                var boundaryLayerOptions = angular.extend(defaultLayerOptions, {zIndex: 2});
+                boundaries.map(function(boundary) {
```
kshepard: This can probably be simplified a little bit by having the map return pairs of [label, boundaryLayer] and calling _.zipObject on it, rather than appending to the _overlays variable.

kshepard:
+1

web/app/scripts/views/map/layers-controller.js
ddohler: I hadn't noticed `_.zipObject` before, very nice.

PR #168, opened 2015-10-05
Update pip command for external sources in ashlar's setup.py

moradology:
This PR should be merged in only when merging in https://github.com/azavea/ashlar/pull/49


ddohler:
+1, looks like this part of pip is kind of a mess.

PR #169, opened 2015-10-05
Update production dockerfile for djsonb install

moradology:
This should fix the build step; which currently fails


moradology:
Trivial - mirrors previous changes

PR #171, opened 2015-10-05
Styling fixes from October 5th

mtedeschi:
A number of things to note in this pull request (each commit has details, and I've started discussions on several issues now labeled "Help wanted" where needed).

•  We have prototypes, and where possible, we should try to use them. I only noticed a few places, particular the filter bar, where we need to pull in that styling.
•  We shouldn't hide buttons, instead use disabled state when not clickable.
•  I've added in a prototype view of the "Saved filters" modal in the `topic/prototype` directory.
•  I made some suggestions on how to improve the TODDOW styling, and added some style to the tooltip which currently gets a class of "n"—if that changes, we'll need to update the Sass.

### DIFF
--- a/app/Dockerfile.base
+++ b/app/Dockerfile.base
@@ -4,7 +4,7 @@ RUN mkdir -p /opt/app
 WORKDIR /opt/app
 
 COPY . /opt/app
-RUN pip install --no-cache-dir -r requirements.txt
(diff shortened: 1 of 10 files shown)


kshepard:
+1, I'll merge this and test it out.

PR #177, opened 2015-10-08
Feature/map persistence

moradology:


web/app/scripts/state/mapstate-service.js
```diff
@@ -0,0 +1,79 @@
...
+        return svc;
+
+        /**
+         * Set filter geojson
+         * @param {object} geojson for a filter
+         */
+        function setFilterGeoJSON(geojson) {
+            $log.debug('Setting drawn layer');
+            filterGeoJSON = geojson;
+            return;
```
kshepard: This `return` is unnecessary here, and also in the `getZoom` function.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -22,6 +22,11 @@
         var recordsUrl = '/tiles/table/ashlar_record/id/ALL/{z}/{x}/{y}.png';
         var recordsUtfgridUrl = '/tiles/table/ashlar_record/id/ALL/{z}/{x}/{y}.grid.json';
         var boundaryUrl = '/tiles/table/ashlar_boundary/id/ALL/{z}/{x}/{y}.png';
+        var filterStyle = {
+            'color': '#f357a1',
+            'fillColor': '#f357a1',
+            'fill': true,
```
kshepard: The quotes around the keys can be removed, as well as the trailing comma.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -116,25 +121,47 @@
...
-
             ctl.editLayers.clearLayers();
-            $rootScope.$broadcast('driver.views.map:filterdrawn', null);
+
+            layer.setStyle(filterStyle);
             ctl.editLayers.addLayer(layer);
+            $rootScope.$broadcast('driver.views.map:filterdrawn');
+
+            // Use GeoJSON instead of a normal layer - theres a strange bug likely stemming from
+            //  race conditions on the Leaflet Map object otherwise
```
kshepard: Weird, sounds scary.

kshepard:
+1

web/app/scripts/views/map/layers-controller.js
flibbertigibbet: Didn't see that issue before. Might be something other than Leaflet.

moradology: It could well be the Leaflet-Draw plugin - it was the layer generated by that library's event which simply would not be added to the map on pain of complaints about map._viewPort.min (this is from memory, so the exact var name is probably slightly different) being undefined.

flibbertigibbet: Seems unlikely, as this has been using `leaflet-draw` since it was set up without issue.

moradology: Yes, but leaflet-draw layers haven't been saved and then applied to a newly initialized map until this PR. Here's what I was seeing from a high level:
1. I save a layer (e.layer from leaflet-draw's event)
2. I attempt to map.addLayer it
3. A stack trace bursts out from deep within the bowels of Leaflet complaining of undefined properties

I'm just creating a layer afresh - that's literally the only difference - so I have a very hard time believing that this somehow involves code we wrote. Now, perhaps I just fundamentally misunderstand the leaflet ontology - if that is so, I'd definitely like to be corrected - but with the concepts I've got at my disposal, I just don't know how else to think about this problem. 

Would you like to take a look before I merge this in? Like I said, if there's something I'm just ignorant of, I'd like to be corrected

flibbertigibbet: Ok, yeah, extracting it to GeoJSON for the re-load makes sense. I don't think Leaflet layers are intended to be serializable.

moradology: Cool, I'll go ahead and merge. And seriously: I appreciate the second guess

PR #178, opened 2015-10-09
Feature/nominatim

sharph:


README.md
```diff
@@ -8,6 +8,9 @@ DRIVER - Data for Road Incident Visualization, Evaluation, and Reporting
 Requires Vagrant 1.5+, Ansible 1.8+ and the following plugins:
   - `vagrant-hostmanager`
 
+Copy `development/ansible/group_vars/all.example` to `development/ansible/group_vars/all`
```
kshepard: s/development/deployment/

sharph: Whoops! Good catch.

README.md
```diff
@@ -8,6 +8,9 @@ DRIVER - Data for Road Incident Visualization, Evaluation, and Reporting
 Requires Vagrant 1.5+, Ansible 1.8+ and the following plugins:
   - `vagrant-hostmanager`
 
+Copy `deployment/ansible/group_vars/all.example` to `deployment/ansible/group_vars/all`
+and add the API key for pickpoint under `web_js_nominatim_key`.
```
kshepard: Maybe add a link to pickpoint. I wasn't familiar with the service before seeing it here.

web/app/scripts/nominatim/module.js
```diff
@@ -0,0 +1,8 @@
+(function () {
+    'use strict';
+
+    angular.module('driver.nominatim', [
```
kshepard: Probably an annoying request, but I think it would be nicer if all of these references to `nominatim` (module, service, variables) were changed to something more general and self-explanatory, like `geocoder`. It would also make it less confusing if we were to do something such as switching geocoders in the future.

sharph: I hear what you're saying, but the nominatim doesn't behave like other geocoders. It seems to be geared toward matching points up with place names, like the name of a street or a building, rather than a street address, although it does that too. The technical spec references using the OSM nominatim specifically, of which I chose one of a few providers. Since what the "nominatim" does seems to be slightly different than a geocoder as we know it, I gave it a distinct name.

If you think we should think of this feature more genericly, I can rename it to geocoder.

kshepard: I see, that's good reasoning. I'm fine keeping it how it is. If we do need to support multiple geocoders at some point, maybe we can then create a geocoder service which calls out to the individual distinct geocoders.

web/app/scripts/views/record/add-edit-partial.html
```diff
@@ -25,6 +25,14 @@
                         </div>
                     </div>
                     <div class="col-md-12">
+                        <div class="form-group">
+                            <input type="text" class="form-control"
```
kshepard: This could also use a label: maybe something like "Location label", or something along those lines.

web/app/scripts/views/record/add-edit-controller.js
```diff
@@ -18,10 +20,14 @@
             ctl.onSaveClicked = onSaveClicked;
             ctl.occurredFromChanged = occurredFromChanged;
             ctl.onGeomChanged = onGeomChanged;
+            ctl.geocode = geocode;
+            ctl.geocodeSelect = geocodeSelect;
 
             ctl.occurredFromOptions = {format: dateTimeFormat};
             ctl.occurredToOptions = {format: dateTimeFormat};
 
+            ctl.reverseGeocoder = '';
```
kshepard: This name can be a bit more descriptive for what it represents. Probably match it up for whatever is chosen for the label of that field.

kshepard:
This code looks good, but there are a couple pieces of functionality that should be fixed before merging:
* When selecting a location from the dropdown, the map should pan & zoom to that location
* After selecting a location from the dropdown, a reverse geocode request is currently being issued. This is superfluous and also leads to some confusion since the reverse geocoded response doesn't always match up with the originally selected location.

sharph:
I've addressed issues you've pointed out.

sharph:
Also, I've changed mentions of "geocoder" to "nominatim."

kshepard:
+1. One thing I just noticed: it'd probably be a good idea to add _.debounce to the `forward` call, since currently, a request will be sent with each character typed.

sharph:
Added `typeahead-wait-ms` to input box.

kshepard:
Nice, didn't know about `typeahead-wait-ms`, that's pretty handy.

PR #180, opened 2015-10-09
Scale dashboard charts on resize

kshepard:


moradology:
+1; can't believe it was so little code!

PR #184, opened 2015-10-12
Recent events dashboard map

ddohler:
Uses a 14-day cutoff for recent events; if none occurred, none will appear on the map. I added a service for constructing tile layer urls because we use them in several spots, but I didn't have time to fully integrate it with some of the other maps so I put in a todo.
prsrecentevents

### DIFF
--- a/web/app/scripts/views/dashboard/dashboard-partial.html
+++ b/web/app/scripts/views/dashboard/dashboard-partial.html
@@ -8,11 +8,11 @@
             </div>
         </div>
         <div class="block block-1x2 quick-map">
-            <div element-stats class="block-inner">
-                <driver-stepwise chart-data="ctl.records"></driver-stepwise>
+            <div class="block-inner">
+                <div class="map" leaflet-map recent-events></div>
             </div>
-                <h3>Recent Accidents
-                    <a href="javascript:void(0);" class="small pull-right">View all accidents »</a>
+                <h3>{{ ctl.recordType.plural_label }} during the past two weeks
+                  <a ui-sref="map" class="small pull-right">View all {{ ctl.recordType.plural_label | lowercase }} »</a>
                 </h3>
             </div>
         </div>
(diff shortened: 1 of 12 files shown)


flibbertigibbet:
Is the `View all accidents` link not implemented? Seems like there should be a way to toggle between the two.

flibbertigibbet:
I'm a bit concerned that the refactor started here might not get finished. Hopefully you'll have a chance to complete that soon.

+1

ddohler:
Good call; I assume that link is actually meant to take you to either the Map page or the Record List page. I'll set it to the Map page for now. I don't think that link was meant to be a toggle though.

I'll make sure to finish up the refactor by the end of the sprint, I should be able to build it into other cards.

web/app/scripts/views/dashboard/dashboard-partial.html
```diff
@@ -8,10 +8,10 @@
             </div>
         </div>
         <div class="block block-1x2 quick-map">
-            <div element-stats class="block-inner">
-                <driver-stepwise chart-data="ctl.records"></driver-stepwise>
+            <div class="block-inner">
+                <div class="map" leaflet-map recent-events></div>
             </div>
-                <h3>Recent Accidents
+                <h3>Accidents during the past two weeks
```
kshepard: Maybe use `{{ ctl.recordType.plural_label }}` here, so it could easily apply to multiple record types.

ddohler: Nice, done.

web/app/scripts/views/dashboard/dashboard-partial.html
```diff
@@ -8,11 +8,11 @@
...
         <div class="block block-1x2 quick-map">
-            <div element-stats class="block-inner">
-                <driver-stepwise chart-data="ctl.records"></driver-stepwise>
+            <div class="block-inner">
+                <div class="map" leaflet-map recent-events></div>
             </div>
-                <h3>Recent Accidents
-                    <a href="javascript:void(0);" class="small pull-right">View all accidents »</a>
+                <h3>{{ ctl.recordType.plural_label }} during the past two weeks
+                    <a ui-sref="map" class="small pull-right">View all accidents »</a>
```
kshepard: Also here :)

PR #185, opened 2015-10-12
Remove element-stats directive

kshepard:
It is no longer needed, since chart resizing is being handled in a different manner.


moradology:
+1 on travis success

PR #188, opened 2015-10-12
Feature/back on cancel

flibbertigibbet:
* Moves the add new record button to the filterbar, so new records can be added from either the map or record list view (wireframes had it in both places).

* Uses browser history to allow `Cancel` button to go back to record list or map view, as is appropriate.

### DIFF
--- a/web/app/scripts/filterbar/filterbar.html
+++ b/web/app/scripts/filterbar/filterbar.html
@@ -1,5 +1,5 @@
 <div class="navbar navbar-default drop-shadow filterbar collapse navbar-collapse">
-    <div class="pull-left form-inline nav navbar=nav">
+    <div class="navbar-left form-inline nav navbar=nav">
         <div class="navbar-header">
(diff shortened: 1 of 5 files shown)


kshepard:
+1

web/app/scripts/filterbar/filterbar.html
```diff
@@ -11,6 +11,12 @@
                 <li ng-if="value.format=='number'" numeric-range-field label="key" data="value" class="dropdown"></li>
                 <li ng-if="value.fieldType=='selectlist'" options-field label="key" data="value" class="dropdown"></li>
             </ul>
+            <ul class="nav navbar-nav navbar-right">
+                <!-- TODO: this is a temporary location for the Add Record button which has been removed -->
```
kshepard: This TODO can safely be removed. It's in a good location now, and there's an issue open for styling it better.

flibbertigibbet: Ok, cool

PR #189, opened 2015-10-12
Fix edit/delete draw tools

kshepard:
The reasons why this wasn't working is that the `featureGroup` key was incorrectly capitalized, and the `edit` key was inside the `draw` object, when it should have been parallel to it.


ddohler:
+1

PR #190, opened 2015-10-12
Fix latlng position and precision

fungjj92:
Round lat long on driver record views to 5 decimal places (< 1 m error) and move under map


kshepard:
This looks good, but the change also needs to be made on the add/edit page (add-edit-partial.html).

fungjj92:
Heard & done!

fungjj92:
Heard & done!

kshepard:
+1, nice. I'll merge this since you're not a member of the org yet.

Issue #176, opened 2015-10-08
Retina tiles issue


jbranigan:
On retina screens, tile labels are unreadably small.

http://leafletjs.com/reference.html#tilelayer-detectretina

flibbertigibbet:
We are using `detectRetina`: https://github.com/WorldBank-Transport/DRIVER/blob/32316ef96cc46bfd1c669247a2cad81e050169b2/web/app/scripts/views/map/layers-controller.js#L172

jbranigan:
Then maybe it's not working - not a super high priority, but it's something we as a team should know how to manage.

Issue #159, opened 2015-09-29
Style TODDOW on Dashboard view


mtedeschi:
00 through 23 for hours is probably more universal and fits better. Style hover effect popovers. Give the squares a nicer color and light gray when no incidents are reported.

mtedeschi:
Did we have a request to shift it to 12-hour time from 24-hour? If not, I think we could switch it to what it was in the wireframes unless we have a good reason not to.

mtedeschi:
For @mtedeschi: FIx carat to align to the bottom of the tooltip, set pointer-event: none;

PR #192, opened 2015-10-13
TODDOW fixups

ddohler:
This does most of the tasks from #159; I'll leave it up to @mtedeschi to close that
 since there are some tasks there for him too.

The bulk of the changes, however, are related to the refactor mentioned in #184 so that all three maps use URLs generated by the `TileUrlService`.

prstoddow
(This toddow is using the old imported events so it doesn't have a good range of dates and times)

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -46,7 +46,7 @@
                 ctl.baseMaps = baseMaps.promise;
                 var streetsOptions = {
                     attribution: cartoDBAttribution,
-                    detectRetina: true,
+                    detectRetina: false,
                     zIndex: 1
                 };
                 TileUrlService.baseLayerUrl().then(function(streetsUrl) {
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -4,7 +4,7 @@
     /* ngInject */
     function DriverLayersController($q, $log, $scope, $rootScope, $timeout,
                                     WebConfig, FilterState, RecordState, GeographyState,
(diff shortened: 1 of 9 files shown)


kshepard:
Judging by the screenshot: the hour numbers should be shifted a little bit so they line up over the center of the squares.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -213,87 +204,103 @@
...
-            // layer with heatmap of events
-            var heatmapOptions = angular.extend(defaultLayerOptions, {zIndex: 4});
-            var heatmapLayer = new L.tileLayer(ctl.getFilteredRecordUrl(heatmapUrl), heatmapOptions);
+            $q.all([TileUrlService.recTilesUrl(ctl.recordType),
+                    TileUrlService.recUtfGridTilesUrl(ctl.recordType),
+                    TileUrlService.recHeatmapUrl(ctl.recordType)]).then(function(tileUrls) {
+                var baseRecordsUrl = tileUrls[0];
+                var baseUtfGridUrl = tileUrls[1];
+                var baseHeatmapUrl = tileUrls[2];
+                var defaultLayerOptions = {attribution: 'PRS', detectRetina: true};
```
kshepard: While you're in here, I believe if you switch these `detectRetina` values to `false`, it'll fix #176.

kshepard:
+1

web/app/scripts/views/map/layers-controller.js
ddohler: Done; I only changed this to false for the base layer, so boundaries and events should still come in at high res.

ddohler:
I nudged the numbers a bit; they're not truly centered but they're pretty close.

Issue #191, opened 2015-10-13
Schema editor not paginated


flibbertigibbet:
The schema editor UI only shows the first ten items, without pagination, so there's no way to access the rest.

PR #193, opened 2015-10-13
Feature/scrolling dropdowns

flibbertigibbet:
Depends on companion PR azavea/ashlar#52.

image

Partially addresses #191 by fetching all record types into the schema editor, but paginating with screen controls would probably be better than a scrolling page with everything. Also, the editor sidebar apparently doesn't scroll, so the geographies get cut off if there are too many record types listed.


kshepard:
+1

PR #196, opened 2015-10-14
Feature/50 results

fungjj92:
Record list displays 50 results
Edit config file to contain new global variable should we want to quick change search result quantity
Add OSX file ignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,9 @@ var/
 .installed.cfg
 *.egg
 
+#OSXIgnore
+**/.DS_Store
+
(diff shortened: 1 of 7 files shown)


kshepard:
+1, looks good. Before merging, let's clean up those merge commits, which are making for a messy commit history. I'll give you a hand with that.

kshepard:
Oh, and as for the failing tests: see the note regarding Travis in `web-config-js.j2`. You can run the tests and jshint locally with `scripts/grunt.sh web travis`.

PR #201, opened 2015-10-15
Boundary filtering / zooming

ddohler:
- Adds zooming to selected boundaries to both the big map and the recent events map.
- Adds filtering by selected boundary to the Record list view
- Adds logic for, but does not enable, filtering the maps by the selected boundary. This is blocked pending updates to the Windshaft / Django filtering process.
- Does not filter the TODDOW or 30-90-365 views; those were being changed in parallel to this and should be revisited along with the map filtering.


flibbertigibbet:
State is (or should be) set up so that on re-visiting the map page, it will go back to its previous center and zoom level. Will that work with this, and do so without panning/zooming twice?

ddohler:
Good call, I had forgotten about that. I've moved the zoom-on-init code out of the directive and into the recent events map itself. The current behavior of the big map is now preserved.

flibbertigibbet:
Nice! +!

PR #202, opened 2015-10-15
Feature/stepwise reformat

flibbertigibbet:
Formatting for the stepwise graph.

* Rotates x-axis labels and changes the format
* Adds interactivity with event counts

@jbranigan I've left in the dots, formatted grey and translucent now, so the interactivity points are visible.

image


kshepard:
+1

PR #203, opened 2015-10-15
Dashboard cleanup/split out into widgets

kshepard:


web/app/scripts/recent-counts/recent-counts-directive.js
```diff
@@ -13,6 +13,6 @@
     }
 
     angular.module('driver.recentCounts')
-    .directive('recentCounts', RecentCounts);
+    .directive('driverRecentCounts', RecentCounts);
```
moradology: This is better. Wasn't sure what to call it.

kshepard: We have a `driver` prefix on all the directives, so I just changed it to be consistent.

web/test/spec/recent-counts/recent-counts-controller.spec.js
```diff
@@ -0,0 +1,42 @@
...
+        $httpBackend.expectGET(recordTypeUrl).respond(200, ResourcesMock.RecordTypeResponse);
+        $httpBackend.expectGET(recordTypeUrl).respond(200, ResourcesMock.RecordTypeResponse);
+
+        var recordType = ResourcesMock.RecordType;
+        var recordTypeId = recordType.uuid;
+        var countsUrl = new RegExp('api/recordtypes/' + recordTypeId +
+                                   '/recent_counts/\\?limit=all');
+        $httpBackend.expectGET(countsUrl).respond(200);
+        $httpBackend.expectGET(countsUrl).respond(200);
+        $httpBackend.expectGET(countsUrl).respond(200);
```
moradology: I'm going to make an issue to investigate the measures taken to prevent redundant requests...

moradology:
+1

web/test/spec/recent-counts/recent-counts-controller.spec.js
kshepard: Cool, good idea. I was going to ask you about that -- only just saw it yesterday when I noticed that a bunch of the tests had duplicate requests.

PR #205, opened 2015-10-16
Feature/options filter

moradology:
This lays the groundwork for completion of all djsonb filtering - as you can see (in a TODO) all we'll have to do is add a containment filter function for `related_contains`

### DIFF
--- a/web/app/scripts/filterbar/filterbar.html
+++ b/web/app/scripts/filterbar/filterbar.html
@@ -8,12 +8,12 @@
                 <li date-range-field class="dropdown"></li>
             </ul>
             <ul class="nav navbar-nav" ng-repeat="(key, value) in filterbar.filterables">
-                <li ng-if="value.format=='number'"
+                <li ng-if="value.format === 'number'"
                     numeric-range-field
                     label="key"
                     data="value"
                     class="dropdown"></li>
-                <li ng-if="value.fieldType=='selectlist'"
+                <li ng-if="value.fieldType === 'selectlist'"
                     options-field
                     label="key"
                     data="value"
--- a/web/app/scripts/filterbar/filterbar.html
+++ b/web/app/scripts/filterbar/filterbar.html
@@ -8,8 +8,17 @@
                 <li date-range-field class="dropdown"></li>
             </ul>
             <ul class="nav navbar-nav" ng-repeat="(key, value) in filterbar.filterables">
-                <li ng-if="value.format=='number'" numeric-range-field label="key" data="value" class="dropdown"></li>
-                <li ng-if="value.fieldType=='selectlist'" options-field label="key" data="value" class="dropdown"></li>
+                <li ng-if="value.format=='number'"
+                    numeric-range-field
+                    label="key"
+                    data="value"
+                    class="dropdown"></li>
+                <li ng-if="value.fieldType=='selectlist'"
+                    options-field
+                    label="key"
+                    data="value"
+                    class="dropdown">
+                </li>
             </ul>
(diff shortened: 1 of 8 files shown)


web/app/scripts/filterbar/label-formatter.js
```diff
@@ -6,14 +6,15 @@
...
+         * @return {String} Label prettified to just `child`
          */
         return function(label) {
             if (!label) {
                 return '';
             }
 
-            return label.replace(/#/, ': ');
+            var split = label.split('#');
+            return split[split.length - 1];
```
flibbertigibbet: Might be a little more readable to just `.pop()` here.

web/app/scripts/filterbar/filterbar.html
```diff
@@ -8,8 +8,17 @@
                 <li date-range-field class="dropdown"></li>
             </ul>
             <ul class="nav navbar-nav" ng-repeat="(key, value) in filterbar.filterables">
-                <li ng-if="value.format=='number'" numeric-range-field label="key" data="value" class="dropdown"></li>
-                <li ng-if="value.fieldType=='selectlist'" options-field label="key" data="value" class="dropdown"></li>
+                <li ng-if="value.format=='number'"
```
kshepard: To align with our js standards, this should be `value.format === 'number'`, same thing with the selectlist (I realize this wasn't added here, but would still be worth updating).

web/app/scripts/filterbar/options-directive.js
```diff
@@ -35,11 +38,22 @@
...
-                    var filter = angular.copy(scope.filter);
-                    filter.contains = [filter.contains];
-                    filterbarController.updateFilter(filterLabel, _.merge({'_rule_type': 'containment'}, filter));
+                    // only include filters that actually do something
+                    if (scope.filter.contains.length) {
+                        // handle syntactic differences necessitated by having related objects
+                        if (scope.data.multiple) {
+                            // TODO: Implement a filter for related containment in djsonb
+                            filterbarController.updateFilter(filterLabel,
+                                                             _.merge({'_rule_type': 'related_containment'},
```
kshepard: I think this should maybe be something like `multiple_containment` instead of `related_containment`. They're all related objects, but some can be multiple (e.g. Vehicles) and some can be singular (e.g. Road Details).

web/test/spec/filterbar/options-directive.spec.js
```diff
@@ -35,11 +35,11 @@ describe('driver.filterbar: Options', function () {
 
     it('should handle restoring filter selection', function () {
         // should have no selection yet
-        expect(Element.find('select').val()).toEqual('');
+        expect(Element.find('select').val()).toEqual(null);
         $rootScope.$broadcast('driver.filterbar:restore', [{'foo#bar': {'_rule_type': 'containment', 'contains': ['baz']}}, null]);
         $rootScope.$digest();
         // should have baz selected now
-        expect(Element.find('select').val()).toEqual('string:baz');
+        expect(Element.find('select').val()).toEqual(['string:baz']);
```
kshepard: Would be good to add a test for the new type of containment as well.

kshepard:
+1

web/app/scripts/filterbar/options-directive.js
moradology: I'm inclined to agree that the name I chose is not adequate - I also don't like `multiple_containment` because both support filtering for multiple values. What are your thoughts?

kshepard: Maybe `multiple_related_containment` and `single_related_containment`?

PR #207, opened 2015-10-16
Feature/stepwise monthly

flibbertigibbet:
* Fix event count aggregation (was comparing date to date string such that count was only ever 0 or 1)
* Aggregate by month rather than by week if have more than 12 weeks of data
* Never show duplicate labels
* Add event period start date to interactivity


kshepard:
+1

PR #217, opened 2015-10-19
Add logic to populate filterbar when first navigated to

moradology:


moradology:
Fixing some tests - I'll reassign when ready for review

web/app/scripts/filterbar/filterbar-controller.js
```diff
@@ -54,9 +61,11 @@
         /**
          * When the record type changes, request the new schema
          */
-        $scope.$on('driver.state.recordstate:selected', function(event, selected) {
-            if (selected) {
+        $scope.$on('driver.state.recordstate:selected',
+                   function(event, selected) { onRecordSelected(selected); });
```
flibbertigibbet: Not sure what's going on here, but a triggered event should send itself.

moradology: onRecordSelected is wrapped so that the first argument isn't needed in the core function (which only uses `selected`)

flibbertigibbet: Yes, but shouldn't the first argument sent by `driver.state.recordstate:selected` still be the event itself, and not `selected`? Looks like it's sending the event as a parameter named `selected` to `onRecordSelected`, instead of a second parameter from `driver.state.recordstate:selected` that should be the `selected` thing.

moradology: I think you might be misreading it:
function(event, selected) { onRecordSelected(selected); });
This will take an `event` (first arg) and `selected` (second arg), apply onRecordSelected with only the second arg and delegate `event` to garbage collection.

flibbertigibbet: Right, nevermind.

flibbertigibbet:
+1 when tests pass

PR #223, opened 2015-10-19
Miscellaneous bugfixes

kshepard:


web/app/scripts/views/map/layers-controller.js
```diff
@@ -146,17 +146,23 @@
...
-                    var layer =  L.geoJson(MapState.getFilterGeoJSON());
+                    var layer = L.geoJson(MapState.getFilterGeoJSON());
                     layer.setStyle(filterStyle);
                     ctl.editLayers.addLayer(layer);
                 }
             });
         };
 
+        // Clears editLayers and informs the map state service
+        function clearLayers() {
```
flibbertigibbet: Might be good to name this to `clearEditLayers` or something, since it doesn't clear all the map layers.

web/app/scripts/navbar/navbar-partial.html
```diff
@@ -23,7 +23,7 @@
                         {{ ctl.recordTypeSelected.plural_label }}
                         <span class="caret"></span></a>
                     <ul class="dropdown-menu">
-                        <li class="nav-rt-item" ng-repeat="recordType in ctl.recordTypeResults | orderBy:'label'">
+                        <li class="nav-rt-item" ng-repeat="recordType in ctl.recordTypeResults | orderBy:'plural_label'">
```
flibbertigibbet: I think it might make sense to remove the angular list reorder filter, and just handle results ordering django-side, to speed up the client a bit. If needed, we could send in an ordering parameter instead of relying on a default sort order from the API.

flibbertigibbet:
+1

web/app/scripts/views/map/layers-controller.js
kshepard: Good call, will do.

web/app/scripts/navbar/navbar-partial.html
kshepard: Yeah, I'd be fine with that.

PR #225, opened 2015-10-19
Use static title label for options filters

### LINKED ISSUE #220: Filter name updates to include selections
This will quickly break the layout. Let's just keep the name of the filter, and append a class to it so that we can style it slightly differently if something's actively selected.

kshepard:
Also adds an `active` class on the div for styling purposes.
Currently the text is underlined when active as a proof-of-concept.

Fixes #220


flibbertigibbet:
+1

PR #229, opened 2015-10-20
Feature/add location really

fungjj92:


web/app/scripts/views/record/add-edit-controller.js
```diff
@@ -46,7 +46,7 @@
             $scope.$on('driver.views.record:map-moved', function(event, data) {
                 bbox = data;
             });
-
+
```
flibbertigibbet: Some whitespace to remove here

flibbertigibbet:
+1

PR #234, opened 2015-10-20
Add service for storage of records pulled down to prevent redundancies

moradology:
This PR does a couple of things.

First, the `RecordState` is now `RecordTypeState` - thus bringing the service name in line with what it *actually* holds the state of.
Second, `RecordState` is now a service for storing the most recent record request and the parameters used for requesting it (this allows us to cut down on requests - I'm seeing two fewer requests on the dashboard, for instance).

### DIFF
--- a/web/app/index.html
+++ b/web/app/index.html
@@ -94,6 +94,7 @@
     <script src="scripts/state/filters-service.js"></script>
     <script src="scripts/state/mapstate-service.js"></script>
     <script src="scripts/state/recordstate-service.js"></script>
+    <script src="scripts/state/recordtypestate-service.js"></script>
     <script src="scripts/state/boundarystate-service.js"></script>
     <script src="scripts/state/geographystate-service.js"></script>
     <script src="scripts/state/zoom-to-boundary-directive.js"></script>
--- a/web/app/scripts/filterbar/filterbar-controller.js
+++ b/web/app/scripts/filterbar/filterbar-controller.js
@@ -2,7 +2,7 @@
     'use strict';
 
     /* ngInject */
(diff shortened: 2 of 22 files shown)


kshepard:
Looks like this didn't cut down on any of the duplicate requests within the tests. Have you explored why those duplicates may be occurring?

moradology:
Yeah, honestly, I have no idea why the tests don't have fewer requests. For sure though, we are sending out fewer requests - take a look at the log that's sent off whenever a request is saved

ddohler:
I think someone had mentioned maybe having the services track inflight requests and then providing back the same promise for all of them; not sure if that's the problem or whether it would help.

moradology:
Yeah, that's a strategy the state objects all try to implement (this one is slightly different in that the parameters used to generate the query need to match for it to save a redundant call). Promises for in-flight requests *should* be sent back in this case too, provided that condition is met

flibbertigibbet:
I assume it's at least partially related to records being loaded before filters come in, then getting re-loaded when filtered.

moradology:
That sounds exactly right.

kshepard:
I'm not seeing any fewer requests when I test this locally. 'saved a request' is being printed, but that's outside the if block where the cached records are being used. I added the print statement to an else block and it's no longer being called.

moradology:
D'oh. I'll figure this out first thing in the morning - guessing there's a problem with how I'm checking equality there.

moradology:
This should be good to go - it *now* appears to be saving quite a few requests as you navigate around the app.

kshepard:
+1

moradology:
You know, that buggy code saved a lot more requests than this functional code does - maybe a change of strategy is in order...

PR #243, opened 2015-10-21
Revert to sending redundant requests

moradology:
Reverts WorldBank-Transport/DRIVER#234


moradology:
This is a response to there being a few tiny bugs which - though superficial - reared their heads at the wrong time. Presumably some modified version of this PR will be re-added later.

moradology:
Merging back to a previously working state

PR #248, opened 2015-10-22
Add django-redis configuration.

flibbertigibbet:
Connect to same redis instance used by Windhsaft.


ddohler:
+1 -- I like the idea of going with a max number of entries plus culling rather than a timeout.

flibbertigibbet:
To connect to the same redis db from within the driver-app container:
```
redis-cli -h 192.168.11.101 -n 1
```

PR #249, opened 2015-10-22
Move details section to top of editor form

flibbertigibbet:


flibbertigibbet:
Tests pass locally. Not sure what's going on with the failures here.

ddohler:
Have you looked into inserting the `propertyOrder` keyword into the Details property in the record schema? I think it would probably need to happen here: https://github.com/WorldBank-Transport/DRIVER/blob/develop/schema_editor/app/scripts/views/recordtype/add-controller.js#L47

flibbertigibbet:
Not being part of what current `json-editor` structure expects, would it respect it?

ddohler:
It may; `propertyOrder` is a custom keyword [defined by json-editor](https://github.com/jdorn/json-editor#property-ordering) so it will hopefully work although json-editor can sometimes be fiddly about such things.

flibbertigibbet:
It only gets added to the subsections, not the parent sections at the level of `Details`; none of the other sections have `propertyOrder` defined.

flibbertigibbet:
Yay, the editor does respect `propertyOrder` on the parent items, even though it doesn't write anything for that property. It must use the default value of 1000 internally only.

ddohler:
+1

flibbertigibbet:
@kshepard I expect we will need the data loader to set `propertyOrder` on the `Details` section when it creates the schema as well; you might want to put that into the new data loader.

kshepard:
Good call, I will do so.

kshepard:
Interestingly the propertyOrder doesn't seem to work its magic on the other properties. I have Accident Details, Person, and Vehicle, and no matter what, I can't get Vehicle to come before Person.

flibbertigibbet:
Well, @jbranigan says he's only interested in pushing Details to the top at the moment. Good to know, though.

kshepard:
Yeah, it's going to be annoying for data input though, since Person has a reference to Vehicle, so vehicles need to be entered first.

flibbertigibbet:
Ah, good point.

PR #250, opened 2015-10-23
Add loader script for v3 of the schema

kshepard:
This script loads data from multiple CSV files that are in the same format. There are currently four sources of data.


ddohler:
+1

PR #251, opened 2015-10-23
Make record type dropdown visibility configurable

kshepard:
A web config setting of `recordType.visible` controls visibility.
If the record type is configured to not be visible, the record type
is chosen by using the config setting of `recordType.primaryLabel`.

A seting for `recordType.secondaryLabel` has also been added, though
it is not yet in use. It will be used when displaying interventions.

### DIFF
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -34,5 +34,8 @@ redis_port: 6379
 web_js_nominatim_key: ""
 web_js_html5mode: "true"
 web_js_html5mode_prefix: "!"
+web_js_record_type_visible: "false"
+web_js_record_type_primary_label: "Accident"
+web_js_record_type_secondary_label: "Intervention"
 editor_js_html5mode: "true"
 editor_js_html5mode_prefix: "!"
--- a/web/app/scripts/navbar/navbar-controller.js
+++ b/web/app/scripts/navbar/navbar-controller.js
@@ -3,7 +3,7 @@
 
     /* ngInject */
     function NavbarController($log, $window, $rootScope, $scope, $state,
-                              GeographyState, RecordState, BoundaryState) {
+                              GeographyState, RecordState, BoundaryState, WebConfig) {
         var ctl = this;
         var _ = $window._;
         init();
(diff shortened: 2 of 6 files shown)


deployment/ansible/group_vars/all.example
```diff
@@ -34,5 +34,8 @@ redis_port: 6379
 web_js_nominatim_key: ""
 web_js_html5mode: "true"
 web_js_html5mode_prefix: "!"
+web_js_recordType_visible: "false"
```
flibbertigibbet: The mix of camel and snake case seems strange; might be good to rename for consistency

web/app/scripts/navbar/navbar-controller.js
```diff
@@ -13,6 +13,7 @@
         ctl.onStateSelected = onStateSelected;
         ctl.navigateToStateName = navigateToStateName;
         ctl.getBoundaryLabel = getBoundaryLabel;
+        ctl.recordTypesVisible = WebConfig.recordType.visible;
```
flibbertigibbet: Shouldn't this be `display` here?

web/test/mock/config.js
```diff
@@ -18,6 +18,11 @@
         },
         record: {
             limit: 50
+        },
+        recordType: {
+            visible: false,
```
flibbertigibbet: Should be `display` here too (or change `web-config-js` template to use `visible`

web/app/scripts/state/recordstate-service.js
```diff
@@ -78,13 +78,21 @@
...
         function setSelected(selection) {
             if (!initialized) {
-                var oldRecordType = localStorageService.get('recordtype.selected');
-                if (oldRecordType) {
+                // If the record type is not configured as visible, always use the primary.
+                // The primary is obtained by finding a record type label matching the primaryLabel.
+                if (!WebConfig.recordType.visible) {
                     selection = _.find(options, function(d) {
-                        return d.uuid === oldRecordType.uuid;
+                        return d.label === WebConfig.recordType.primaryLabel;
```
flibbertigibbet: Does the `uuid` property not need to be set here?

deployment/ansible/group_vars/all.example
kshepard: Good call, I'll rename.

web/test/mock/config.js
kshepard: Oops: I renamed it from display to visible before pushing, but forgot to change the one reference.

flibbertigibbet:
+1

PR #252, opened 2015-10-23
Feature/form error feedback

flibbertigibbet:
* Modify notifications to alternatively display sanitized HTML
* Disable form save when missing constant field(s)
* Should validation errors from any field make it past 'save' press, display list as HTML notification
* Display non-validation errors in notification, like so:

image


kshepard:
+1 on Travis success.

PR #254, opened 2015-10-26
Feature/show required

flibbertigibbet:
Depends on azavea/json-editor#1.

* Sets a `required` class on labels for required fields (here, changes the font color for those from grey to black; will be easy to change if we want something else).

* Adds standard error class and 'Value required' message on the constant fields after form edit, to match behavior of non-constant fields.


kshepard:
+1

Issue #228, opened 2015-10-20
Region picker showing duplicates


jbranigan:
I checked the source file - there's one polygon per region, but the interface is showing about 20.

flibbertigibbet:
Apparently related to state. Pick something for all the available region/sub-region combinations and reload the page, and the duplicates are gone.

PR #255, opened 2015-10-27
Add service for tracking initial state

### LINKED ISSUE #198: BoundaryState not updating when GeographyState changes
If the GeographyState is changed to select a new geometry, the Boundary dropdown will go blank, implying that no Boundary is selected, but calls to `BoundaryState.getSelected()` will still return the previously selected Boundary, if there was one.

kshepard:
All pages of the app rely on the record type, geography, and boundary.
This adds a central place for keeping track of that initialization.
Previously, all components were individually requesting data on initial
page load, which led to many duplicate requests and race conditions.
This eliminates the duplicate requests and fixes a couple bugs.

Fixes #198, #228

### DIFF
--- a/web/app/scripts/navbar/navbar-controller.js
+++ b/web/app/scripts/navbar/navbar-controller.js
@@ -3,10 +3,12 @@
 
     /* ngInject */
     function NavbarController($log, $window, $rootScope, $scope, $state,
-                              GeographyState, RecordState, BoundaryState, WebConfig) {
+                              BoundaryState, GeographyState, InitialState,
+                              RecordState, WebConfig) {
         var ctl = this;
-        var _ = $window._;
-        init();
+        var initialized = false;
+
+        InitialState.ready().then(init);
         ctl.onGeographySelected = onGeographySelected;
         ctl.onBoundarySelected = onBoundarySelected;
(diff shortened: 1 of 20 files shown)


web/app/scripts/navbar/navbar-controller.js
```diff
@@ -3,10 +3,12 @@
 
     /* ngInject */
     function NavbarController($log, $window, $rootScope, $scope, $state,
-                              GeographyState, RecordState, BoundaryState, WebConfig) {
+                              BoundaryState, GeographyState, InitialState,
+                              RecordState, WebConfig) {
         var ctl = this;
-        var _ = $window._;
-        init();
+        var initialized = false;
```
flibbertigibbet: Just checking: the deletion of the underscore redefinition was intentional? (Not sure why it was there.)

flibbertigibbet:
:+1:

web/app/scripts/navbar/navbar-controller.js
ddohler: Yeah I think it can be deleted https://github.com/WorldBank-Transport/DRIVER/pull/77#discussion_r37782812

PR #257, opened 2015-10-27
Feature/date range style

moradology:
This should remove the time picker. For good measure, I also updated the style a bit (after reading that ranges are supported in bootstrap timepicker) and set up a default window for filtering (the last 90 days).
screen shot 2015-10-27 at 11 59 06 am 2


web/app/index.html
```diff
@@ -8,7 +8,7 @@
     <!-- Place favicon.ico and apple-touch-icon.png in the root directory -->
     <!-- build:css(.) styles/vendor.css -->
     <!-- bower:css -->
-    <link rel="stylesheet" href="bower_components/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.min.css" />
+    <link rel="stylesheet" href="bower_components/bootstrap-datepicker/dist/css/bootstrap-datepicker3.css" />
```
flibbertigibbet: Date/time picker is still in use for adding and editing records. I imagine it will need its CSS for that. Please confirm that it's still working there ok.

moradology: Oh, weird. I kept it in bower, so I didn't even question that. Things appear to be operating correctly on review: 
screen shot 2015-10-27 at 12 12 31 pm 2

flibbertigibbet: Your docker image might have some things cached. I'd expect we'd need to keep all the references to it.

flibbertigibbet:
+1

PR #259, opened 2015-10-28
Filter dates with an inclusive range

moradology:


web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -65,9 +65,12 @@
                  * @param filterLabel {string} label of which field to filter
                  * @param filterObj {object} filter data
                  */
-                scope.updateFilter = function(filterLabel, filterObj) {
+                scope.updateFilter = function() {
+                    var day = moment.duration({days: 1});
+                    var inclusiveRange = _.clone(dtRange);
+                    inclusiveRange.max = new Date(moment(new Date(inclusiveRange.max)) + day);
```
moradology: This is slightly ugly, but calling new Date() prior to moment prevents an error from being belched up within moment.js

kshepard:
+1

PR #260, opened 2015-10-28
Make windshaft and client-side code use tilekey

### LINKED ISSUE #213: Filter maps by the selected boundary
See note in #201 -- this will probably need to be held off until after we implement the Windshaft changes.

kshepard:
The client-side code sends the `tilekey` parameter, and Windshaft
uses it to obtain the stored SQL from redis. Specifying sql directly
from the client is no longer possible. The `polygon_id` parameter
is also now being sent along with the request, so filtering map point
based on the selected region now works.

Also fixes a problem with boundaries not showing up on dev instances.

Fixes #213

### DIFF
--- a/web/app/scripts/map-layers/recent-events/recent-events-layer-directive.js
+++ b/web/app/scripts/map-layers/recent-events/recent-events-layer-directive.js
@@ -39,7 +39,6 @@
                     leafletController.getMap().then(updateLayers);
                 });
 
-                // TODO: Remove when the record type picker is removed.
                 scope.$on('driver.state.recordstate:selected', function() {
                     leafletController.getMap().then(updateLayers);
                 });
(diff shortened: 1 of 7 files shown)


web/app/scripts/map-layers/recent-events/recent-events-layer-directive.js
```diff
@@ -74,25 +73,20 @@
...
+                return BoundaryState.getSelected().then(function(boundary) {
+                    return QueryBuilder.unfilteredDjangoQuery(0, {
+                        tilekey: true,
+                        /* jshint camelcase: false */
+                        occurred_min: occurredMin.toISOString(),
+                        polygon_id: boundary.uuid
+                        /* jshint camelcase: true */
+                    }).then(function(result) {
+                        var tilekeyParam = (baseUrl.match(/\?/) ? '&' : '?') + 'tilekey=';
+                        return baseUrl + tilekeyParam + result.tilekey;
```
flibbertigibbet: I suppose our tile keys don't need / would be broken by URI encoding?

windshaft/server.js
```diff
@@ -20,8 +30,7 @@ var config = {
         base_url_notable: '/tiles/table/:tablename',
         req2params: function(req, callback) {
             try {
-                req.params = driver.getRequestParameters(req);
-                callback(null, req);
+                driver.setRequestParameters(req, callback, redisClient);
```
flibbertigibbet: Nice async fix.

flibbertigibbet:
:+1:

web/app/scripts/map-layers/recent-events/recent-events-layer-directive.js
kshepard: Yeah, they're just UUIDs (numbers, letters, hyphens), so it didn't seem necessary.

Issue #209, opened 2015-10-16
Upon loading map page, it doesn't zoom to the selected polygon


kshepard:
To reproduce:
 * Load dashboard page
 * Select polygon
 * Browse to map -- it doesn't zoom to the selected polygon

ddohler:
This is intentional, although I agree it's a bit awkward: The map page keeps track of its most recent location and zoom, and we didn't want selecting a polygon on a different page to override that. I actually had implemented functionality to zoom on page load, but removed it after @flibbertigibbet pointed out that it would make the map's location-memory feature redundant. This might be a good thing for @jbranigan and @mtedeschi to look at because it seems like the functionality may be pulling user expectations in different directions.

kshepard:
Oh, ok, that's a good point. Yeah, we'll have to talk through a good approach, as it seemed a bit unexpected at first.

jbranigan:
Changing the primary geography should actually update the map state.

flibbertigibbet:
My concern with state vs zooming to region is that it should only do one or the other and not both upon loading, depending on which is appropriate.

jbranigan:
In the wireframes, a user logs in and chooses their primary geography before even hitting the dashboard page. It's kind of a constant theme across the app. I don't understand where the problem is.

PR #261, opened 2015-10-28
Fix issue #209

fungjj92:
Map now initializes to selected boundaries. Maintains map memory switching between record states.


kshepard:
This is part of the fix, but it also needs to cover the scenario where when the user selects a new boundary from the dropdown while not on the map and then browses back to the map, the new region should be shown (not the previously saved map state).

Also, would you be able to amend your commit message to explain what's being fixed? It'll make it easier to quickly be able to tell at a glance what's happening in the commit without needed to go look up the referenced issue. You can put `Fixes #209` on a separate line in the commit message.

PR #266, opened 2015-10-29
Fix #237 numeric range btn styling

### LINKED ISSUE #237: Style numerical filters and change label
They're currently showing up with the record type + # prefix. Also the min/max should probably exist on the same line (or rounded corners removed).

fungjj92:
Restyle num range button to reflect date range picker and other buttons

screen shot 2015-10-29 at 3 25 41 pm

### DIFF
--- a/web/app/scripts/filterbar/numeric-range.html
+++ b/web/app/scripts/filterbar/numeric-range.html
@@ -1,12 +1,12 @@
-<a id="record-type"
+<button id="record-type"
    role="button"
    ng-class="error.btnClasses"
-   class="ng-binding dropdown-toggle"
+   class="ng-binding btn btn-default dropdown-toggle"
    data-toggle="dropdown"
    aria-haspopup="true"
    aria-expanded="true">
   {{ label | labelFormatter }} <span class="caret"></span>
-</a>
+</button>
 <ul class="dropdown-menu alert"
     ng-class="error.classes"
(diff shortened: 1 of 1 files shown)


web/app/scripts/filterbar/numeric-range.html
```diff
@@ -19,20 +19,23 @@
...
                        name="minimum"
-                       placeholder="Minimum"
                        class="form-control"
+                       placeholder="Min"
                        min="{{ data.minimum }}"
                        max="{{ data.maximum }}"
                        ng-model="filter.min"
-                       ng-change="updateFilter(label, filter)">
+                       ng-change="updateFilter(label, filter)"
+                       style="width:80px">
```
kshepard: We try to avoid using in-line styles on HTML elements. Would you be able to move these two in-line styles to a Sass file?

kshepard:
+1 after removing in-line styling.

web/app/scripts/filterbar/numeric-range.html
fungjj92: It was in date-range.html so I thought it was okay. Should that be edited too?

kshepard: Ah, yeah, looks like that one also snuck in. Yes, if you don't mind, move that one out too.

PR #271, opened 2015-10-30
Feature/section ordering

flibbertigibbet:
* Assign section order in the `$schema` JSON to put `Vehicle` above `Person`
* Use tagged release of json-editor fork so containers will update their version without being destroyed


kshepard:
Can the other `propertyOrder` attributes (the ones on the root objects) be removed now?

kshepard:
+1

flibbertigibbet:
The other `propertyOrder` attributes don't hurt, and at least for the case of the details section, it does work in the old location, so I'm inclined to leave them both in, at least for the details. I suppose I might as well remove the other two.

PR #273, opened 2015-10-30
Fix various state issues

kshepard:
This addresses several state issues, particularly
around maintaining the state of filters when navigating
between pages and on page refreshes.

This also wires up the toddow to the boundary picker, and
also uses filters on the map page.

Please pull down this branch and test it out to make sure the state is looking good.

### DIFF
--- a/web/app/scripts/resources/query-builder-service.js
+++ b/web/app/scripts/resources/query-builder-service.js
@@ -101,31 +101,41 @@
             var paramObj = {};
             /* jshint camelcase: false */
             if (doFilter) {
-                // An exceptional case for date ranges (not part of the JsonB we filter over)
+                // An exceptional case for date ranges (not part of the JsonB we filter over).
+                // If no dates are specified, the last 90 days are used.
+                var now = new Date();
+                var duration = moment.duration({ days:90 });
+                var maxDateString = now.toJSON().slice(0, 10);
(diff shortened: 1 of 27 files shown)


web/app/scripts/resources/query-builder-service.js
```diff
@@ -101,31 +101,41 @@
...
                 }
 
                 var jsonFilters = svc.assembleJsonFilterParams(_.omit(FilterState.filters, '__dateRange'));
 
                 // Handle cases where no json filters are set
                 if (!_.isEmpty(jsonFilters)) {
-                    paramObj = _.extend(paramObj, {'jsonb': jsonFilters});
+                    paramObj = _.extend(paramObj, { jsonb: jsonFilters });
                 }
+                paramObj.limit = WebConfig.record.limit;
```
flibbertigibbet: Might there be any places that use this that expect a non-paginated result, like the map?

flibbertigibbet:
:+1:

web/app/scripts/resources/query-builder-service.js
kshepard: Yeah, I agree -- this parameter was already being set here, but I don't think it's a good thing to have by default. It caused problems with the toddow, so I needed to delete it in another section of code. Having it here isn't breaking anything at the moment (map queries and other queries are working fine), so I say we keep it for now, but I'll make an issue to move it out.

PR #281, opened 2015-11-05
Only show records for the last 90 days

moradology:
This adds filtering such that only records from the last 90 days are shown for the TODDOW on the dashboard. It also adds a label to the TODDOW that clarifies this point.


flibbertigibbet:
+1

PR #282, opened 2015-11-05
Use a list of 24 hours rather than 25

moradology:
The extra hour we had recently been seeing was the result of daylight savings time - there were 25 hours last sunday. I've changed the logic to just use 01/01/2001's week (no daylight savings time weirdness) as we only need a canonical week for the chart to work.

PR BONUS ROUND: I also added an offset to the tooltip that should center the tooltip


kshepard:
+1. Wow, I would have never guessed this was related to DST!

PR #284, opened 2015-11-06
Feature/stepwise aggregation endpoint - WIP

moradology:
This PR aims to provide aggregation (with filters) for stepwise data. The client will need to pad missing weeks with 0s

### DIFF
--- a/app/data/tests/test_views.py
+++ b/app/data/tests/test_views.py
@@ -56,6 +56,22 @@ class DriverRecordViewTestCase(APITestCase):
             else:
                 self.assertEqual(toddow['count'], 2)
 
+    def test_stepwise(self):
+        """Test that date filtering is working appropriately and that data is being binned properly
+        """
+        url = ('/api/records/stepwise/?record_type={uuid}&occurred_max={maximum}&occurred_min={minimum}'
+               .format(uuid=str(self.record_type.uuid),
+                       minimum=(self.then - timedelta(hours=1)).isoformat() + 'Z',
+                       maximum=datetime.now().isoformat() + 'Z'))
+
+        response = json.loads(self.client.get(url).content)
+        self.assertEqual(len(response), 2)
+        for step in response:
+            if step['week'] == self.now.isocalendar()[1]:
+                self.assertEqual(step['count'], 1)
+            else:
+                self.assertEqual(step['count'], 2)
+
     def test_arbitrary_filters(self):
         base = '/api/records/toddow/?record_type={rt}&occurred_max={dtmax}Z&occurred_min={dtmin}Z'
 
--- a/app/data/tests/test_views.py
+++ b/app/data/tests/test_views.py
@@ -61,7 +61,7 @@ class DriverRecordViewTestCase(APITestCase):
         """
         url = ('/api/records/stepwise/?record_type={uuid}&occurred_max={maximum}&occurred_min={minimum}'
                .format(uuid=str(self.record_type.uuid),
-                       minimum=(self.then - timedelta(hours=1)).isoformat() + 'Z',
+                       minimum=self.beforeThen.isoformat() + 'Z',
                        maximum=datetime.now().isoformat() + 'Z'))
 
         response = json.loads(self.client.get(url).content)
--- a/app/data/transformers.py
+++ b/app/data/transformers.py
@@ -15,34 +15,4 @@ class WeekTransform(Transform):
         return IntegerField()
 
 
-# Currently unused - potentially useful if we want to aggregate on our stepwise graph by month
-class MonthTransform(Transform):
-    """A custom transform to turn a date field into the integer which represents its month's
-    temporal rank in the year (e.g. January has the value 1)"""
-    lookup_name = 'week'
-
-    def as_sql(self, compiler, connection):
-        lhs, params = compiler.compile(self.lhs)
-        return "Extract('MONTH' FROM %s)" % lhs, params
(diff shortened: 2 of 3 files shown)


moradology:
Writing a couple tests before I remove WIP

app/data/transformers.py
```diff
@@ -0,0 +1,48 @@
...
+        lhs, params = compiler.compile(self.lhs)
+        return "Extract('WEEK' FROM %s)" % lhs, params
+
+    @property
+    def output_field(self):
+        return IntegerField()
+
+
+# Currently unused - potentially useful if we want to aggregate on our stepwise graph by month
+class MonthTransform(Transform):
```
ddohler: What's the difference between this and https://docs.djangoproject.com/en/1.8/ref/models/querysets/#month ? Possibly the same question for the `YearTransform` although the docs indicate that the builtin `__year` may have a different SQL representation.

app/data/transformers.py
```diff
@@ -0,0 +1,48 @@
...
+    """A custom transform to turn a date field into the integer which represents its month's
+    temporal rank in the year (e.g. January has the value 1)"""
+    lookup_name = 'week'
```
ddohler: This should be something else; probably not `month` since that conflicts with the `__month` mentioned above.

app/data/transformers.py
```diff
@@ -0,0 +1,48 @@
...
+        return "Extract('MONTH' FROM %s)" % lhs, params
+
+    @property
+    def output_field(self):
+        return IntegerField()
+
+
+class YearTransform(Transform):
+    """A custom transform to peel the year out from a postgres datetime field"""
+    lookup_name = 'year'
```
ddohler: This one should also probably be renamed to avoid conflicts with builtins.

app/data/views.py
```diff
@@ -29,6 +32,43 @@ def list(self, request, *args, **kwargs):
...
+            end_date = parse_date(request.GET['occurred_max'])
+        except KeyError:
+            return Response(status=HTTP_400_BAD_REQUEST)
+
+        # The min year can't be after or more than 2000 years before the max year
+        year_distance = end_date.year - start_date.year
+        if year_distance < 0 or year_distance > 2000:
+            return Response(status=HTTP_400_BAD_REQUEST)
+
+        qryset = self.get_queryset()
```
ddohler: I'm not sure the reduction in characters is worth the reduction in readability here; maybe use either `qset` or `queryset`.

app/data/transformers.py
```diff
@@ -0,0 +1,18 @@
+from django.db.models import IntegerField, DateTimeField, Transform
+
+
+class WeekTransform(Transform):
```
ddohler: It'd be great to have a test that does something along the lines of this:
```
transformer = WeekTransform(lhs="foo") # Totally just made up the constructor
sql_str = transformer.as_sql(compiler, connection)
self.assertEqual("EXTRACT('WEEK' FROM foo)", sql_str)
```

app/data/views.py
```diff
@@ -29,12 +32,49 @@ def list(self, request, *args, **kwargs):
...
+        try:
+            start_date = parse_date(request.GET['occurred_min'])
+            end_date = parse_date(request.GET['occurred_max'])
+        except KeyError:
+            return Response(status=HTTP_400_BAD_REQUEST)
+
+        # The min year can't be after or more than 2000 years before the max year
+        year_distance = end_date.year - start_date.year
+        if year_distance < 0 or year_distance > 2000:
+            return Response(status=HTTP_400_BAD_REQUEST)
```
ddohler: In both this and the query parameter checking above it might be a good idea to re-raise a `ParseError` with a detail message.

app/data/views.py
```diff
@@ -29,12 +32,49 @@ def list(self, request, *args, **kwargs):
         return response
 
     @list_route(methods=['get'])
+    def stepwise(self, request):
+        """Return an aggregation counts the occurrence of events per week (per year) between
+        two bounding datetimes
+        """
+        # We'll need to have minimum and maximum dates specified to properly construct our SQL
+        try:
+            start_date = parse_date(request.GET['occurred_min'])
```
ddohler: I think `request.query_params` is preferred over `request.GET`, at least in DRF-world.

app/data/views.py
```diff
@@ -29,12 +32,49 @@ def list(self, request, *args, **kwargs):
         return response
 
     @list_route(methods=['get'])
+    def stepwise(self, request):
+        """Return an aggregation counts the occurrence of events per week (per year) between
+        two bounding datetimes
```
ddohler: It'd be good to have a short example of the JSON that is returned here.

app/data/tests/test_views.py
```diff
@@ -56,6 +56,22 @@ def test_toddow(self):
             else:
                 self.assertEqual(toddow['count'], 2)
 
+    def test_stepwise(self):
+        """Test that date filtering is working appropriately and that data is being binned properly
+        """
+        url = ('/api/records/stepwise/?record_type={uuid}&occurred_max={maximum}&occurred_min={minimum}'
+               .format(uuid=str(self.record_type.uuid),
+                       minimum=(self.then - timedelta(hours=1)).isoformat() + 'Z',
```
ddohler: I think you can probably use `self.beforeThen` here.

ddohler:
+1 after a bit of cleanup.

PR #285, opened 2015-11-06
Feature/users groups

flibbertigibbet:
* Enable token authentication for DRF
* Require authentication to access API
* Add user and group models to DRF
* Enable front-end authentication, with login and out for web and editor
* Update django-oauth-toolkit version to fix known issue fetching tokens


app/data/serializers.py
```diff
@@ -0,0 +1,32 @@
...
+
+class UserSerializer(serializers.HyperlinkedModelSerializer):
+    class Meta:
+        model = User
+        fields = ('id', 'url', 'username', 'email', 'groups', 'password',
+                  'is_staff', 'is_superuser',)
+        read_only_fields = ('id',)
+        write_only_fields = ('password',)
+
+    def create(self, validated_data):
```
kshepard: Should this endpoint be locked down to only allow administrators to create users?

schema_editor/app/scripts/views/login/auth-controller.js
```diff
@@ -0,0 +1,47 @@
...
+        };
+
+        $scope.authenticate = function() {
+            $scope.alerts = [];
+            $scope.authenticated = AuthService.authenticate($scope.auth);
+            $scope.authenticated.then(function(result) {
+                if (result.isAuthenticated) {
+                    // redirect to editor main page, with a full page reload,
+                    // to pick up the newly set credentials
+                    $window.location.href = '/editor/';
```
kshepard: This is only the url to editor main page on the built app. In development on port 7001, this url won't exist.

web/app/scripts/navbar/navbar-controller.js
```diff
@@ -27,6 +30,7 @@
         }
 
         $rootScope.$on('$stateChangeSuccess', function(event, toState) {
+            ctl.authenticated = AuthService.isAuthenticated();
```
kshepard: I think it'd be good to perhaps use this variable in the navbar template with an ng-if to not render the dropdowns until the user is authenticated. Otherwise there will be failed xhr requests for those resources and the dropdowns will be empty.

kshepard:
+1, really nice. One other thing that'd be good is if a user could be automatically created during provisioning. Alternatively, instructions can be added to the readme for user creation. Reminder: hold off on merging this until today's demo is over.

schema_editor/app/scripts/views/login/auth-controller.js
flibbertigibbet: Since both endpoints are valid in development, templating the URL via ansible would still break one or the other. It's just the post-login redirect. Running via grunt changes the endpoint to root; if it just changed the port, this would still redirect correctly. Instead of having to insert some sort of ugly port check here just for `grunt serve`, it would be better if `grunt serve` served on the `/editor/` endpoint, which would be consistent.

web/app/scripts/navbar/navbar-controller.js
flibbertigibbet: Ok. I was thinking we'd make a login page without the navbar. Will do.

flibbertigibbet:
There is an admin user created during provisioning, at least in development.

app/data/serializers.py
flibbertigibbet: That makes sense. We can configure the OpenID setup to create new users on first login separately.

PR #288, opened 2015-11-11
Feature/stepwise filtering

moradology:
This changeset encompasses more than indicated by the branch name:
1. Aggregation is now being used for the 'stepwise' graph
2. Our 'stepwise' chart is now a bar chart (to match the func spec description) that dynamically resizes
3. Filters now work as expected

Frustratingly, things are still referred to as 'stepwise'. I've got a branch that's in the process of fixing that.

### DIFF
--- a/app/data/transformers.py
+++ b/app/data/transformers.py
@@ -14,5 +14,18 @@ class WeekTransform(Transform):
     def output_field(self):
         return IntegerField()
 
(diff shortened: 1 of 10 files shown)


moradology:
Note also that this encompasses the two cards I currently have out - hovering and filtering

kshepard:
:+1:

PR #291, opened 2015-11-11
Ensure that the RVM bin path is added to PATH

hectcastro:
Something in the Travis CI image is preventing this from happening.


flibbertigibbet:
+1

PR #293, opened 2015-11-12
Minor map changes

kshepard:
Temporarily use crossorigin.me to work around a CORS problem with pickpoint.io.

Also, while I was in there, I switched out the viewbox query param sent to pickpoint.io with countrycodes. Setting a country code seems to do a much better job of limiting the query.

Also changed the location selection zoom to zoom in a few more levels (a more reasonable default when someone is looking to drop a point on a specific road location).

### DIFF
--- a/web/app/scripts/nominatim/nominatim-service.js
+++ b/web/app/scripts/nominatim/nominatim-service.js
@@ -4,7 +4,8 @@
     /* ngInject */
     function Nominatim($http, WebConfig) {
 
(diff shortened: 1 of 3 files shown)


flibbertigibbet:
+1

It might make sense to filter by bounding box in addition to country code, if the intention is to only show things in the current view. The combination of the two might work better than bounds alone.

kshepard:
Cool, I can try that out.

PR #298, opened 2015-11-13
Add details modal and use on record list view

kshepard:


web/app/scripts/views/record/details-modal-partial.html
```diff
@@ -0,0 +1,24 @@
+<div class="incident-report modal-content">
+    <div class="modal-content">
+        <div class="close" ng-click="modal.close()">
+            &times;
+        </div>
+        <div class="report-header">
+            <a ui-sref="record.details({ recorduuid: modal.record.uuid })" target="_blank">
```
flibbertigibbet: Should the record type UUID also be passed here?

flibbertigibbet:
Code looks good.

web/app/scripts/views/record/details-modal-partial.html
kshepard: No, the record type UUID is no longer obtained from the state. This was implemented a while ago, but looks like it was never removed from the ui-sref link.

PR #299, opened 2015-11-16
Add link to detail page 

moradology:
This PR introduces some minor changes to the popup on the map screen - ready for review

### DIFF
--- a/web/app/scripts/state/mapstate-service.js
+++ b/web/app/scripts/state/mapstate-service.js
@@ -45,11 +45,11 @@
         }
 
         /**
-         * Get zoomlevel
+         * Get zoomlevel - default to 5
          */
         function getZoom() {
-            $log.debug('Getting stored map zoom...');
-            return zoom;
+            $log.debug('Getting stored map zoom: ' + JSON.stringify(zoom));
+            return zoom || 5;
         }
 
         /**
@@ -65,7 +65,7 @@
          * Get coordinates for the center of the map
          */
         function getLocation() {
-            $log.debug('Getting stored map location...');
+            $log.debug('Getting stored map location: ' + JSON.stringify(location));
             return location;
 
         }
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -2,13 +2,14 @@
     'use strict';
 
     /* ngInject */
-    function DriverLayersController($q, $log, $scope, $rootScope, $timeout,
-                                    WebConfig, FilterState, RecordState, GeographyState,
-                                    BoundaryState, Records, QueryBuilder, MapState,
-                                    TileUrlService, InitialState) {
+    function DriverLayersController($q, $log, $scope, $rootScope, $timeout, $compile,
+                                    FilterState, RecordState, GeographyState,
+                                    RecordSchemaState, BoundaryState, QueryBuilder,
+                                    MapState, TileUrlService, InitialState) {
         var ctl = this;
 
         ctl.recordType = 'ALL';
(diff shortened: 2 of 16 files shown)


web/app/scripts/state/mapstate-service.js
```diff
@@ -45,11 +45,11 @@
...
 
         /**
-         * Get zoomlevel
+         * Get zoomlevel - default to 5
          */
         function getZoom() {
-            $log.debug('Getting stored map zoom...');
-            return zoom;
+            $log.debug('Getting stored map zoom: ' + JSON.stringify(zoom));
+            return zoom || 5;
```
moradology: I'm not sure how travis was passing before, but without this, map state wasn't being reset properly on return to the map page

web/app/scripts/views/map/layers-controller.js
```diff
@@ -340,39 +358,38 @@
...
-            var startingUnderscoreRegex = /^_/;
 
             // add header with event date constant field
             /* jshint camelcase: false */
-            var str = '<div class="record-popup">';
-            str += '<div><h3>Occurred on: ' + record.occurred_from + '</h3>';
+            var str = '<div id="record-popup" class="record-popup">';
+            str += '<div><h3>' + ctl.recordTypeLabel + '</h3>';
+            // The ng-click here refers to a function which sits on the map-controller's scope
+            str += '<p>Date: <a ng-click="showDetailsModal(\'' + record.uuid + '\')">' +
```
kshepard: This can use some css to change the hover cursor to a pointer.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -340,39 +358,38 @@
...
+                  return _.map(filterPathArray, function(filterPath) { return filterPath[1]; });
+              }).value();
+
+            // Now, we need to parse the record data and crawl through it to generate a quick view
+            //  of all the filterable data on this record
+            _(propsToShow).keys().each(function(key) {
+                str += '<h4>' + key + '</h4>';
+                if (data.hasOwnProperty(key)) {
+                    _.each(propsToShow[key], function(val) {
+                        var toPrint = data[key][val] || 'UNDEFINED';
```
kshepard: I think if the value doesn't exist, we should just skip displaying that row. In my testing, most of my records are showing a bunch of UNDEFINEDs, and I think it'd look better if they were left out.

kshepard:
Something weird is happening with the detail popups on the map page. If you open a modal for the same record on both the map page and the list page to compare them, the one on the map page is missing some information. Only the `constant` fields are showing up, and the others aren't (in the case of the record I'm viewing, it should also be showing the severity and description). In addition, the map doesn't zoom in to the location as it should.

kshepard:
+1 when that issue is worked out.

web/app/scripts/views/map/layers-controller.js
moradology: We already talked about this but some record might make sense - I think it might be useful for Mike to get a look at what the data we're presenting actually might look like so that he can come up with a solution (either the one you've suggested or another to illustrate what's missing)

kshepard: Interestingly, I'm reviewing/testing Mike's open styling PR now, and it looks like it includes his solution for this. It only shows the date, and then buttons for viewing and editing. So the additional information isn't needed in this window at all.

kshepard: I'm going to merge his PR now. Would you be able to rebase, and wire up the buttons? It seems the logic for building up the HTML will also be able to be removed.

PR #300, opened 2015-11-18
Feature/ganalytics

moradology:
This adds a little snippet to index.html and templates it out for the google analytics ID.


moradology:
I'm getting some strange programming_error exception during provisioning in the django migrations step... Unclear why, looking into it.

kshepard:
Is it: `"skip_authorization" of relation "oauth2_provider_application" already exists`? I just provisioned for the first time in a few days and got the same error.

moradology:
Indeed - is there a working solution?

kshepard:
Not sure what the problem is: I just ran `ALTER TABLE oauth2_provider_application DROP skip_authorization;` and reran migrations as a quick workaround, but there's likely a fundamental problem that should be fixed.

flibbertigibbet:
I ran into this first when setting up the authorization stuff, though oauth2 was already in the app. It's apparently a known issue for the oauth2 library. Running migrations with `--fake` before running again (without `--fake`) will work around the issue.

flibbertigibbet:
Provisioning a fresh DB will not have the issue.

deployment/ansible/roles/driver.web/templates/index.html.j2
```diff
@@ -22,6 +22,12 @@
     <!-- endbuild -->
   </head>
   <body ng-app="driver">
+
+    <!-- Google Tag Manager -->
+    <noscript><iframe src="//www.googletagmanager.com/ns.html?id={{ google_analytics_id }}"height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-5B5G65');</script>
```
flibbertigibbet: Hardcoded ID towards the end of this snippet

flibbertigibbet:
Should add comment to head of templated `index.html` about it being written by ansible and will be overwritten by reprovisioning.

flibbertigibbet:
Could we maybe include a partial or some such in the `index.html` with the analytics snippet and template that out instead? I'm thinking that might be preferable, since `index.html` is something people are going to go edit whenever they're adding or changing something to include, and it would be easy to mistakenly edit the wrong thing.

moradology:
That's a good thought - this sort of thing has caused trouble in previous projects.

moradology:
I did a little bit of digging around and it seems as though even the angular libraries for incorporating GA (most of which appear to not be well maintained) seem to expect your GA snippet to be loaded before other JS. It also looks like we very, very rarely end up making changes directly on index.html so - while I'm inclined to agree this could be a headache for someone at some point, I think it will be rare.

flibbertigibbet:
Ok. +1 after adding template warning comment about ansible overwrite

PR #303, opened 2015-11-19
Correct transformation for multiselect checkboxes

sharph:
This commit also requires a change in ashlar to validate against json
schema draft 3 instead of draft 4.

PR in ashlar coming soon.


ddohler:
+1 after adding a a test for the corrected transformation.

I think it would be good to also open an issue for fixing this in in our fork of JSON-Editor so that we can return to using draft 4.

sharph:
I don't know that there is a good way to represent "multiselect" within the constraints of draft 4, otherwise that is the solution I would prefer.

PR #304, opened 2015-11-19
Add backend and API for storing saved filters

ddohler:
Sets up API views and a backend model to support saved filters. The filters are stored in Postgres as JSON, which is slightly more space-efficient than JSONB, but slower to query. I think this is a reasonable tradeoff since we don't plan to query the contents of filters as far as I know, and it makes serialization and deserialization easy since those work exactly the same as JSONB fields.

### DIFF
--- a/app/user_filters/models.py
+++ b/app/user_filters/models.py
@@ -1,0 +1,13 @@
+import uuid as uuidlib
+
+from django.contrib.auth.models import User
+from django.db import models
+from djsonb import fields as jfields
+
+
+class SavedFilter(models.Model):
+    """A filter saved by the user for later reference"""
+    uuid = models.UUIDField(primary_key=True, default=uuidlib.uuid4, editable=False)
+    label = models.CharField(max_length=255)
+    owner = models.ForeignKey(User)
+    filter_json = jfields.JsonField()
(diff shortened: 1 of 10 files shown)


moradology:
It should also be faster to write as JSON - I think this choice makes perfect sense.

app/user_filters/models.py
```diff
@@ -0,0 +1,13 @@
...
+from django.db import models
+from djsonb import fields as jfields
+
+
+class SavedFilter(models.Model):
+    """A filter saved by the user for later reference"""
+    uuid = models.UUIDField(primary_key=True, default=uuidlib.uuid4, editable=False)
+    label = models.CharField(max_length=255)
+    owner = models.ForeignKey(User)
+    filter = jfields.JsonField()
```
kshepard: Just going to note this: `filter` is a built-in function in Python, so this is going to override its definition and could potentially cause confusion if that function was going to be used in this scope. Odds of that happening are probably pretty low however.

kshepard:
+1

app/user_filters/models.py
moradology: Probably best not to write over reserved words though, in general?

moradology:
Thanks for jumping on this, @kshepard

app/user_filters/models.py
kshepard: Yeah, it's technically not a reserved word, just a built-in function, so the language will allow it, but it still is good practice to not override them.

ddohler: Oh good call; I was considering other names for this anyway, so I'll make the change.

PR #306, opened 2015-11-20
Filter map graphs with user-drawn polygons

ddohler:
Filters the toddow and bar graphs on the map using a user-drawn polygon, if there is one. Also fixes a problem with date formatting that was preventing the bar graph from showing up on FireFox.

screenshot from 2015-11-20 11 21 45
screenshot from 2015-11-20 11 21 05


flibbertigibbet:
Does this work with the selected region polygon filter as well?

+1 when tests pass

ddohler:
Yep, it works with selected regions too:
screenshot from 2015-11-20 13 59 04

Looking into the test failures, will merge when once I figure it out.

PR #310, opened 2015-11-20
Fix reference to boundary polygon mock

kshepard:


kshepard:
The tests were failing because of an undefined object that was being passed into `$httpBackend.whenGET`. I believe that it started failing suddenly, because the http call in the test was never actually being made, and something ancillary was fixed that caused the http request to start being made again. `whenGET` doesn't actually fail the test if the http request isn't made, so the undefined object wasn't causing an issue. A key takeaway is that we should always prefer to use `expectGET` as it's more precise and won't allow errors like this to lurk around.

Going to merge this as its a trivial change and passed tests on Travis.

flibbertigibbet:
I think that test may have been written before there was a default boundary, so depending on the test, so there may have not been a response for that request.

kshepard:
Cool, that could be it!

PR #312, opened 2015-11-23
Remove OAuth2

flibbertigibbet:


flibbertigibbet:
Companion to azavea/ashlar#63

ddohler:
+1. Do you know how the python module for `oauth2_provider` was being installed? I was expecting to see it deleted from a `requirements.txt` file somewhere, but it's not there. It might be good to confirm that `from oauth2_provider.models import Application` fails after rebuilding the app container.

flibbertigibbet:
Good call.

PR #313, opened 2015-11-24
Add saved filters modal and dashboard widget

kshepard:


web/app/scripts/saved-filters/saved-filters-modal-controller.js
```diff
@@ -0,0 +1,63 @@
...
+
+    /* ngInject */
+    function SavedFiltersModalController($modalInstance, $rootScope,
+                                         FilterState, Notifications, SavedFilters) {
+        var ctl = this;
+        ctl.label = '';
+        ctl.save = save;
+
+        // linter complains if this function is defined below for some reason
+        ctl.close = function() { $modalInstance.close(); };
```
ddohler: That's weird; if it was defined the same ways as `save` is I don't see why there would be a problem. Does changing the name of the function (maybe `closeModal`) help?

ddohler:
+1, great.

web/app/scripts/saved-filters/saved-filters-modal-controller.js
kshepard: Yeah, totally bizarre. I'll try changing the name to see what happens.

kshepard: Works fine with a different name. Guess I'll go with that!

PR #322, opened 2015-11-24
Feature/setup google oidc

flibbertigibbet:
Basic setup for Google OIDC login.

With this, clicking the 'Google' button on the form at `/openid/login` will automatically create a user after authenticating, and log the new user in with session authentication, which will work for accessing the DRF pages as the new user.

This doesn't integrate yet with the Angular frontend login and its token authentication. Another thing to note is that the user ID that gets created is a random number and not the user's email address (which is stored on the email field.)


kshepard:
+1, nice start!

flibbertigibbet:
Looks like the username number actually comes from Google.

PR #324, opened 2015-11-27
Feature/integrate oidc frontend

flibbertigibbet:
Also refactors rest of the auth-related things to the new auth sub-project.


kshepard:
+1, nice.

PR #327, opened 2015-11-30
Use hostname instead of IP address for API.

flibbertigibbet:
For SSO, which does not allow IP addresses in redirect.


flibbertigibbet:
Trivial, merging.

PR #328, opened 2015-12-01
Feature/show acct info

flibbertigibbet:
Display user details on account page.

Also display user email on navbar drop-down.

image


web/app/scripts/navbar/navbar-controller.js
```diff
@@ -19,6 +19,8 @@
         ctl.navigateToStateName = navigateToStateName;
         ctl.getBoundaryLabel = getBoundaryLabel;
         ctl.recordTypesVisible = WebConfig.recordType.visible;
+        ctl.userEmail = 'My Account';
```
kshepard: Since this default string is used twice, it should be shifted to a variable so if we want to change it we only need to change it in one place.

kshepard:
+1

PR #329, opened 2015-12-01
Apply saved filters

sharph:
Fix filter selection drop down to allow filter loading multiple times

### DIFF
--- a/web/app/scripts/saved-filters/saved-filters-controller.js
+++ b/web/app/scripts/saved-filters/saved-filters-controller.js
@@ -27,8 +27,10 @@
         }
 
         // Activates the selected filter
-        function viewFilter(/*savedFilter -- commented out for jshint*/) {
-            // TODO: implement view filter logic
+        function viewFilter(savedFilter) {
+            /* jshint camelcase: false */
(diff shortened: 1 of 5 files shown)


web/app/scripts/saved-filters/saved-filters-controller.js
```diff
@@ -27,8 +27,8 @@
         }
 
         // Activates the selected filter
-        function viewFilter(/*savedFilter -- commented out for jshint*/) {
-            // TODO: implement view filter logic
+        function viewFilter(savedFilter) {
+            $scope.$emit('driver.savedFilters:filterSelected', savedFilter.filter_json);
```
kshepard: I'm guessing jshint won't like that underscore.

kshepard:
This looks good. One more thing it needs to do though is when `view` is clicked from the dashboard widget, it should switch over to the map view.

kshepard:
+1

kshepard:
I just tried this out on staging, and there appears to be a problem with applying filters from the dashboard. If I clear fliters, go to the dashboard and apply a filter, it brings me to the map page, but with no filters applied.

sharph:
I'll have a look.

PR #330, opened 2015-12-02
Feature/admin only editor

flibbertigibbet:


kshepard:
+1. We should also restrict writing to the various schema endpoints (recordtype, recordschema, etc.) to admins, so non-admins can't modify schemas via the API. Perhaps this can be a card for next sprint.

flibbertigibbet:
Yes, I was thinking that restriction to the API endpoints might be a good task to go with the 'custom permissions' card we've got.

PR #332, opened 2015-12-02
Log out session also on frontend logout

flibbertigibbet:
Clear previous session as well as token when logging out through front end.


kshepard:
+1

PR #338, opened 2015-12-04
Add frontend support for text search capabilities

moradology:
A test was removed because of the difficulty of negotiating promises directly within the tests.

Other than that, the `getFilterables` call which returns a list of all filterable fields is used to determine which fields should be filtered using text search. Note that `contains` and `pattern` can't appear on the same `containment` or `containment_multiple` node - this is by design, as it would make combining these filtering options all but useless.

Ashlar and djsonb both need to be updated for this to work.


kshepard:
Didn't have a chance to test it out, but this looks good to me. The search text should make it into the saved filters list as well. Would you be able to verify that works and add it in if it doesn't? Probably just means editing saved-filter-as-html.js to make sure it renders correctly. +1 if that works.

moradology:
Superceded by #378

PR #339, opened 2015-12-04
Disable scroll zoom everywhere (except the big Map)

ddohler:
There's a comment in the source, but if we need to do more significant customization, we may want to come up with a way to pass in configuration options directly to the `leaflet-map` directive; currently we have to set them manually in one of the dependent directives' `bind` methods (or something that it calls). I talked to @CloudNiner about this and we didn't come up with any easy and elegant solutions, but we may want to revisit the issue later.


flibbertigibbet:
+1

PR #341, opened 2015-12-07
Use configured local time zone everywhere

kshepard:


web/app/scripts/views/record/add-edit-controller.js
```diff
@@ -97,10 +105,42 @@
             constantFieldsValidationErrors();
         }
 
+        /**
+         * Since the date and time pickers rely on the browser's local timezone with
+         * no way to override, we need to modify the occurred datetime before it gets
```
flibbertigibbet: Ugh. There are some pickers that use moment.js, such as the date/time picker and [this thing](https://www.npmjs.com/package/angular-bootstrap3-datepicker).

flibbertigibbet:
+1

The datepicker manipulation seems potentially fragile.

web/app/scripts/views/record/add-edit-controller.js
kshepard: Yeah, I hear you. But that library and the one we were using are both based on Eonasdan/bootstrap-datetimepicker, which has known major bugs with time zones (and is no longer being maintained). I tried getting it to work, and it would require even more involved workarounds. This is at least pretty self-contained and has a unit test. I'd definitely be up for exploring better date/time pickers at some point, or just writing our own.

flibbertigibbet: Yikes, ok. Suppose we may well end up making our own at some point then.

PR #342, opened 2015-12-07
Don't place marker on double-click

ddohler:
Delays placing a marker in order to wait for a double click. If a double
click happens, the marker is not placed.

### DIFF
--- a/web/app/scripts/views/record/embed-map-controller.js
+++ b/web/app/scripts/views/record/embed-map-controller.js
@@ -7,13 +7,13 @@
     var zoomInLevel = 17;
 
     /* ngInject */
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
+1

I suppose there wouldn't be any additional benefit to using [Rick's library](https://github.com/azavea/Leaflet.favorDoubleClick)?

ddohler:
I talked to Rick and he said that his library only works on limited versions of Leaflet and he hasn't had a chance to update it.

flibbertigibbet:
Ok, good to know.

ddohler:
Had to rework one of the tests after this change, and I added another one for good measure.

PR #345, opened 2015-12-15
Only return Accident Details to read-only users

ddohler:
Filters out fields besides Accident Details if a user is read-only. Relies on a forthcoming PR to Ashlar for the `MethodTransformJsonField`.
publicdetails
admindetails

### DIFF
--- a/app/data/tests/test_views.py
+++ b/app/data/tests/test_views.py
@@ -5,6 +5,7 @@ import uuid
 import mock
 import pytz
 
+from rest_framework.request import Request
 from rest_framework.test import APIClient, APITestCase, APIRequestFactory, force_authenticate
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
@@ -12,7 +13,8 @@ from django.conf import settings
 
 from ashlar.models import RecordSchema, RecordType, Record
 
-from data.views import DriverRecordViewSet
+from data.views import DriverRecordViewSet, DriverRecordSchemaViewSet
+from data.serializers import DetailsReadOnlyRecordSerializer, DetailsReadOnlyRecordSchemaSerializer
 
 
 class DriverRecordViewTestCase(APITestCase):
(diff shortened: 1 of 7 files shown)


app/data/tests/test_views.py
```diff
@@ -123,3 +125,31 @@ def test_tilekey_param(self):
...
+
+    def test_get_serializer_class(self):
+        """Test that get_serializer_class returns read-only serializer correctly"""
+        with mock.patch('data.views.is_admin_or_writer') as views_not_public:
+            views_not_public.return_value = False
+            view = DriverRecordViewSet()
+            mock_req = mock.Mock(spec=Request)
+            mock_req.user = self.admin
+            view.request = mock_req
+            # Even though the user is an admin, we've patched is_admin_or_writer to return False
```
flibbertigibbet: Why not just use a non-admin user? Also not clear why `is_admin_or_writer` is being patched in the test below when the user is public.

app/driver/settings.py
```diff
@@ -351,3 +351,6 @@
         'rest_framework.authentication.SessionAuthentication',
         'rest_framework.authentication.TokenAuthentication',
     )
+
+# These fields will be visible to read-only users
+READ_ONLY_FIELDS = ['Accident Details']
```
flibbertigibbet: Do we not want this filtering to apply to details fields generally? Seems like we should be able to match for top-level sections ending in ` Details` rather than have to configure this through the settings.

app/data/views.py
```diff
@@ -147,14 +157,15 @@ class DriverRecordTypeViewSet(RecordTypeViewSet):
     permission_classes = (IsAdminOrReadOnly,)
 
 
-class DriverSchemaViewSet(SchemaViewSet):
```
flibbertigibbet: Nice find

flibbertigibbet:
+1 when tests pass

app/data/tests/test_views.py
ddohler: Good call; I had thought I could get away with mocking just the one function, but that didn't work out so it ended up kind of a mishmash. I'll clean that up.

app/driver/settings.py
ddohler: Yeah, that was something I was thinking about -- it's possible that `Accident` might be renamed, but the client may also want to do something like making anonymized information on Passengers or Vehicles available to the public, in which case we'd need to find another approach. I'm not sure yet what is most likely to happen, so I went with this as something that would allow us to address a fairly wide variety of changes (relatively) quickly, and we can come up with tailor-made solutions once we know more about the likely use cases.

flibbertigibbet: Ok, sounds good

PR #346, opened 2015-12-16
Remove remaining edit links for public users

kshepard:
Also removes trailing pipe from saved filters display.


flibbertigibbet:
:+1:

PR #347, opened 2015-12-17
Feature/admin user mgmt

flibbertigibbet:
Interface for admins to add/edit/delete users from schema editor.


schema_editor/app/scripts/views/usermgmt/add-controller.js
```diff
@@ -0,0 +1,34 @@
...
+        ctl.userGroup = '';
+        ctl.groups = ASEConfig.api.groups;
+
+        ctl.submitForm = function() {
+
+            ctl.user.groups = [ctl.userGroup];
+
+            UserService.User.create(ctl.user, function() {
+                Notifications.show({text: 'User ' + ctl.user.username + ' created successfully',
+                                    displayClass: 'alert-info',
```
ddohler: I think the rest of the schema editor uses `alert-success` for create / save operations, so this should be switched to match.

ddohler:
I tried deleting a user and received 
```update or delete on table "auth_user" violates foreign key constraint "saved_filters_savedfi_owner_id_2563ae195232bf5c_fk_auth_user_id" on table "saved_filters_savedfilter"
DETAIL:  Key (id)=(2) is still referenced from table "saved_filters_savedfilter"``` . 

That looks like a pre-existing backend problem so it doesn't necessarily need to be solved as part of this, but if there's time left on the current card and it's easy, it might be good to address it now. I'll make an issue for it in either case.

+1 nice

PR #351, opened 2015-12-18
Fix filterbar

sharph:


web/app/scripts/views/dashboard/dashboard-controller.js
```diff
@@ -27,8 +28,10 @@
                 loadRecords();
             });
             $scope.$on('driver.savedFilters:filterSelected', function(event, selectedFilter) {
-                FilterState.restoreFilters(selectedFilter);
                 $state.go('map');
+                $timeout(function () {
+                    FilterState.restoreFilters(selectedFilter);
+                }, 2000);  // this needs to be quite long to avoid race conditions, unfortunately
```
kshepard: Wow, that's longer than I was expecting. What's the maximum duration you've tested that resulted in a problem? Just wondering if something like 1s is feasible or not.

kshepard:
+1

web/app/scripts/views/dashboard/dashboard-controller.js
sharph: 750ms had problems, and so I'm considering 2 * 1000 to be safe.

PR #358, opened 2016-01-08
Add models/views for black spots

kshepard:
Mock black spot loading will be in a coming PR.

### DIFF
--- a/app/black_spots/models.py
+++ b/app/black_spots/models.py
@@ -1,0 +1,38 @@
+import uuid as uuidlib
+
+from django.conf import settings
+from django.contrib.gis.db import models
+
+from ashlar.models import AshlarModel
+
+
+class BlackSpot(AshlarModel):
+    """A black spot -- an area where there is an historical/statistical concentration of records"""
+
+    #: Buffered road segment polygon where black spot analysis is performed
+    geom = models.PolygonField(srid=settings.ASHLAR['SRID'])
+
+    #: Number that determines the severity of this black spot. May be used for color-coding on map.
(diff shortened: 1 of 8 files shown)


app/black_spots/models.py
```diff
@@ -0,0 +1,38 @@
...
+    """A black spot -- an area where there is an historical/statistical concentration of records"""
+
+    #: Buffered road segment polygon where black spot analysis is performed
+    geom = models.PolygonField(srid=settings.ASHLAR['SRID'])
+
+    #: Number that determines the severity of this black spot. May be used for color-coding on map.
+    severity_score = models.FloatField()
+
+    #: Number of records accounted for in the polygon while analyzing
+    num_records = models.PositiveIntegerField()
```
ddohler: Is the reason you made these fields rather than `@property` because it would be easier to display black spots using windshaft? That makes sense to me but I initially thought it might be better to do them as `@property`, before I thought about the Windshaft case, so I'm curious what your reasoning was.

ddohler:
+1

app/black_spots/models.py
kshepard: I made them fields because they represent the state of the data at the point in time in which the future machine-learning algorithm calculates and inserts the black spots into the db. It'll have that information at the time, and that information may possibly change in the future if records are deleted or edited (i.e. Department of Health updates records to fatal). So I think it makes sense to store the data in the db to have a snapshot of what the polygon looked like at the time of processing.

ddohler: Ah, okay, that makes perfect sense. Thanks!

PR #359, opened 2016-01-11
Add black spot mock data loading script

kshepard:
Note: there is a black_spots.json file on fileshare for testing this script, but it doesn't have great mock data. One with more realistic mock data is in the works and will be uploaded shortly.

### DIFF
--- a/scripts/load_black_spots.py
+++ b/scripts/load_black_spots.py
@@ -1,0 +1,117 @@
+"""Loads mock black spot data"""
+import argparse
+from datetime import datetime
+import logging
+import json
+import requests
+import sys
+
+
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger()
+
+
+def get_record_type(api_url, headers, label):
+    """Gets the record type"""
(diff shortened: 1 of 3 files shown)


scripts/load_black_spots.py
```diff
@@ -0,0 +1,112 @@
...
+    results = response.json()['results']
+    rectype_id = [rt['uuid'] for rt in results if rt['label'] == label][0]
+    logger.info('record type id: {}'.format(rectype_id))
+    return rectype_id
+
+
+def read_black_spots(input_path):
+    """Reads the input json file of black spots"""
+    with open(input_path) as json_file:
+        black_spots_json = json.load(json_file)
```
ddohler: It'd be good to add exception handling here for cases where the JSON isn't valid; there was a minor syntax error in the `black_spots.json` file (I fixed it on the fileshare), and the default JSON exception could be friendlier. Something like `Error parsing JSON: <original exception message>` would make it clear what's happening.

ddohler:
+1, works great!

scripts/load_black_spots.py
kshepard: Ah, cool, I'll do that. Guess I should have tested that mock file after I made slight modifications to it! Thanks for fixing it.

PR #361, opened 2016-01-15
Persist selected layers through filter changes

Lknechtli:
Refactor layer controller setRecordLayers function

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -18,7 +18,7 @@
         ctl.map = null;
         ctl.overlays = null;
         // baseMaps was renamed to bMaps because the test framework does an
-        // unguarded string replace on the word "base" with with the base url
+        // unguarded string replace on the word "base" with the base url
         // in its error messages... 
         ctl.bMaps = null;
         ctl.editLayers = null;
@@ -270,18 +270,20 @@
             }
 
             if(ctl.layerSwitcher){
-                getUrls()
-                    .then(updateLayerGroups);
+                getUrls().then(updateLayerGroups);
             } else {
-                ctl.layerSwitcher = {init: true};
-                getUrls()
-                    .then(updateLayerGroups)
+                getUrls().then(updateLayerGroups)
                     .then(addBoundaryLayers)
                     .then(addLayerSwitcher);
             }
 
         };
 
+        /**
+         * Returns a promise which resolves to the URls for the record,
+         * utfgridtile, heatmap, and blackspot layers as an array of form:
+         * [baseRecordsUrl, baseUtfGridUrl, baseHeatmapUrl, blackspotsUrl]
+         */
         function getUrls() {
             return $q.all(
                 [TileUrlService.recTilesUrl(ctl.recordType),
@@ -297,9 +299,16 @@
                         }
                         return undefined;
                     })
-                ]);
+                ]
+            );
         }
 
+        /**
+         * Given an array of urls in form:
+         * [baseRecordsUrl, baseUtfGridUrl, baseHeatmapUrl, blackspotsUrl],
+         * updates the layer groups so that the layer switcher does not
+         * need to be re-initialized every time layer urls change
+         */
         function updateLayerGroups(urls) {
             var baseRecordsUrl = urls[0];
             var baseUtfGridUrl = urls[1];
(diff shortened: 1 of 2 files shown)


web/app/scripts/views/map/layers-controller.js
```diff
@@ -17,10 +17,17 @@
         ctl.drawControl = null;
         ctl.map = null;
         ctl.overlays = null;
-        ctl.baseMaps = null;
+        // baseMaps was renamed to bMaps because the test framework does an
+        // unguarded string replace on the word "base" with with the base url
```
ddohler: Extra 'with' in the comment.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -28,15 +35,20 @@
...
 
         // Ensure initial state is ready before initializing layers
         ctl.initLayers = function(map) {
-            // TODO (maybe): This approach to setting map options is only tenable so long as most
-            // of our maps use the same options, with very limited exceptions. If we ever
-            // need to have lots of customized options for each map, then we'll need to find a way to
-            // insert arbitrary config options into each map directive instance. One possibility
-            // would be to define them as attributes, similar to how angular-leaflet-directive does
-            // it, but there may be better approaches.
+            // TODO (maybe): This approach to setting map options is only tenable so
```
ddohler: Not a super big deal, but our preferred line length is 100 so these are a bit short.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -251,154 +262,207 @@
...
+                ctl.layerSwitcher = {init: true};
+                getUrls()
+                    .then(updateLayerGroups)
+                    .then(addBoundaryLayers)
+                    .then(addLayerSwitcher);
+            }
+
+        };
+
+        function getUrls() {
```
ddohler: It'd be good to add a comment explaining what `getUrls` and `updateLayerGroups` do and the order the layers are in.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -251,154 +262,207 @@
...
-                    detectRetina: true
-                };
 
-                // remove overlays if already added
-                if (ctl.overlays) {
-                    angular.forEach(ctl.overlays, function(overlay) {
-                        ctl.map.removeLayer(overlay);
+            if(ctl.layerSwitcher){
+                getUrls()
+                    .then(updateLayerGroups);
```
ddohler: Not sure there needs to be a linebreak here since there's only one `.then` call.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -251,154 +262,207 @@
...
+                        'effective_at': FilterState.getDateFilter().maxDate,
+                        'record_type': ctl.recordType
+                    }).$promise.then(function(blackspotSet) {
+                        var set = blackspotSet[blackspotSet.length - 1];
+                        if (set) {
+                            return TileUrlService.blackspotsUrl(set.uuid);
+                        }
+                        return undefined;
+                    })
+                ]);
```
ddohler: I'd drop the `)` onto a newline so it aligns with `return`.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -251,154 +262,207 @@
...
-                    heatmapOptions);
+        }
 
-                // interactivity for record layer
-                var utfGridRecordsLayer = new L.UtfGrid(ctl.getFilterQuery(baseUtfGridUrl, ctl.tilekey), {
-                    useJsonP: false,
-                    zIndex: 5
-                });
+        function addLayerSwitcher() {
+            if (!ctl.layerSwitcher.init) {
```
ddohler: Could you accomplish the same thing by moving the check for `if (ctl.layerSwitcher)` into this function and then returning if it's not null? That way the layerSwitcher won't get initialized twice but you shouldn't have to set it to this intermediate `{init: true}` state.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -251,154 +262,207 @@
...
+            var baseUtfGridUrl = urls[1];
+            var baseHeatmapUrl = urls[2];
+            var blackspotsUrl = urls[3];
+
+            updateEventLayer(baseRecordsUrl, baseUtfGridUrl);
+            updateHeatmapLayer(baseHeatmapUrl);
+            updateBlackspotLayer(blackspotsUrl);
+        }
+
+        function addBoundaryLayers() {
```
ddohler: This and `addLayerSwitcher` could also use some documentation.

ddohler:
+1, looks like this works. There are a few more of the refactored functions that could use documentation.

PR #367, opened 2016-01-25
Add audit logging for changes to records

ddohler:
Audit logs are generated on each create, update, or delete operation for a Record.
The audit-log endpoint is accessible only by admin users, and even for
them it is read-only. The endpoint is unpaginated by default, and
supports filtering by date ranges. It supports CSV output, since the
desired functionality is for users to be able to download monthly CSVs
of audit log activity.

### DIFF
--- a/app/data/filters.py
+++ b/app/data/filters.py
@@ -1,0 +1,14 @@
+import django_filters
+
+from models import RecordAuditLogEntry
+
+
+class RecordAuditLogFilter(django_filters.FilterSet):
+    """Allow filtering audit log entries by user, record, min_date, max_date"""
+    min_date = django_filters.IsoDateTimeFilter(name="date", lookup_type='gte')
+    max_date = django_filters.IsoDateTimeFilter(name="date", lookup_type='lte')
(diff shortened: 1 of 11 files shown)


flibbertigibbet:
If we do not require pagination on the audit log endpoint, maybe we should instead enforce a maximum number of log entries that may be returned at a time, as the logs will likely become quite large.

ddohler:
I think what I'll do is enforce a limit on the min_date and max_date so that it's not possible to search for more than a month's worth of log entries at once, since that's what the UI is going to provide. This may still be too much, but if that's the case then we'll have to generate the CSVs asynchronously.

flibbertigibbet:
That sounds good.

ddohler:
Okay, the audit-log endpoint has been updated to verify that `min_date` and `max_date` are specified and that they are no more than 31 days apart.

I also changed the default behavior of the endpoint to return JSON, which made it easier to debug because the CSV output swallows error messages. It still supports CSV output, but an `Accept text/csv` header needs to be provided.

flibbertigibbet:
:+1:

PR #368, opened 2016-01-28
Add task runner infrastructure

sharph:


app/driver/settings.py
```diff
@@ -239,7 +239,7 @@
 CACHES = {
     "default": {
         'BACKEND': 'django_redis.cache.RedisCache',
-        'LOCATION': 'redis://' + REDIS_HOST + ':' + REDIS_PORT + '/1',
+        'LOCATION': 'redis://' + REDIS_HOST + ':' + REDIS_PORT + '/2',
```
kshepard: If this redis database is changing to 2, it will also need to be reflected in windshaft. That'd be in windshaft/server.js. The current redis database is being selected with: redisClient.select(1).

kshepard:
+1 after windshaft modification. Nice.

PR #369, opened 2016-01-28
Add blackspot popups

Lknechtli:


web/app/scripts/state/mapstate-service.js
```diff
@@ -40,7 +40,7 @@
          * @param {number} mapZoom The zoom level of the leaflet map
          */
         function setZoom(mapZoom) {
-            $log.debug('Setting zoom to: ' + JSON.stringify(mapZoom));
+            //$log.debug('Setting zoom to: ' + JSON.stringify(mapZoom));
```
kshepard: Good call noticing that these were no longer needed. Might as well just go ahead and delete them.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -508,13 +513,59 @@
...
+                    .setLatLng(e.latlng)
+                    .setContent(ctl.buildBlackspotPopup(e.data))
+                    .openOn(ctl.map);
+
+                $compile($('#blackspot-popup'))($scope);
+            });
+        }
+
+        /**
+         *
```
kshepard: Looks like this is missing a comment.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -270,35 +270,39 @@
...
+            return BlackspotSets.query({
+                'effective_at': FilterState.getDateFilter().maxDate,
+                'record_type': ctl.recordType
+            }).$promise;
+        }
+
         /**
          * Returns a promise which resolves to the URls for the record,
          * utfgridtile, heatmap, and blackspot layers as an array of form:
          * [baseRecordsUrl, baseUtfGridUrl, baseHeatmapUrl, blackspotsUrl]
```
kshepard: This can be updated since it now has an additional item.

web/app/scripts/state/mapstate-service.js
Lknechtli: Forgot I did this lol.  Disabled them because it was getting in the way of my debug messages.  I'll get rid of the rest.

kshepard:
One minor thing: I just realized that the black spot layer is above the event layer. I think this layer order should be changed, otherwise you can't click on the events that exist inside a black spot.

kshepard:
+1

Lknechtli:
made issue #371 to address click events - it isn't as simple as fixing the layer z order

PR #370, opened 2016-01-28
Add group filter to user management view

KlaasH:
In the process:
* Changed the Gruntfile to stop ignoring bootstrap.js, since I wanted it
for the dropdown
* Added a .button-container class inside .form-area-heading to get the
floating to work with multiple buttons


kshepard:
+1, looks great!

PR #372, opened 2016-01-28
Add UI for downloading audit log CSVs

ddohler:
Adds a popup where users can download CSVs dumps of audit logs, by month. I ran out of time to add the link output that we discussed in RRP; I'll make an issue for that so it doesn't get forgotten.
prsauditlogs

--Edit-- The popup is accessible from the user dropdown.


flibbertigibbet:
+1

PR #374, opened 2016-01-29
Feature/add photos section

flibbertigibbet:
Add photos to uploader script schema. Also updates a role for a package version fix.


kshepard:
+1

Issue #376, opened 2016-01-29
Headers for collapsed edit sections do not use title/plural title


flibbertigibbet:
Collapsed sections with multiple items in `json-editor` form are shown with a label that is the field name, not its title.

PR #377, opened 2016-01-29
Feature/fix field names

### LINKED ISSUE #375: Encoded URI references do not match field names
URI references on the schema are now properly encoded by uncommenting the fix [here](#336), but the corresponding section names are not also encoded, and so do not match.

flibbertigibbet:
Fixes #375 for use of schemas with other libraries. Brought to light #376; collapsed sections, apart from details, show with a lower case label.


kshepard:
+1 on Travis success.

PR #378, opened 2016-02-01
Implement pattern-based search for records

moradology:
This PR implements pattern-based text search capabilities. When testing, *be sure to use a minimum date that includes 2013*.
Some of these fields have *no data* for *all* of the records after 2013. Assuming the independence of the values in the data can lead to some faulty assumptions/headaches/etc.


kshepard:
I have the same comment from the superceded PR about making sure saved filters render properly (i.e. updating saved-filter-as-html.js).

kshepard:
My computer's not in a good state to test it out, but this looks good to me. +1 with saved filter rendering and passing tests.

moradology:
Yeah, testing it out for yourself is rough - we can run through it on my machine if you like

PR #379, opened 2016-02-01
Add blackspot polygon filtering

Lknechtli:


app/black_spots/views.py
```diff
@@ -6,13 +9,47 @@
...
+            if sql:
+                tilekey_queryset = BlackSpot.objects.raw(sql)
+                tilekey_serializer = BlackSpotSerializer(tilekey_queryset, many=True)
+                tilekey_serializer.data.insert(0, {'count': len(tilekey_serializer.data)})
+                response = Response(tilekey_serializer.data)
+        else:
+            response = super(BlackSpotViewSet, self).list(self, request, *args, **kwargs)
+        return response
+
+    def _generate_query_sql(self, request):
```
ddohler: I wonder if we should break this and the redis connection code out into a separate function or base class so that we don't repeat this logic; we'll probably need similar functionality for interventions as well.

app/black_spots/models.py
```diff
@@ -31,7 +34,8 @@ class BlackSpotSet(AshlarModel):
     #: DateTime when the black spots become effective
     effective_start = models.DateTimeField()
 
-    #: DateTime when the black spotd are no longer effective. Should be null when first inserted.
+    #: DateTime when the black spots are no longer effective.
+    #  Should be null when first inserted.
```
kshepard: We've decided on max 100 columns per line, so these comments don't need to be broken up. May want to update your editor defaults.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -297,14 +306,29 @@
...
          */
-        function getUrls(blackspotSet) {
-            var set = blackspotSet[blackspotSet.length - 1];
+        function getUrls(response) {
+            var blackspotUrl = $q.when('');
+            var blackspotUtfUrl = $q.when('');
+            var blackspotTileKey = false;
+            if (response && response[0] && response[0].tilekey) {
+                var data = response[0];
+                if( data.tilekey ) {
```
kshepard: Spacing

web/app/scripts/views/map/layers-controller.js
```diff
@@ -499,25 +524,32 @@
          * blackspot layer and adds a new one to the layer group with
          * the updated URL
          */
-        function updateBlackspotLayer(blackspotsUrl, blackspotsUtfGridUrl) {
+        function updateBlackspotLayer(blackspotsUrl, blackspotsUtfGridUrl, tilekey) {
+            // TODO: update to pass in a tilekey which will filter on polygon
```
kshepard: Is this TODO still applicable? If there is some missing polygon filtering functionality, I think it should be cleared up before merging.

kshepard:
+1, possibly with some abstraction per @ddohler's good suggestion.

web/app/scripts/views/map/layers-controller.js
Lknechtli: nope, missed it in my cleanup sweep

kshepard:
+1

PR #380, opened 2016-02-01
Add celeryjob and duplicaterecord models

Lknechtli:


app/data/models.py
```diff
@@ -34,7 +35,45 @@ class ActionTypes(object):
...
         )
+
         @classmethod
         def as_list(cls):
             return [cls.CREATE, cls.UPDATE, cls.DELETE]
+
     action = models.CharField(max_length=6, choices=ActionTypes.choices)
+
+
+class CeleryJob(models.Model):
```
ddohler: I think we discussed this briefly in RRP, but I'm skeptical of the need for this model unless we want long-term persistence of activities since most of this functionality is already provided natively by the Celery API.

KlaasH: In conversation with Lucien and Kenny, we were talking about needing to track runs of this because it will take forever to look for duplicates in all records so after the first run we'll want to do smart things based on when it was run previously.  So yeah, in this case long-term persistence is useful.

What are the other celery tasks we're planning on?  For exports, I would agree that we don't need extra models on top of what celery provides.  For blackspots, I could see a case for persistence, but maybe there are ways to do it just as well without?

ddohler: Cool, that makes sense; I think there will also eventually be a report generation feature which would likely need to be asynchronous, but probably wouldn't need to be tracked over the long term. Not sure about blackspots; we might be able to get all the information we need from the blackspot and blackspotset models themselves.

kshepard: Yeah, I think this is the only case where we actually really care about when the last task was run since we'll need to make use of that value in the task itself for optimization.

app/data/models.py
```diff
@@ -34,7 +35,45 @@ class ActionTypes(object):
...
+            (STARTED, 'Started'),
+            (SUCCESS, 'Success'),
+            (ERROR, 'Error'),
+        )
+
+    class TaskType(object):
+        DEDUPE = 'DEDUPE'
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    date = models.DateTimeField(auto_now_add=True, db_index=True)
```
KlaasH: This is just like the one above, but my preference would be for a DateTime field to be called 'datetime' (or 'created') rather than just 'date'.

app/data/models.py
KlaasH: So yeah, I'd be on board with working under the assumption that this will be the only one we need and maybe calling it DuplicateFinderTask.  (and getting rid of `task_type`)

Lknechtli: agreed, makes sense to get rid of the task type if this is the only likely use case.  Worst case we expand it later

Lknechtli:
rebased on develop

KlaasH:
Finally got through provisioning and managed to try this out in shell_plus. Looks good.

PR #381, opened 2016-02-01
Fix details tabs and order them.

flibbertigibbet:
Record view tabs other than details section were broken by #377, causing them to show no data. (Sections were looked up by `title` instead of field name).

Also sort tab order by defined propertyOrder instead of sorting alpabetically.


kshepard:
+1

PR #383, opened 2016-02-01
Make minor fixes to Celery config

ddohler:
This fixes a few minor issues I ran into with Celery tasks. I'm making this as a separate PR since it sounds like some other folks may be working on Celery tasks soon too.

### DIFF
--- a/deployment/ansible/roles/driver.celery/templates/upstart-celery.conf.j2
+++ b/deployment/ansible/roles/driver.celery/templates/upstart-celery.conf.j2
@@ -21,7 +21,7 @@ pre-start script
 end script
 
 exec /usr/bin/docker run \
-  --name driver-app \
+  --name driver-celery \
   --publish 4000:4000 \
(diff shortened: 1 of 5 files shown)


kshepard:
+1, looks good to me. @sharph, want to take a quick look?

deployment/ansible/roles/driver.celery/templates/upstart-celery.conf.j2
```diff
@@ -21,7 +21,7 @@ pre-start script
 end script
 
 exec /usr/bin/docker run \
-  --name driver-app \
+  --name driver-celery \
```
sharph: The cronjob may need to be altered to match this.

sharph:
:+1:

deployment/ansible/roles/driver.celery/templates/upstart-celery.conf.j2
ddohler: Good call, fixed.

PR #384, opened 2016-02-02
Feature/interventions

fungjj92:
Sample intervention data available in the fileshare

### DIFF
--- a/scripts/interventions_schema.json
+++ b/scripts/interventions_schema.json
@@ -1,0 +1,106 @@
+{
+    "description": "",
+    "title": "",
+    "plural_title": "",
+    "definitions": {
+        "Intervention Details": {
+            "multiple": false,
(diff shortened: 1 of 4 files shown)


scripts/interventions_schema.json
```diff
@@ -0,0 +1,106 @@
+{
+    "description": "",
+    "title": "",
+    "plural_title": "",
+    "definitions": {
+        "Intervention Details": {
```
kshepard: There was a PR yesterday (https://github.com/WorldBank-Transport/DRIVER/pull/377) that changed the format of these ids. The gist is that this (and its references) should change to "interventionDetails". You'll probably want to rebase and/or make the change here manually.

kshepard:
+1 after updating the Intervention Details id.

PR #387, opened 2016-02-03
Modify RecordDuplicate model

KlaasH:
@Lknechtli: I cherry-picked your commits from feature/dedupe-celery-task and added one of my own--making it an AshlarModel to get 'created' and 'modified' attributes. Particularly I could see 'modified' being useful.


Lknechtli:
:+1: rebased my own branch onto it and it works fine

PR #388, opened 2016-02-03
Feature/schema loader cleanup

flibbertigibbet:
Correct section name references in loader script, and make text fields searchable


kshepard:
+1

PR #389, opened 2016-02-03
Duplicates list view

KlaasH:
Adds an API endpoint and a front-end view for record duplicates.

To test, use shell_plus to make a DedupeJob, find some records that match your current recordType, and create some RecordDuplicates like `RecordDuplicate.objects.create(job=DedupeJob.objects.first(), record=r[0], duplicate_record=r[1])`

The 'Resolve' button doesn't work.  That's the next card.  I'm also going to make tests for all of this as part of that card.

### DIFF
--- a/web/app/scripts/views/duplicates/duplicates-list-controller.js
+++ b/web/app/scripts/views/duplicates/duplicates-list-controller.js
@@ -1,0 +1,101 @@
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function DuplicatesListController($scope, $rootScope, $log, $modal, $state, AuthService,
+                                      InitialState, WebConfig, Duplicates, RecordState,
+                                      RecordSchemaState ) {
+        var ctl = this;
+        ctl.currentOffset = 0;
+        ctl.numDuplicatesPerPage = WebConfig.record.limit; //Just use the records limit
+        ctl.getPreviousDuplicates = getPreviousDuplicates;
+        ctl.getNextDuplicates = getNextDuplicates;
+        ctl.showResolveModal = showResolveModal;
+        ctl.userCanWrite = false;
+
+        InitialState.ready().then(init);
+
+        function init() {
+            ctl.isInitialized = false;
+            ctl.userCanWrite = AuthService.hasWriteAccess();
+
+            RecordState.getSelected().then(function(selected) { ctl.recordType = selected; })
+                .then(loadRecordSchema)
+                .then(loadDuplicates);
+        }
+
+        function loadRecordSchema() {
+            /* jshint camelcase: false */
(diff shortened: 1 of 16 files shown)


web/app/scripts/views/duplicates/duplicates-list-controller.js
```diff
@@ -0,0 +1,101 @@
...
+                if (ctl.currentOffset) {
+                    paramsOffset = ctl.currentOffset;
+                }
+            } else {
+                ctl.currentOffset = 0;
+                paramsOffset = 0;
+            }
+
+            /* jshint camelcase: false */
+            return Duplicates.query({record_type: ctl.recordType.uuid,
```
KlaasH: Is this incorrect to be filtering on record_type? It looked like that was what the record list was doing so I figured it was the thing to do. I'm not totally clear yet on the whole schema and record type setup.

Lknechtli:
This comment is something you should keep in mind for duplicate record resolution: 
https://github.com/WorldBank-Transport/DRIVER/pull/391#issuecomment-179874866

web/app/scripts/views/duplicates/duplicates-list-controller.js
Lknechtli: Huh. using record_type never occurred to me.  For large datasets with many record types, that could be a significant optimization for the dedupe task

Lknechtli: To answer your question though - I can see a pretty good usecase for allowing filtering duplicates by record_type.  We might actually want to add it as a RecordDuplicate field to make things easier if that's a usecase we want to support.

ddohler: Yeah, I think filtering on `record_type` makes sense here; the record type represents the system's understanding of the general class of spatiotemporal occurrences that are being stored, be it Accidents, Bird Sightings, Interventions, etc. The Schema meanwhile represents the information that's being stored for each `Record`, e.g. vehicles involved in an accident. `Schema`s are versioned and can change over time, and `Record`s with different `Schema`s can all be associated with the same `RecordType`. In this case, most of the controllers in the front-end app are going to need to filter down to the `RecordType` representing Accidents, so that we don't mix Accidents together with Interventions, which are also stored as `Record` objects but of a different `RecordType` (controllers dealing with Interventions will want to limit their queries to the `RecordType` for Interventions, of course).

kshepard:
+1 after adding tests.

KlaasH:
Upon reflection, rather than add and fix stuff on this branch and have to merge that into the branch I'm already working on for the Resolve Duplicate modal, I'm just going to scrap this PR and lump it all together in the other one.

PR #390, opened 2016-02-03
Create task to generate CSVs of Records

ddohler:
When `export_csv.delay('<tile-key>')` is called, this will kick off a Celery job to export Records as CSV files. Related info types are handled by creating multiple CSV files linked using the `_localId` keys. It uses the existing tile-key / redis process to filter the records so that they match what the user sees in the UI. Since there could potentially be multiple schemas among the records, this will theoretically create individual folders containing records for each schema, although I haven't fully tested that. There are likely to be some rough edges with certain field types and values, especially free-form text fields like Description.

The final output is placed in `/var/www/media/<tile-key>.tar.gz` on the Celery worker.

The speed is decent so far; exporting about 33,000 records took 24 seconds although I expect it to take longer with records where the related info is more fleshed out than in our test data set.

### DIFF
--- a/app/data/tasks/export_csv.py
+++ b/app/data/tasks/export_csv.py
@@ -1,0 +1,276 @@
+import csv
+import os
+import tarfile
+import tempfile
+
+from celery import shared_task
+from celery.utils.log import get_task_logger
+
+from django_redis import get_redis_connection
+
+from ashlar.models import Record
+
+logger = get_task_logger(__name__)
+
+
+@shared_task
+def export_csv(query_key):
+    """Exports a set of records to a series of CSV files and places them in a compressed tarball
+    :param query_key: A UUID corresponding to a cached SQL query which will be used to filter
+                      which records are returned. This is the same key used to generate filtered
+                      Windshaft tiles so that the CSV will correspond to the filters applied in
+                      the UI.
+    """
+    # Get Records
+    records = get_queryset_by_key(query_key)
+    # Get the most recent Schema for the Records' RecordType
+    # This assumes that all of the Records have the same RecordType.
+    schema = records[0].schema.record_type.get_current_schema()
+    # Create files and CSV Writers from Schema
+    record_writer = AshlarRecordExporter(schema)
+
+    # Write records to files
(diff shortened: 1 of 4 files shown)


kshepard:
I'm a little worried about treating the schemas individually and having a folder for each. I imagine after some use of the system, there will be many schemas, since a new one is created anytime someone add/removes a field, changes the searchability of a field, etc. It would then be a headache for someone who exported the data to make use of it. I think ideally the most recent schema for the record type would be used for deciding the fields to output for all the records, regardless of the schema it was referencing on insertion. This is similar to how it's done in the rest of the app: e.g. on the details view, the latest schema is pulled in and the fields of the record are displayed according to it. I haven't had a chance to review the code yet, but are there any major technical barriers involved in implementing it that way?

ddohler:
That's a good point, it had kind of slipped my mind that we're doing best-effort display here. I think that should be a pretty trivial change to make -- the output code already assumes that fields can be missing since most of them are missing from the test data anyway, and it should just ignore unexpected fields. I'll go ahead and update with that change.

kshepard:
Awesome, thanks. Sounds good.

ddohler:
Okay, I've updated this to use just the most recent Schema for the Records' RecordType, so I think this should be ready for review.

app/data/tasks/export_csv.py
```diff
@@ -0,0 +1,276 @@
...
+
+class RecordModelExporter(BaseRecordExporter):
+    """Exports records' model fields to CSV"""
+    def __init__(self, csv_columns, source_fields=dict(), value_transforms=dict()):
+        """Creates a record exporter
+        :param csv_columns: List of columns names to write out to the CSV.
+                            E.g. ['latitude', 'longitude']
+        :param source_fields: Dictionary mapping column names to the name of the model field where
+                              the appropriate value can be found.
+                              E.g. {'latitude': 'geom', 'longitude': geom}
```
kshepard: There are some missing single-quotes around the second `geom`.

kshepard:
+1, this looks exceptionally nice.

PR #391, opened 2016-02-03
Add dedupe task

Lknechtli:
Add migration for model change


Lknechtli:
Todo:
- [ ] Move constants to django config
- [ ] handling logic in case of job fail (OOM etc)
- [ ] general cleanup
- [ ] rename some of the methods to make more sense
- [ ] verify/test logic for incrimental task runs (after the first)

Lknechtli:
One note to keep in mind about the way record duplicates are stored -
If the a record appears as the primary record in a RecordDuplicate, it cannot appear as the secondary record in another RecordDuplicate.  This is to avoid creating two RecordDuplicates for each duplicate pair.

For 3 identical records A, B, and C, three record duplicates will be created - AB, AC, and BC.
This is important to keep in mind when handling record resolution: If AB is resolved in favor of B, then all other RecordDuplicates with A as the primary record should be marked as resolved as well (AC in this case)

Lknechtli:
consider filtering on record_type before geo/time
https://github.com/WorldBank-Transport/DRIVER/pull/389/files#r51882094

Lknechtli:
total runtime: 37:46 with 126376 records.

ddohler:
Looks like a lot of these commits can probably be squashed down before merging.

app/data/models.py
```diff
@@ -2,7 +2,7 @@
 
 from django.db import models
 from django.contrib.auth.models import User
-from ashlar.models import AshlarModel, Record
+from ashlar.models import AshlarModel, Record, RecordType
```
ddohler: Unused import? This should be a lint warning.

app/data/tests/test_dedupe_task.py
```diff
@@ -0,0 +1,96 @@
...
+        self.assertEqual(DedupeJob.objects.count(), 1)
+        self.assertEqual(RecordDuplicate.objects.count(), 3)
+        self.assertEqual(
+            RecordDuplicate.objects.filter(
+                Q(record=self.record4) | Q(duplicate_record=self.record4)
+            ).count(),
+            0
+        )
+
+        # test incrimental dedupe task
```
ddohler: incrimental -> incremental

app/data/tests/test_dedupe_task.py
```diff
@@ -0,0 +1,96 @@
...
+        # and one that shouldn't be considered a duplicate
+        self.record4 = Record.objects.create(
+            occurred_from=self.now,
+            occurred_to=self.now,
+            geom='POINT (0 5)',
+            location_text='somewhere else1',
+            schema=self.schema
+        )
+
+    def test_find_duplicate_records(self):
```
ddohler: As a general note, I think it would be good to add some tests that directly target your helper functions rather than just the top-level celery job. This guarantees that they're exercised even if the synthetic situation in the test wouldn't trigger all the code paths. It's also a really good gut-check on modularity; if it's easy to call the helper functions from tests then it's a good indication that they're fairly composable.

Lknechtli: gotcha.  Shouldn't take me too long.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    and location
+
+    Heuristic:
+    Records that have times within 1 day of each other
+    Records that are located ~100m from each other
+
+    Multiple tasks running:
+    Each job will only consider records added after the last task started.
+    This should allow multiple tasks to run at the same time without identifying
+    duplicates more than once.
```
ddohler: I'm not sure about this; it seems likely to me that duplicates could be added quite some time apart. For example, if two officers respond to the same incident, and one uploaded the records on her phone an hour later, while the other uploaded the records on his phone the next day.

I guess one option for how to handle this would be to consider records within a certain time into the past, but only create duplicates between those records and records that have been added after the most recent job, that way you're still guaranteed not to duplicate duplicates.

Another possibility might be to make a portion of the job atomic so that it can check for existing duplicates and then only create another duplicate if no duplicate exists for the found pair. This could possibly be done with Django's `@transaction.atomic`, and Celery also seems to have some type of locking mechanism although I don't know how well it works.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Job failure:
+    On job failure, the task should be retried after deleting the duplicates
```
ddohler: If you need to do cleanup tasks on failure you could use a Celery `errback` to be reasonably sure that the cleanup will occur even if the Celery task fails in a manner that doesn't update the task object status: http://docs.celeryproject.org/en/latest/userguide/calling.html#linking-callbacks-errbacks

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+                    page_num, paginator.num_pages, duplicatecount,
+                    datetime.datetime.now() - starttime
+                )
+            )
+        logger.info("Finished searching for duplicate records.")
+        job.status = DedupeJob.Status.SUCCESS
+    except Exception as e:
+        # TODO: retry job next time it runs with same start time
+        job.status = DedupeJob.Status.ERROR
+        logger.error("Failed: exception=\n%s" % (e, ))
```
ddohler: I'd also add a line here to set the Celery status to failure as well so that it lines up with the status in `job`. http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.update_state

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+from data.models import (DedupeJob, RecordDuplicate)
+from ashlar.models import Record
+
+import datetime
+from pytz import timezone
+
+logger = get_task_logger(__name__)
+
+
+def get_time_allowance():
```
ddohler: Would it work to make these default parameters to the job itself; then they could be overridden if we wanted to call the task manually? Something like 
```python
def find_duplicate_records(time_allowance=None, distance_allowance=None):
    if distance_allowance is None:
        distance_allowance = settings.DEDUPE_DISTANCE_DEGREES
```

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    return datetime.timedelta(hours=settings.DEDUPE_TIME_RANGE_HOURS)
+
+
+# .001 degrees is ~110m give or take 10 depending on location, which should
+# be good enough
+def get_distance_allowance():
+    return settings.DEDUPE_DISTANCE_DEGREES
+
+
+def calculate_score(record, duplicate):
```
ddohler: I'd change this to `calculate_similarity_score` so it's clear what it does -- the `duplicate` parameter could perhaps also be renamed since at this point we're not sure it's a duplicate yet.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    """ Get the time range to dedupe
+    Calculate from X hr before the last dedupe run to the current task run's
+    start time, where X is the heuristic's time allowance
+
+    Returns:
+        dictionary containing the start and end times for the next dedupe task
+        {starttime: timestamp, endtime: timestamp}
+    """
+    # TODO: get config var determining matching time - use value of 1 hr for now
+    time_allowance = get_time_allowance()
```
ddohler: I might be misunderstanding, but I'm not sure we should use the same time allowance for both the time range to dedupe over and for determining the similarity between records; these two things may have different sweet spots in terms of performance vs. false positives so we may want to set them to different values.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    """ Get the set of records to dedupe
+    return a set of uuids to check for duplicates, and the queryset used to
+    generate it
+    """
+    queryset = Record.objects.filter(
+        occurred_from__range=(
+            extent['start_time'], extent['end_time']
+        )
+    ).exclude(uuid__in=get_dedupe_ids())
+    ids = list(queryset.values_list(queryset.model._meta.pk.name, flat=True))
```
ddohler: I don't think it's necessary to use `list()` here; `values_list` should return a simple list in this case.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    return (ids, queryset)
+
+
+def get_dedupe_ids(job=None):
+    """ Get the ids of records to dedupe
```
ddohler: This comment doesn't seem to match what the function does; it looks like it returns a list of ids of records that have ever been flagged as a potential duplicate, and this causes `get_dedupe_set` to ignore those records.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+def get_dedupe_set(extent):
+    """ Get the set of records to dedupe
+    return a set of uuids to check for duplicates, and the queryset used to
+    generate it
+    """
+    queryset = Record.objects.filter(
+        occurred_from__range=(
+            extent['start_time'], extent['end_time']
+        )
+    ).exclude(uuid__in=get_dedupe_ids())
```
ddohler: Could this prevent the system from detecting triplicates?

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    if the celery worker itself goes down.  Otherwise, the task status should be
+    set properly
+    """
+    time_extent = get_time_extent()
+    ids, queryset = get_dedupe_set(time_extent)
+    # this can be adjusted depending on the memory required
+    paginator = Paginator(ids, 1000)
+    job = DedupeJob(status=DedupeJob.Status.STARTED)
+    job.save()
+    starttime = datetime.datetime.now()
```
ddohler: `starttime` and `duplicatecount` would be a bit more idiomatic as `start_time` and `duplicate_count`.

app/data/tasks/deduplicate_records.py
ddohler: Also, I think `values_list('pk')` should work.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+        logger.info("Finding duplicates. 1000 records/batch, %d batches total" %
+                    (paginator.num_pages))
+        for page_num in paginator.page_range:
+            page = paginator.page(page_num)
+            records = queryset.in_bulk(page.object_list)
+            done = []
+            for record in records:
+                done.append(record)
+                duplicates = find_duplicates_for_record(record)
+                recordDuplicates = []
```
ddohler: Same here, this should be `record_duplicates`.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+                for duplicate in duplicates:
+                    if duplicate.uuid not in done:
+                        # handle this in the find_duplicates_for_record call
+                        # to reduce number of queries
+                        num_reversed = RecordDuplicate.objects.filter(
+                            record__uuid=duplicate.uuid,
+                            duplicate_record__uuid=record
+                        ).count()
+                        if num_reversed == 0:
+                            rec = Record.objects.get(uuid=record)
```
ddohler: `Queryset.in_bulk` returns a dict of id/Model pairs, so it should be possible to avoid this extra database hit by iterating over `for rec_id, record in records.iteritems():`

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+            page = paginator.page(page_num)
+            records = queryset.in_bulk(page.object_list)
+            done = []
+            for record in records:
+                done.append(record)
+                duplicates = find_duplicates_for_record(record)
+                recordDuplicates = []
+                for duplicate in duplicates:
+                    if duplicate.uuid not in done:
+                        # handle this in the find_duplicates_for_record call
```
ddohler: Is this comment a TODO?

ddohler:
Looking good so far, needs a bit of cleanup.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    on the criteria.  score is a float between 0 and 1
+
+    normalized dimensions are considered equal (1 day equally as important as 100m)
+    """
+    tdelta = abs(record.occurred_from - duplicate.occurred_from)
+    tscore = 0
+    if tdelta.total_seconds() == 0:
+        tscore = 1.0
+    else:
+        tscore = tdelta.total_seconds() / get_time_allowance().total_seconds()
```
kshepard: Is this logic backwards? It looks like the score increases the farther apart the times are, when it seems it should decrease, since 1.0 is a perfect match.

app/data/tasks/deduplicate_records.py
Lknechtli: values('uuid') does

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+    ddelta = record.geom.distance(duplicate.geom)
+    dscore = 0
+    if ddelta == 0.0:
+        dscore = 1.0
+    else:
+        dscore = ddelta / get_distance_allowance()
```
kshepard: Same comment regarding potentially inversed logic.

app/data/tasks/deduplicate_records.py
ddohler: Right, but so does 'pk' and then you don't have to go through `_meta` to get the field name; it's just a bit shorter:
```python
In [15]: Record.objects.all()[:2].values_list('pk', flat=True) 
Out[15]: [UUID('d4ee4d6b-0fd7-46dd-9624-89f9b0524708'), UUID('7ff5d4ac-ee9e-43d5-9c01-9f2460ae4562')]
```

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+                                    record=rec,
+                                    duplicate_record=duplicate,
+                                    job=job,
+                                    score=calculate_score(rec, duplicate)
+                                )
+                            )
+                            duplicatecount += 1
+                RecordDuplicate.objects.bulk_create(recordDuplicates)
+            logger.info(
+                "Batch %d of %d done. %d duplicates found, %s elapsed" % (
```
kshepard: Our current preferred way of doing string interpolation is to use `.format` rather than the `%` way.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,185 @@
...
+            done = []
+            for record in records:
+                done.append(record)
+                duplicates = find_duplicates_for_record(record)
+                recordDuplicates = []
+                for duplicate in duplicates:
+                    if duplicate.uuid not in done:
+                        # handle this in the find_duplicates_for_record call
+                        # to reduce number of queries
+                        num_reversed = RecordDuplicate.objects.filter(
```
kshepard: Seems like `reversed` isn't the right term for this, unless I'm misunderstanding.

app/data/tasks/deduplicate_records.py
Lknechtli: shouldn't - I tested it with 3 identical records, and if spit out 3 duplicate pairs (for A,B,C, you should get AB,AC,BC)

app/data/tasks/deduplicate_records.py
Lknechtli: deleted this and did what the comment said

app/data/tasks/deduplicate_records.py
Lknechtli: good catch - remember thinking about this but I forgot to go back and fix it :)

Lknechtli:
tests to add still:
- [ ] get_time_extent()
- [ ] find_duplicates_for_record()
- [ ] calculate_similarity_score()

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def on_success(self, retval, task_id, args, kwargs):
+        job = DedupeJob.objects.get(celery_task=task_id)
+        job.status = DedupeJob.Status.SUCCESS
+
+    def on_failure(self, exc, task_id, args, kwargs, einfo):
+        """ On failure, delete duplicates found in the unsuccessful run"""
+        job = DedupeJob.objects.get(celery_task=task_id)
+        job.status = DedupeJob.Status.ERROR
+        RecordDuplicate.objects.filter(job=job).delete()
```
ddohler: There should be a call to `save()` here and in `on_success`, I think.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,187 @@
...
+
+def calculate_similarity_score(record1, record2, time_allowance, distance_allowance):
+    """
+    Set a score representing the likelyhood the two records are duplicated, based
+    on the criteria.  score is a float between 0 and 1
+
+    normalized dimensions are considered equal (1 day equally as important as 100m)
+    """
+    tdelta = abs(record1.occurred_from - record2.occurred_from)
+    tscore = 0
```
ddohler: I think these double assignments for `tscore` and `dscore` can be removed.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,187 @@
...
+        """ On failure, delete duplicates found in the unsuccessful run"""
+        job = DedupeJob.objects.get(celery_task=task_id)
+        job.status = DedupeJob.Status.ERROR
+        RecordDuplicate.objects.filter(job=job).delete()
+
+
+def calculate_similarity_score(record1, record2, time_allowance, distance_allowance):
+    """
+    Set a score representing the likelyhood the two records are duplicated, based
+    on the criteria.  score is a float between 0 and 1
```
ddohler: It would be good to note that the records need to be separated by no more than `time_allowance` and `distance_allowance` -- if the separation is more than that, this may output values below 0.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,187 @@
...
+    """ Get the set of records to dedupe
+    return a set of uuids to check for duplicates, and the queryset used to
+    generate it
+    """
+    queryset = Record.objects.filter(
+        created__range=(
+            extent['start_time'], extent['end_time']
+        )
+    ).exclude(uuid__in=get_dedupe_ids())
+    ids = queryset.values_list('uuid', flat=True)
```
ddohler: I still have a slight preference for `'pk'` here since it's a stand-in for "whatever key is the primary key", so it'll work with any Django model, not just those that have a primary key called `uuid`. This works for this model though so it's not a major issue.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,187 @@
...
+
+    Returns:
+        dictionary containing the start and end times for the next dedupe task
+        {starttime: timestamp, endtime: timestamp}
+    """
+    # TODO: get config var determining matching time - use value of 1 hr for now
+    last_job = None
+    if DedupeJob.objects.all().count() > 0:
+        last_job = DedupeJob.objects.latest()
+    start_time = None
```
ddohler: You actually don't need this declaration; `start_time` will be in scope even though it's initialized inside the `if` statement.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,187 @@
...
+    if last_job:
+        start_time = last_job.datetime
+    else:
+        start_time = Record.objects.earliest('created').created
+    end_time = datetime.datetime.utcnow().replace(tzinfo=timezone('UTC'))
```
ddohler: I think we usually try to avoid `replace`; I would probably go with `datetime.datetime.now(pytz.utc)`

app/data/tasks/deduplicate_records.py
Lknechtli: done

app/data/tasks/deduplicate_records.py
Lknechtli: I'm avoiding this by using the create date on the job

app/data/tasks/deduplicate_records.py
Lknechtli: handling this using on_failure and on_succuss

app/data/tasks/deduplicate_records.py
Lknechtli: Fixed by using the created time to get uuids for dedupeable records, and occurred_from on the matching filter

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,192 @@
...
+
+    Returns:
+        dictionary containing the start and end times for the next dedupe task
+        {starttime: timestamp, endtime: timestamp}
+    """
+    last_job = None
+    if DedupeJob.objects.all().count() > 0:
+        try:
+            last_job = DedupeJob.objects.exclude(uuid=job.uuid).latest()
+        except:
```
ddohler: If there's an exception in particular that you have in mind here, it's better to catch that specifically; a blanket
```python
except:
    pass
```
can be really tough to debug.

Lknechtli: whups. I just need to ignore the case that the job isn't found.

app/data/tasks/deduplicate_records.py
```diff
@@ -0,0 +1,192 @@
...
+    return {'start_time': start_time, 'end_time': end_time}
+
+
+def get_dedupe_set(extent):
+    """ Get the set of records to dedupe
+    return a set of uuids to check for duplicates, and the queryset used to
+    generate it
+    """
+    queryset = Record.objects.filter(
+        created__range=(
```
ddohler: I'm still not sure this entirely fixes the problem; I don't think it will detect a duplicate if a job runs in between when the first record is added and when the second is added:
```python
recordA = Record.objects.create(geom=Point(0,0), occurred_from=2016-01-01T00:00:00.000Z)
find_duplicate_records()
sleep(30)
recordB = Record.objects.create(geom=recordA.geom, occurred_from=recordA.occurred_from)
find_duplicate_records()
```

because the most recent duplicate task was created after recordA, I think line 88 will cause recordA to be excluded from the queryset.

ddohler:
Really good changes; +1 after taking a look at the behavior when a duplicate-detection job runs in between two duplicates being added.

app/data/tasks/deduplicate_records.py
Lknechtli: Works as intended

PR #393, opened 2016-02-03
Add generate black spots input script

kshepard:
This script reads in a shapefile of roads and a csv of records,
separates out intersections and divides up segements, matches
up records to nearest segments, aggregates counts by year, and
writes the data to a shapefile and a csv file. The output is
intended to be used in a machine learning algorithm to generate
black spots.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ To load mock black spots, run `python scripts/load_black_spots.py --authz 'Token
 
 To load mock interventions, run `python scripts/load_interventions.py --authz 'Token YOUR_AUTH_TOKEN' /path/to/interventions_sample_pts.geojson`.
 
+To generate black spot inputs, run `python scripts/generate_black_spots_training_input.py /path/to/roads.shp /path/to/records.csv`.
+
 ## Production
 
 TODO: Notes on creating a production superuser and adding a production OAuth2 application
--- a/app/Dockerfile.base
+++ b/app/Dockerfile.base
@@ -10,6 +10,12 @@ RUN apt-get update && apt-get install -y \
     libffi-dev \
     python-dev
 
+# install dependencies for generate_black_spots_training_input script
+RUN apt-get update && apt-get install -y \
+    libgeos-dev \
+    libspatialindex-dev \
+    libgdal1-dev
+
 RUN pip install --no-cache-dir --process-dependency-links --allow-external djsonb -r requirements.txt
 
 EXPOSE 4000
--- a/app/requirements.txt
+++ b/app/requirements.txt
@@ -14,3 +14,9 @@ pytz==2015.7
 mock==1.3.0
 djangorestframework-csv==1.4.1
 git+https://github.com/azavea/ashlar.git@develop#egg=ashlar
+# Dependencies for generate_black_spots_training_input script
+argparse==1.2.1
+Fiona==1.6.3
+pyproj==1.9.5.1
+Rtree==0.8.2
+Shapely==1.5.13
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -9,3 +9,5 @@ accident_schema_v3.json requires the directory `data_for_v3`
 load_black_spots.py requires `black_spots.json`
 
 load_interventions.py requires `interventions_sample_pts.geojson`
+
+generate_black_spots_training_input.py requires `blackspot_training/roads_utm.*` and `blackspot_training/all_crashes_2008-2012.csv`
--- a/scripts/generate_black_spots_training_input.py
+++ b/scripts/generate_black_spots_training_input.py
@@ -1,0 +1,423 @@
+#!/usr/bin/env python
+
+"""Loads records and roads shapefile, outputs data needed for training black spots"""
+import argparse
+import csv
+from dateutil import parser
+from dateutil.relativedelta import relativedelta
+import fiona
+from functools import partial
+import itertools
+import logging
+import os
+import pyproj
+import pytz
+import rtree
+from shapely.geometry import mapping, shape, LineString, Point
+from shapely.ops import transform, unary_union
+
+
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger()
+
+
+def read_roads(roads_shp):
+    """Reads shapefile and extracts roads and projection
+    :param roads_shp: Path to the shapefile containing roads
+    """
+    shp_file = fiona.open(roads_shp)
+    roads = [
+        shape(road['geometry'])
+        for road in shp_file
+        # We're only interested in non-bridge, non-tunnel highways
+        if (road['properties']['class'] == 'highway'
+            and road['properties']['bridge'] == 0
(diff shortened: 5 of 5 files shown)


README.md
```diff
@@ -72,6 +72,8 @@ To load mock black spots, run `python scripts/load_black_spots.py --authz 'Token
 
 To load mock interventions, run `python scripts/load_interventions.py --authz 'Token YOUR_AUTH_TOKEN' /path/to/interventions_sample_pts.geojson`.
 
+To generate black spot inputs, run `python scripts/generate_black_spots_input.py /path/to/roads.shp /path/to/records.csv`.
```
ddohler: I remember with HunchLab having a really hard time keeping all the inputs and outputs straight at first because the output from one thing was the input to another stage, so it might make this friendlier to new users to include "training" in the file name so that it is clear it's making features for machine-learning

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+def read_roads(roads_shp):
+    """Reads shapefile and extracts roads and projection"""
+    shp_file = fiona.open(roads_shp)
+    return ([shape(road['geometry']) for road in shp_file], shp_file.crs)
+
+
+def get_intersections(roads):
+    """Calculates the intersection points of all roads"""
+    intersections = []
+    for road1, road2 in itertools.combinations(roads, 2):
```
ddohler: Didn't know about this, neat.

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def read_roads(roads_shp):
+    """Reads shapefile and extracts roads and projection"""
+    shp_file = fiona.open(roads_shp)
+    return ([shape(road['geometry']) for road in shp_file], shp_file.crs)
+
+
+def get_intersections(roads):
+    """Calculates the intersection points of all roads"""
```
ddohler: It'd be good to include a :param: description for `roads` so it's clear what type of objects they are.

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+    for road1, road2 in itertools.combinations(roads, 2):
+        if road1.intersects(road2):
+            intersection = road1.intersection(road2)
+            if 'Point' == intersection.type:
+                intersections.append(intersection)
+            elif 'MultiPoint' == intersection.type:
+                intersections.extend([pt for pt in intersection])
+            elif 'MultiLineString' == intersection.type:
+                multiLine = [line for line in intersection]
+                first_coords = multiLine[0].coords[0]
```
ddohler: I'm not sure this handles the case where the two roads overlap in several places but it may not be worth worrying about since roads shouldn't really do that.

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+        if proj_dist > half_length:
+            mid_point = line.interpolate(half_length)
+            lines = []
+            lines.extend(
+                split_line(LineString(coords[:idx] + [(mid_point.x, mid_point.y)]), max_line_meters)
+            )
+            lines.extend(
+                split_line(LineString([(mid_point.x, mid_point.y)] + coords[idx:]), max_line_meters)
+            )
+            return lines
```
ddohler: I think a slightly shorter alternative could be `return split_line(head_line) + split_line(tail_line)`

scripts/generate_black_spots_input.py
kshepard: I was thinking that too after I submitted the PR: I'll go through all the functions and document the params.

README.md
kshepard: Good idea.

scripts/generate_black_spots_input.py
kshepard: Nice.

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+            non_int_lines.append(road)
+
+    # Union all lines found within a buffer, treating them as a single unit
+    int_multilines = [unary_union(lines) for _, lines in segments_map.items()]
+
+    # Split any long non-intersecting segments. It's not important that they
+    # be equal lengths, just that none of them are exceptionally long.
+    split_non_int_lines = []
+    for line in non_int_lines:
+        splits = split_line(line, max_line_meters)
```
ddohler: Since `split_line` is guaranteed to return an array of lines, you should be able to just do `split_non_int_lines.extend(split_line(line, max_line_meters))` (and you can apparently avoid the for-loop entirely with a nested list comprehension but those are usually confusing so I'm not sure that'd actually be an improvement).

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+        if splits:
+            split_non_int_lines.extend(splits)
+
+    # Return a tuple of intersection multilines and non-intersecting segments
+    return (int_multilines, split_non_int_lines)
+
+
+def read_records(records_csv, road_projection, record_projection, tz,
+                 col_id, col_x, col_y, col_occurred, col_severe):
```
ddohler: Nice generalization.

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
...
+
+    # Optional arguments
+    parser.add_argument('--output-dir', help='Directory where files are output', default='.')
+    parser.add_argument('--combined-segments-shp-name', help='Combined segments output .shp name',
+                        default='combined_segments.shp')
+    parser.add_argument('--combined-segments-csv-name', help='Combined segments output .csv name',
+                        default='combined_segments.csv')
+    parser.add_argument('--intersection-buffer-meters', help='Meters to buffer each intersection',
+                        default=5)
+    parser.add_argument('--max_line_meters', help='Maximum meters allowed for line segment',
```
ddohler: I wonder if this parameter (and some of the others) should have a more general name; it looks like the geometries will use whatever units the roads Shapefile is in, which might not always be meters. Alternatively, it looks like it may be possible to extract the projection units from the Shapefile and raise an exception if it's not meters, but I haven't dug into that enough to know how feasible it is.

kshepard: That's true, there's no reason it needs to be in meters. I can switch it to `units` or something similar in case someone wants to use something like feet.

scripts/generate_black_spots_input.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python
+
+"""Loads records and roads shapefile, outputs data needed for generating black spots"""
+import argparse
+import csv
+from dateutil import parser
+from dateutil.relativedelta import relativedelta
+import fiona
```
ddohler: We haven't done this with past import scripts on this project but this one has enough dependencies that I wonder if it might be a good idea to install them into the `app` or `celery` VM.

ddohler:
+1

scripts/generate_black_spots_input.py
kshepard: It sure does. I'll do that.

PR #396, opened 2016-02-08
Add viewset for controlling CSV export jobs

ddohler:
This sets up a viewset for kicking off and querying the status of CSV export jobs. It also sets up a proxy endpoint for downloading the results from the Celery worker. I'm not sure whether this will work on staging; if we've got internal DNS set up within the VPC then it should, otherwise it likely won't.

Aside from that, the biggest outstanding task is to clean up the exported files periodically so that they don't accumulate on the celery worker.

An export job can be started by POSTing to /csv-export/ with the
following JSON
{ "tilekey": "<tile-key-here>" }

The tilekey is the same used to ensure that Windshaft tiles match the UI filter settings.

If the endpoint is able to successfully start a Celery job, then it will return a unique UUID for the export job. Job status will then be available at `/csv-export/<UUID>/`. Note that because of the way Celery treats unknown job ids, accessing `/csv-export/<invalid job id>/` will return a status of PENDING, rather than a 404.

At the moment, it's not possible to cancel export jobs. However, it looks like if we switch to a database-based results backend, then it should be possible: http://docs.celeryproject.org/en/latest/reference/celery.contrib.abortable.html

### DIFF
--- a/app/data/views.py
+++ b/app/data/views.py
@@ -3,7 +3,10 @@ import uuid
 from dateutil.parser import parse as parse_date
 from datetime import timedelta
 
-from django.db import connection, transaction
+from celery import states
+
+from django.conf import settings
+from django.db import transaction
 from django.db.models import (Case,
                               When,
                               IntegerField,
@@ -17,8 +20,8 @@ from rest_framework import viewsets
 from rest_framework.decorators import list_route, detail_route
 from rest_framework.exceptions import ParseError
 from rest_framework.response import Response
-from rest_framework import serializers
 from rest_framework.settings import api_settings
+from rest_framework import status
 
 from rest_framework_csv import renderers as csv_renderer
 
@@ -34,6 +37,7 @@ from driver_auth.permissions import (IsAdminOrReadOnly,
                                      ReadersReadWritersWrite,
                                      IsAdminAndReadOnly,
                                      is_admin_or_writer)
+from data.tasks import export_csv
 
 import filters
 from models import RecordAuditLogEntry, RecordDuplicate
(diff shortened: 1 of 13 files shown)


app/data/views.py
```diff
@@ -229,3 +235,48 @@ def get_serializer_class(self):
...
+    def create(self, request, *args, **kwargs):
+        """Create a new CSV export task, using the passed filterkey as a parameter
+
+        filterkey is the same as the "tilekey" that we pass to Windshaft; it must be requested
+        from the Records endpoint using tilekey=true
+        """
+        filter_key = request.data.get('tilekey', None)
+        if not filter_key:
+            return Response({'errors': {'tilekey': 'This parameter is required'}},
+                            status=status.HTTP_400_BAD_REQUEST)
```
KlaasH: What about grabbing the tilekey here and returning an error immediately if it doesn't exist?  E.g. expanding the error-checking to:
<pre>        if filter_key:
            redis_conn = get_redis_connection('default')
            if redis_conn.get(filter_key) is None:
                return Response({'errors': {'tilekey': 'Unknown tilekey'}},
                                status=status.HTTP_400_BAD_REQUEST)
        else:
            return Response({'errors': {'tilekey': 'This parameter is required'}},
                            status=status.HTTP_404_NOT_FOUND)</pre>

Should be plenty fast and would save having to poll the task to find out it died a quick and fruitless death.

KlaasH:
+1, with or without the tilekey check.

app/data/views.py
ddohler: I'm going to pass on this for now; it makes testing a bit trickier because the ViewSet tries to access Redis, which isn't available on Travis, and I don't have time to put in the necessary mocking to make it work correctly. Hopefully in practice it won't be much of an issue because the polling will pick up the error relatively quickly.

PR #397, opened 2016-02-09
Duplicate records front-end

KlaasH:
This includes the changes from PR #389 (duplicates API endpoints and list view) and adds tests for that stuff plus the modal to display duplicate records side-by-side and resolve the duplicate.

~~As noted in a comment on PR #391 (https://github.com/WorldBank-Transport/DRIVER/pull/391#issuecomment-179874866), when there are multiple duplicates of a record (AB, AC, BC) and the user chooses one of the duplicate records as better (e.g. resolves AB with B) the other duplicates for that record (i.e. AC) also also get resolved. But resolving a duplicate with the first record doesn't have that effect (choosing A for AB doesn't affect BC).~~

The query isn't filtering by resolved status.  Probably it should, but I wasn't sure what the ideal interaction would look like--I don't think we want resolved duplicates to disappear instantly, but I don't know if we want an explicit filter setting, a refresh button, or just to clear duplicates when they refresh the page or use the next and previous links.  So I thought maybe that would be a later task with design input.

This depends on the feature/add-record-archived-property branch of Ashlar. I changed dev-requirements.txt accordingly for testing, but will throw that change away before merging.

### DIFF
--- a/app/data/tests/test_views.py
+++ b/app/data/tests/test_views.py
@@ -9,21 +9,19 @@ from rest_framework.request import Request
 
 from rest_framework.test import APIClient, APITestCase, APIRequestFactory, force_authenticate
 from rest_framework import status
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, Group
 from django.conf import settings
 
 from ashlar.models import RecordSchema, RecordType, Record
 
 from data.filters import RecordAuditLogFilter
-from data.models import RecordAuditLogEntry
+from data.models import RecordAuditLogEntry, DedupeJob, RecordDuplicate
 from data.views import DriverRecordViewSet, DriverRecordSchemaViewSet, DriverRecordAuditLogViewSet
 from data.serializers import DetailsReadOnlyRecordSerializer, DetailsReadOnlyRecordSchemaSerializer
 
 
-class DriverRecordViewTestCase(APITestCase):
-    def setUp(self):
-        super(DriverRecordViewTestCase, self).setUp()
-
+class ViewTestSetUpMixin(object):
+    def set_up_admin_client(self):
         try:
             self.admin = User.objects.get(username=settings.DEFAULT_ADMIN_USERNAME)
         except User.DoesNotExist:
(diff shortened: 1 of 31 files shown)


app/data/tests/test_views.py
```diff
@@ -9,21 +9,19 @@
...
+from data.models import RecordAuditLogEntry, DedupeJob, RecordDuplicate
 from data.views import DriverRecordViewSet, DriverRecordSchemaViewSet, DriverRecordAuditLogViewSet
 from data.serializers import DetailsReadOnlyRecordSerializer, DetailsReadOnlyRecordSchemaSerializer
 
 
-class DriverRecordViewTestCase(APITestCase):
-    def setUp(self):
-        super(DriverRecordViewTestCase, self).setUp()
-
+class RecordSetUpMixin(object):
```
ddohler: Nice, I like this.

KlaasH:
Note: two ways to make duplicates, both in shell_plus.
Artificially:
<pre>r = Record.objects.filter(schema__record_type=RecordType.objects.order_by('-created'))
job = DedupeJob.objects.create()
for i in range(0, 10, 2):
    RecordDuplicate.objects.create(job=job, record=r[i], duplicate_record=r[i+1])
    RecordDuplicate.objects.create(job=job, record=r[i], duplicate_record=r[i+2])
    RecordDuplicate.objects.create(job=job, record=r[i+1], duplicate_record=r[i+2])</pre>

Or for real:
<pre>from data.tasks.deduplicate_records import find_duplicate_records
find_duplicate_records()</pre>

ddohler:
Still working through the code, but I wanted to mention that I'm a little unclear on how the asymmetry behind the duplicate resolution works -- I may not be understanding the significance of the ordering here. My starting assumption is that if we have three potential duplicates `AB`, `AC`, and `BC`, and we determine that `AB` is a true duplicate and decide to keep `B`, then `AC` is actually now `BC` because we've determined that `A == B`, so we can throw that one out. On the other hand, if we decide to keep `A`, then `BC` becomes `AC` and we can still throw one of the duplicates out. So I think I may be missing something in my understanding of what `AB` means as compared to `BA`.

app/data/views.py
```diff
@@ -229,3 +229,37 @@ def get_serializer_class(self):
...
+        resolved = []
+        recordUUID = request.data.get('recordUUID', None)
+        if recordUUID is None:
+            # No record id means they want to keep both, so just resolve the duplicate
+            pass
+        elif recordUUID == str(duplicate.duplicate_record.uuid):
+            # Picking B means all duplicates with primary record A are resolved, not just this one
+            duplicate.record.archived = True
+            duplicate.record.save()
+            for dup in self.queryset.exclude(pk=pk).filter(record=duplicate.record):
```
ddohler: Have you tried this for duplicates that result in a lot of cascading resolutions? I could see this potentially taking a long time in situations where a bunch of records are all duplicates of each other. It might be more efficient to grab the ids of the records with a `.values_list('pk', flat=True)` and then perform a `.update(resolved=True)`, that way it's limited to two queries no matter what. But it may not be a problem, not sure.

KlaasH:
Yup, you're right.  There shouldn't be any asymmetry.  Will fix.

ddohler:
+1 I'm not going to have a chance to try this out, unfortunately, but it looks good aside from the commutativity (? not sure if that's strictly the correct term here) issues with resolution.

PR #398, opened 2016-02-09
Feature/intervention state2

fungjj92:
Save intervention type to record state object and modifies mock data.

Doesn't pull in schema to the state object to be consistent with how accident data is handled (can be changed) and I think writing it goes hand in hand with the next step which is to put the data on the map. 
Intentionally titles interventions related variables as secondary records to maintain code reusability.

### DIFF
--- a/schema_editor/test/mock/resources/resources-mocks.js
+++ b/schema_editor/test/mock/resources/resources-mocks.js
@@ -107,7 +107,7 @@
         };
 
         var RecordTypeResponse = {
-            'count': 2,
+            'count': 3,
             'next': null,
             'previous': null,
             'results': [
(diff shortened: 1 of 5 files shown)


schema_editor/test/mock/resources/resources-mocks.js
```diff
@@ -130,11 +130,24 @@
...
+                    'plural_label': 'Interventions',
+                    'description': 'Actions to improve traffic safety',
+                    'active': true
                 }
             ]
         };
 
         var RecordType = RecordTypeResponse.results[0];
+
+        var SecondaryRecordType = RecordTypeResponse.results[2];
```
flibbertigibbet: Might make things a little more readable to move the result above bird sightings, so it will be the second thing.

flibbertigibbet:
+1

PR #399, opened 2016-02-10
Add conf modal and remove resolved dups from the list

KlaasH:
* Adds a confirmation when resolving a duplicate by picking a record
* Moves the post-resolution logic into the list controller
* Changes the post-resolution action from just showing 'Resolved' to
  reloading the list.
* Removes the 'Status' column from the list view, since it's not
  informative if we aren't showing resolved duplicates.


ddohler:
+1

PR #402, opened 2016-02-15
Add Export tool tab and CSV export interaction

KlaasH:
Adds the 'Export' tab at the bottom of the map and makes the 'Export CSV' button work.  Clicking it hits the API to create a celery export job, then it polls that job until it succeeds (displaying "Exporting CSV" with an animation) and turns into a link to download it.

There's no model backing these exports and they're tied to tilekey, so hitting the export button gets you an export for your current filters, and if you change your filters it resets the tool.

The answer to the $interval testing mystery was that I needed to call $scope.$apply() again after flushing $httpBackend to resolve the job creation call so that the interval, which is a promise called inside the initial request's callback, would get processed.

### DIFF
--- a/web/app/scripts/export/export-controller.js
+++ b/web/app/scripts/export/export-controller.js
@@ -1,0 +1,92 @@
+(function () {
+    'use strict';
+
+    /**
+     * Handles the interaction for CSV exports.
+     * Makes a request to start an export job then polls for the result and provides the link.
+     * Uses filter query parameters provided by MapController.
+     */
+
+    /* ngInject */
+    function ExportController($rootScope, $scope, $interval, Exports, InitialState, QueryBuilder) {
+        var ctl = this;
+        var pollingInterval;
+        var POLLING_INTERVAL_MS = 1500;
+        var MAX_POLLING_TIME_S = 100;
+
+        InitialState.ready().then(initialize);
+
+        function initialize() {
+            ctl.isOpen = false;
+            ctl.pending = false;
+            ctl.downloadURL = null;
+            ctl.error = null;
+
+            ctl.toggle = toggle;
+            ctl.exportCSV = exportCSV;
+        }
+
+        function toggle() {
+            ctl.isOpen = !ctl.isOpen;
+            if (ctl.isOpen) {
+                $rootScope.$broadcast('driver.export.open');
+            }
+        }
+
+        function exportCSV() {
+            cancelPolling();
+            ctl.error = null;
+            ctl.downloadURL = null;
+
+            var params = _.extend({ tilekey: true, limit: 0 }, ctl.recordQueryParams);
+            // Get a tilekey then trigger an export
+            QueryBuilder.djangoQuery(true, 0, params).then(function(records) {
+                Exports.create({ tilekey: records.tilekey },
+                    function (result) {pollForDownload(result.taskid); },
+                    function () { ctl.error = 'Error initializing export.'; }
+                );
+            });
+        }
+
+        function pollForDownload(taskID) {
+            ctl.pending = true;
+            pollingInterval = $interval(function () {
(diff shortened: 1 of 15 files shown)


web/app/scripts/export/export-partial.html
```diff
@@ -0,0 +1,34 @@
+<div class="tool export" ng-class="{ 'open': ctl.isOpen }">
+    <div class="tool-tab" ng-click="ctl.toggle()" >
+        <span class="glyphicon glyphicon-share"
+              ng-if="ctl.isOpen || (!ctl.pending &amp;&amp; !ctl.downloadURL)"></span>
```
ddohler: I think some linter may have run amok here, I would expect these to be raw `&` characters.

web/app/scripts/export/export-controller.js
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        // Clears any in-progress or completed download, so they can do a new one if they want.
+        function reset() {
+            cancelPolling();
+            ctl.error = null;
+            ctl.downloadURL = null;
+        }
+
+        $scope.$watch(function () { return ctl.tilekey; }, reset);
```
ddohler: I see where you're going with this but I'm wondering whether it might get confusing for users; I could see someone wanting to kick off a CSV export, click around to set up a new set of filters while the first one is exporting, and then start another export job once the first one finishes, and I think it might be surprising if that caused their pending CSV export to disappear.

web/app/scripts/views/map/map-controller.js
```diff
@@ -88,6 +90,11 @@
                 params.polygon = userDrawnPoly;
             }
 
+            // Get a tilekey for the CSV export
+            QueryBuilder.djangoQuery(true, 0, _.extend({ limit: 0 }, params)).then(function(records) {
```
ddohler: It looks like this is causing a duplicate request to `records/?tilekey=true` every time the map filters change. It might be better for this to live in the export modal controller so that it only asks for a tilekey when the export button is clicked. Alternatively (and probably better), we could break out the tilekey functionality into a service so that there is only ever one tilekey request for each filter change and then the rest of the app can just pull from the service. That'd probably be a substantial refactor, however.

ddohler:
+1 after a bit of cleanup.

web/app/scripts/export/export-partial.html
KlaasH: Yeah, my editor makes them white on glaring red.  It's quite a persuasive highlighting scheme.  I should figure out how to add an exception.

web/app/scripts/export/export-controller.js
KlaasH: Yeah, I see your point.  I can make the download link show up below the export link rather than replacing it.  Question: do you think it's worth watching the filter params and only showing the export link when the filters don't match the most recent download, or should we just let them export whenever there's not one already pending, even if the answer will be the same?

I guess technically in production the answer wouldn't necessarily be the same, since other people could be adding or modifying records.  Probably the latter, then.

ddohler: Yeah, I think that all sounds right: disable the "Export" link if there's a pending download, enable it otherwise.

web/app/scripts/views/map/map-controller.js
KlaasH: I haven't checked, but I'm assuming the back-end is able to optimize the limit=0 query to be pretty easy.  But yeah, both points make sense.  I moved the tilekey query into the export call.

ddohler: Yeah, the endpoint is super fast, so it's not really a load concern. There is a potential concern around filling up the redis cache; under high enough load it might be possible for a user's tilekey to get pushed out of the cache before they have a chance to use it. So it's probably good not to treat tilekey requests as completely free, but they're certainly cheap.

PR #403, opened 2016-02-17
CSV export cleanup task

KlaasH:
* Adds a management task and cron job to clean up old (older than 6 hours) CSV export files.
* Changes the name of the duplicate-finder task to better reflect what it does and to match the corresponding cron job.
* Adds a celery_host_address definition to group_vars/staging that I hope will make the export stuff work and stop crashing the server.

Question: is there a way to simulate the staging environment on my machine to see if that last change actually works?


app/data/management/commands/remove_old_exports.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument('--max-age', type=int, default=6,
+                            help="Oldest files to keep, in hours (default: 6).")
+
+    def handle(self, *args, **options):
+        path = settings.CELERY_EXPORTS_FILE_PATH
+        min_mtime = time.time() - (options['max_age'] * 3600)
+        for filename in os.listdir(path):
+            file = os.path.join(path, filename)
```
ddohler: `file` is a Python built-in so it might be better to rename this variable although I think this will still work.

deployment/ansible/roles/driver.celery/tasks/main.yml
```diff
@@ -34,7 +34,12 @@
...
-        job="docker exec $(docker ps -q -f name=driver-celery) manage.py remove_duplicates"
+- name: Add find duplicate records cronjob
+  cron: name="find duplicate records"
+        job="docker exec $(docker ps -q -f name=driver-celery) manage.py find_duplicate_records"
         hour=1 minute=10
+
+- name: Clean up CSV export files cronjob
+  cron: name="export file cleanup"
+        job="docker exec $(docker ps -q -f name=driver-celery) manage.py remove_old_exports"
+        hour=5 minute=10
```
ddohler: I think we could run this quite a bit more frequently since the command itself omits files that are less than 6 hours old. Perhaps remove the `hour` parameter so that it runs once an hour?

deployment/ansible/group_vars/staging
```diff
@@ -17,6 +17,9 @@ redis_bind_address: "{{ hostvars['database']['ansible_ec2_local_ipv4'] }}"
 csrf_session_key: "{{ lookup('env', 'DRIVER_CSRF_KEY') | default('seekrit', true) }}"
 cookie_secret_key: "{{ lookup('env', 'DRIVER_COOKIE_KEY') | default('seekrit', true) }}"
 
+## CELERY SETTINGS
+celery_host_address: "{{ hostvars['celery']['ansible_ec2_local_ipv4'] }}"
```
ddohler: I think this should work. You might be able to test locally by changing https://github.com/WorldBank-Transport/DRIVER/blob/develop/Vagrantfile#L23 to `staging:children` and reprovisioning, but I'm not sure if that'll really be representative of what the staging environment actually looks like.

ddohler:
Looks like the command rename broke a test; +1 after that's fixed.

PR #404, opened 2016-02-17
Add script to output blackspot geometries

ddohler:
This is in support of the blackspot modeling work; we needed a way to export the model output to Shapefiles and GeoJSON so that it can be debugged and imported into the web app eventually. This sets up a script to do that conversion; it takes the model output (which is a CSV with the same format as the model input but with additional columns added) and uses the `id` field to join the model output to geometries from a shapefile. There will likely be some changes needed to this as the modeling work progresses.

I'm assigning this to @kshepard since he knows the most about `fiona` and this doesn't block anything (as well as potentially being subject to revision).

### DIFF
--- a/scripts/output_black_spots_geoms.py
+++ b/scripts/output_black_spots_geoms.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+
+"""Combines a CSV of black spot predictions with road segments to output blackspot geometries
+
+The modeling outputs a CSV containing blackspots and a geometry ID corresponding to the 'id' column
+in the combined segments shapefile. To visualize this or load it into the DRIVER web app, this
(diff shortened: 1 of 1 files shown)


scripts/output_black_spots_geoms.py
```diff
@@ -0,0 +1,137 @@
...
+            output.write(f)
+
+
+def main():
+    """Read command-line arguments and orchestrate processing"""
+    parser = argparse.ArgumentParser(description='Generate black spots geometries')
+
+    # Required arguments
+    parser.add_argument('segments_shp', help='Path to shapefile of segments used for training')
+    parser.add_argument('black_spots_csv', help='Path to CSV containing record data')
```
kshepard: This help text is a little off.

kshepard:
+1, tested, works great.

PR #405, opened 2016-02-18
Use record type plural title

### LINKED ISSUE #376: Headers for collapsed edit sections do not use title/plural title

fungjj92:
Fixes #376 
Hand in hand with json editor PR 9, branch title4: https://github.com/azavea/json-editor/pull/9

Uses plural_title or title instead of object key

### DIFF
--- a/web/app/scripts/views/record/list-controller.js
+++ b/web/app/scripts/views/record/list-controller.js
@@ -81,7 +81,9 @@
                     if (key.indexOf('Details') > -1) {
                         // keep the actual field name
                         // for lookup on ctl.recordSchema.schema.definitions
-                        ctl.detailsPropertyName = key;
+                        /* jshint camelcase: false */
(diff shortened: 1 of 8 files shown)


web/app/scripts/views/record/list-controller.js
```diff
@@ -81,7 +81,9 @@
                     if (key.indexOf('Details') > -1) {
                         // keep the actual field name
                         // for lookup on ctl.recordSchema.schema.definitions
-                        ctl.detailsPropertyName = key;
+                        /* jshint camelcase: false */
+                        ctl.detailsPropertyName = val.plural_title || val.title;
```
flibbertigibbet: Also fall back on `key` here?

flibbertigibbet:
+1

PR #406, opened 2016-02-18
Hide Audit Logs dropdown if user not admin

ddohler:
- Adds an "isAdmin" cookie that specifies whether a user is an administrator or not.
- Hides the My Account dropdown for unauthenticated users
- Hides the Audit Log download option from the My Account dropdown for non-admin users.


flibbertigibbet:
What happens with username/password authentication? It looks like this only sets the new cookie if the user logs in via SSO.

ddohler:
Updated to set cookies on standard login as well.

flibbertigibbet:
+1 on tests pass

Issue #373, opened 2016-01-28
Audit log output should include Record URLs, not just UUIDs.


ddohler:

kshepard:
This has been implemented.

PR #407, opened 2016-02-19
Audit log fixes

KlaasH:
* Adds a column for record detail URL to the CSV (resolves issue #373).
* Uses a package, angular-file-saver, to do the download. This resolves the permission errors that result from linking directly to the API without going through the front-end.
* A bit of error handling and layout tweaks.
* An unrelated change to BlackSpotsController because it was putting errors in my console, which were getting in the way of seeing the errors I was causing...

Speaking of errors, the new package causes Angular Batarang 0.10.6 to throw tons of errors ("Cannot read property '$id' of undefined").  It's caused by a bug in angular-hint, apparently (see https://github.com/angular/batarang/issues/285). I dealt with it by pulling Batarang from GitHub and changing the angular-hint dependency to the commit that resolves the issue.  Disabling Batarang also works, of course.


app/data/views.py
```diff
@@ -207,10 +207,10 @@ def list(self, request, *args, **kwargs):
             raise ParseError('max_date and min_date must be less than one month apart')
         return super(DriverRecordAuditLogViewSet, self).list(request, *args, **kwargs)
 
-    # Override default CSV field ordering for ease of use
+    # Override default CSV field ordering and include URL
     def get_renderer_context(self):
         context = super(DriverRecordAuditLogViewSet, self).get_renderer_context()
-        context['header'] = ('date', 'username', 'record_uuid', 'action', 'uuid',)
+        context['header'] = self.serializer_class.Meta.fields
```
flibbertigibbet: nice

flibbertigibbet:
+1

PR #409, opened 2016-02-19
Allow members of admin group to edit users

ddohler:
Previously the `UserViewSet` relied on the built-in 'is_staff' flag, so users in
the 'admin' group could not see a list of all users. This switches it to use the `is_admin` function used elsewhere.


flibbertigibbet:
:+1:

ddohler:
Going to wait to merge this until the updated Cebu import is finished; the momentary downtime from the deploy process makes the import script choke.

PR #410, opened 2016-02-20
Feature/android model builder

flibbertigibbet:
Adds a docker container for model jar building that lives on celery's machine. It subscribes to a redis channel for schema updates; whenever a schema message comes through, it builds a signed jar file of models representing the schema for the Android app. The jar file contents are then stored in redis; they're <50k, and get cleared out of redis after 30 days of non-use.

Adds a new API endpoint for the model jars that will kick off a jar build task if the jar for the requested schema was not found in redis, or will return the jar file found as a download if `format=jar` is specified in the request. Also adds triggers to automatically build new jars after schema creation/modification.


flibbertigibbet:
Unrelated `toddow` test apparently fails on Saturdays.

README.md
```diff
@@ -13,6 +13,10 @@ Obtain a pickpoint api key from https://pickpoint.io.
 Copy `deployment/ansible/group_vars/all.example` to `deployment/ansible/group_vars/all`
 and add the API key for pickpoint under `web_js_nominatim_key`.
 
+To build schema model jar files for the Android app, copy the signing keystore to `gradle/data`
+and set the password for the keystore under `keystore_password` in `deployment/ansible/group_vars/all`.
```
ddohler: Where can we find the password for the keystore?

flibbertigibbet: I'll message you on slack.

app/data/views.py
```diff
@@ -315,3 +348,55 @@ def create(self, request, *args, **kwargs):
...
+    def retrieve(self, request, pk=None):
+        """Return the model jar if found, or kick of a task to create it if not found"""
+
+        uuid = str(pk)
+        if not uuid:
+            return Response({'errors': {'uuid': 'This parameter is required'}},
+                            status=status.HTTP_400_BAD_REQUEST)
+
+        redis_conn = get_redis_connection('jars')
+        found_jar = redis_conn.get(uuid)
```
ddohler: What does the response look like if the jar building task in is progress but not finished yet?

gradle/run.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/sh
+
+# Build a JAR of updated Android models for a schema.
+# Takes three command-line arguments:
```
ddohler: Looks like this should say `two`.

ddohler:
+1

app/data/views.py
flibbertigibbet: It will create another task. Building takes less than ten seconds once the gradle dependencies are cached, so this shouldn't happen much or have much effect.

gradle/run.sh
flibbertigibbet: Yup

gradle/run.sh
```diff
@@ -0,0 +1,63 @@
...
+# name of json file gets used by jsonschema2pojo for the base class
+DRIVER_SCHEMA_FILE="DriverSchema.json"
+
+# clean up temporary files on exit
+trap "echo 'Cleaning up...'; rm $MODEL_JAR; rm -rf $SCHEMA_SOURCE_DIR; echo 'All done!'" EXIT
+
+set -e
+
+# delete redis key for schema, if it exists
+redis-cli -h $DRIVER_REDIS_HOST -p $DRIVER_REDIS_PORT -n $DRIVER_REDIS_JAR_DB DEL $1
```
ddohler: Thinking about this more, I'm wondering if it's possible for certain access patterns on the endpoint to cause an endless delete-recompile loop here:

1. Client 1 requests Key A, not found -> Kicks off Job 1
2. Client 2 requests Key A, not found -> Kicks off Job 2
3. Job 1 Completes building Jar file, sets Redis key
4. Job 2 starts processing, deletes Redis key
5. Client 1 requests key A, not found -> kicks off Job 3
6. Job 2 finishes
7. Job 3 deletes Redis key
8. Goto 5

flibbertigibbet: If there were an automated-retry process somewhere, then yes. I don't see this occurring in practice though.

The message queue is processed serially, so the subscriber task could re-check for an extant key before calling the bash script.  I'll do that then.

Issue #394, opened 2016-02-04
Filters fail to appear on first load


kshepard:
When loading the dashboard or map page directly, there is a javascript error in the console:
`TypeError: Cannot read property 'uuid' of undefined`

This causes the filters to not be visible. It appears that loading the list page directly doesn't exhibit this error.

KlaasH:
f8d7ada fixed an error like that, but I'm still seeing the filters to fail to load initially.

kshepard:
This appears to be fixed.

PR #415, opened 2016-02-23
Interventions: map layer and detail view

KlaasH:
1. Adds an interventions layer to the map.
2. Changes the record detail view (modal and page) to load record type and schema based on the record rather than using the primary selected type/schema, so that they'll work for interventions (i.e. for records that are not of the primary type).  The edit page also works but needs some attention, which will come in a separate PR.
3. Refactors QueryBuilder.djangoQuery a bit to make it possible to filter on record attributes (date, location, etc) but not on schema attributes (this lets the date and boundary filters work for interventions and keeps the json filters, which would be false since they're for the wrong record type, from eliminating them all).

Also:
* Fixes a situation where RecordSchemaState.get() was being called with no schemaID, which was resulting in the cached schema being cleared until something else called it properly, which I believe was causing issue #394.
* Depends on https://github.com/azavea/ashlar/pull/72.

### DIFF
--- a/web/app/scripts/state/recordstate-service.js
+++ b/web/app/scripts/state/recordstate-service.js
@@ -22,6 +22,7 @@
         svc.getOptions = getOptions;
         svc.setSelected = setSelected;
         svc.getSelected = getSelected;
+        svc.getSecondary = getSecondary;
         init();
 
         /**
@@ -86,9 +87,6 @@
                     selection = _.find(options, function(d) {
                         return d.label === WebConfig.recordType.primaryLabel;
                     });
-                    secondaryType = _.find(options, function(d) {
-                      return d.label === WebConfig.recordType.secondaryLabel;
-                    });
                 } else {
                     var oldRecordType = localStorageService.get('recordtype.selected');
                     if (oldRecordType) {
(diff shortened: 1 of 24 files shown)


web/app/scripts/state/recordstate-service.js
```diff
@@ -22,6 +22,7 @@
         svc.getOptions = getOptions;
         svc.setSelected = setSelected;
         svc.getSelected = getSelected;
```
ddohler: I don't think you made this change but it looks like the previous line is repeated twice?

web/app/scripts/views/map/layers-controller.js
```diff
@@ -322,35 +320,38 @@
...
-            var baseHeatmapUrl = urls[2];
-            var blackspotsUrl = urls[3];
-            var blackspotsUtfGridUrl = urls[4];
-            var blackspotTileKey = urls[5];
-
-            updateEventLayer(baseRecordsUrl, baseUtfGridUrl);
-            updateHeatmapLayer(baseHeatmapUrl);
-            updateBlackspotLayer(blackspotsUrl, blackspotsUtfGridUrl, blackspotTileKey);
+            updateEventLayer(urls.baseRecordsUrl, urls.baseUtfGridUrl);
+            updateSecondaryLayer(urls.secondaryRecordsUrl, urls.secondaryUtfGridUrl);
```
ddohler: How are you feeling about the naming scheme here now that you've used it a bunch? There was a thought when Interventions were first set up that they could be kept general, so that they could represent any arbitrary data that the users wanted to track. But reading this is a bit confusing because it's not clear what the Primary is that would necessitate a Secondary.

ddohler:
+1

web/app/scripts/views/map/layers-controller.js
KlaasH: Yeah, it's a little muddy.  The accidents layer goes by many names--base, event, primary, and that's not to mention the places where it's implicit (e.g. recordType).  Actually, I'm going to do a pass of standardizing that, at least within LayersController.

I think it works, though, for what it is now.  If we wanted to make the primary and/or secondary record types switchable, we'd want to clean up the nomenclature in more places and I think we'd find some spots where hidden assumptions or hard-coding would need to be refactored out.  If we wanted to go further and support an arbitrary number of active record types, that would be another matter.

But yeah, my feeling is at this point the level of abstraction is not too cumbersome and is providing some advantages.

PR #418, opened 2016-02-23
Add support for Ansible 2.0

hectcastro:
Mostly removing a handful of deprecation warnings and upgrading to versions of Ansible roles that work for Ansible 2.0. The role specification itself has changed from a comma-delimited file to YAML.


ddohler:
+1

PR #419, opened 2016-02-23
Keep empty data directory

flibbertigibbet:


flibbertigibbet:
Trivial, merging for build

PR #420, opened 2016-02-23
Pin pyoidc to a commit

flibbertigibbet:


flibbertigibbet:
Merging to fix deploy

PR #422, opened 2016-02-25
Glob postgres package versions

flibbertigibbet:


KlaasH:
What about
<pre>postgresql_package_version: "9.4.*.pgdg14.04*"
postgresql_support_libpq_version: "9.5.*.pgdg14.04*"</pre>
to stop tripping over the +n at the end but still get mad if something goes weird with apt and it would be pulling from the wrong repository?
Or maybe that's not a reasonable worry.

ddohler:
@hectcastro any standard practices here?

hectcastro:
I'd lean toward something like this:

```yaml
postgresql_version: "9.4"
postgresql_package_version: "9.4.*.pgdg14.04+1"
postgresql_support_repository_channel: "main"
postgresql_support_libpq_version: "9.5.*.pgdg14.04+1"
postgresql_support_psycopg2_version: "2.6"
postgis_version: "2.1"
postgis_package_version: "2.1.*.pgdg14.04+1"
```

Unfortunately, the double-glob approach doesn't work. I eventually went with putting it in the patch version slot so that we get bug fixes without having to think about it. Keeping the `pgdg14.04` part is important because otherwise we open ourselves up to installing the Ubuntu maintained PostgreSQL packages vs. the PostgreSQL maintained ones.

flibbertigibbet:
It looks like the suggestion is what we currently have, but provisioning locally seems to be broken at the moment.

flibbertigibbet:
`dpkg` on database VM says `9.4.6-1.pgdg14.04+1` of `postgresql-9.4` is installed.

hectcastro:
I opened https://github.com/WorldBank-Transport/DRIVER/pull/426 with changes to `all.example` that I tested and appear to work.

PR #423, opened 2016-02-25
Correct for field-name drift in read-only fields

KlaasH:
This makes the read-only record details view work again.

It sticks with the approach in PR #345 but adjusts for the drift that has occurred since then ('Accident Details' -> 'accidentDetails') and makes it a tiny bit more robust by using a regex.

Seems like the more complete and flexible solution would be to add metadata to the schema itself, but that would be a battle for another day.


kshepard:
+1, nice.

PR #424, opened 2016-02-25
Feature/manage images

flibbertigibbet:
Only load full record when user with access to full record goes to detail/edit view for the record. Reuses response transformer for public user that omits all but the `Details` section from the record information returned. This prevents potentially large amounts of data being loaded when image data exists on records.

Also fixes some Javascript errors when `secondaryType` is undefined in the map view, and fixes fetching the `Details` section for public users.

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -418,17 +418,17 @@
          * when a new filter is applied, layer selection is preserved
          */
         function addLayerSwitcher() {
-            var overlays = angular.extend(
-                _.zipObject([
-                    /* jshint camelcase: false */
-                    [ctl.recordType.plural_label, ctl.primaryLayerGroup],
-                    [ctl.secondaryType.plural_label, ctl.secondaryLayerGroup],
-                    /* jshint camelcase: true */
-                    ['Heatmap', ctl.heatmapLayerGroup],
-                    ['Blackspots', ctl.blackspotLayerGroup],
-                ]),
-                ctl.boundariesLayerGroup
-            );
+            /* jshint camelcase: false */
+            var recordLayers = [[ctl.recordType.plural_label, ctl.primaryLayerGroup]];
+            if (ctl.secondaryType) {
+                recordLayers.push([ctl.secondaryType.plural_label, ctl.secondaryLayerGroup]);
+            }
+            /* jshint camelcase: true */
+
+            recordLayers.push(['Heatmap', ctl.heatmapLayerGroup]);
+            recordLayers.push(['Blackspots', ctl.blackspotLayerGroup]);
+            var overlays = angular.extend(_.zipObject(recordLayers), ctl.boundariesLayerGroup);
+
             ctl.bMaps.then(
                 function(baseMaps){
                     // only add the layer switcher once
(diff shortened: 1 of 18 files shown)


KlaasH:
Re the secondary layer issues: I had meant to make things check the truthiness of ctl.secondaryLayer before doing secondary-related things, but clearly I lost sight of that part in the course of changing things.  I prefer checking and just not doing the things rather than setting a blank label (esp. since the latter gives me a blank entry in my layer picker).
E.g. changing the beginning of addLayerSwitcher() to this:
<pre>/* jshint camelcase: false */
var recordLayers = [[ctl.recordType.plural_label, ctl.primaryLayerGroup]];
if (ctl.secondaryType) {
    recordLayers.push([ctl.secondaryType.plural_label, ctl.secondaryLayerGroup]);
}
/* jshint camelcase: true */
recordLayers.push(['Heatmap', ctl.heatmapLayerGroup]);
recordLayers.push(['Blackspots', ctl.blackspotLayerGroup]);
var overlays = angular.extend(_.zipObject(recordLayers), ctl.boundariesLayerGroup);
</pre>

And changing updateSecondaryLayer() to start with:
<pre>if (!ctl.secondaryType) {
    ctl.secondaryLayerGroup = null;
} else {</pre>
and have the rest inside that `else`.

KlaasH:
I think the idea of downloading less information for the list view is a good one, but I think non-read-only users viewing details is going to be a pretty common use case and this makes it slow for any records with pictures, and over a slower connection having to re-request even a small record before showing the details modal seems likely to make that feel slow as well.

It would mean more changes to more things, but I think a more focused approach--either adding a 'deferred' option to the schema (i.e. in properties.photo.options) and making it load when the tab is clicked or something involving thumbnails.

flibbertigibbet:
Ok, this should be good to go now.

KlaasH:
The modal is called from the map popup as well.  I think it just needs `details_only: 'True'` added to the params being passed to Records.get() in MapController (https://github.com/WorldBank-Transport/DRIVER/blob/develop/web/app/scripts/views/map/map-controller.js#L31).

flibbertigibbet:
The map actually already has the details-only record, but isn't passing it all in to the modal method. I'll just modify it to pass along the record it already has.

flibbertigibbet:
Ok

web/app/scripts/views/map/layers-controller.js
```diff
@@ -771,7 +781,7 @@
         }
 
         function getTilekeys() {
-            var primary = QueryBuilder.djangoQuery(0, getAdditionalParams()).then(
+            var primary = QueryBuilder.djangoQuery(0, getAdditionalParams(), true, true, false).then(
```
KlaasH: I think we would want a details-only query here, but actually I don't think it would make a difference.  The tilekey saves the SQL query, which windshaft uses to get the records directly, so the special serializer handling doesn't come into play.

This would also explain why I'm seeing one 10MB tile in my network log...

flibbertigibbet: Here it should be just getting the tile key and not the actual records. I set this one for all fields so that Django would have what it needs to apply the filters.

flibbertigibbet: Windshaft interactivity was pulling all the record data into the UTFGrid response, although the data JSON doesn't actually get used any more (just the date and UUID, which are separate fields), so that was an easy fix.

KlaasH: Good point, and hooray for an easy fix.

KlaasH:
+1

PR #426, opened 2016-02-26
Pin PostgreSQL versions to a known working state

hectcastro:
Pin PostgreSQL and its dependencies to specific package versions within the PostgreSQL APT repository.

### DIFF
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -12,13 +12,16 @@ docker_options: "--storage-driver=aufs"
 python_version: "2.7.*"
 pip_version: "1.*"
 
-postgis_package_version: "2.1.*"
-postgresql_support_libpq_version: "9.5.*"
-
-postgresql_package_version: "9.4.*.pgdg14.04+1"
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
+1

Is there a reason still for these package versions to be specified within the project instead of updating the respective ansible roles? My intention with adding package versions here was to use a more inclusive glob than what the roles use, so we might not have to make fixes for package updates so often. The versions with this change are specific, though.

hectcastro:
> Is there a reason still for these package versions to be specified within the project instead of updating the respective ansible roles?

I think the goal is still for the roles to stay in working order, at least in isolation. The main reason why updating them at the project level is useful is so that you can quickly determine the database specifics in one file vs. having to look through several (some projects are on 9.3, others on 9.4).

> My intention with adding package versions here was to use a more inclusive glob than what the roles use, so we might not have to make fixes for package updates so often. The versions with this change are specific, though.

Not sure how to loosen them any further without potentially mixing PostgreSQL specific packages (`pgdg`) with Ubuntu specific ones.

PR #427, opened 2016-03-01
Fix record property display in list view

KlaasH:
The fields in record list view had started breaking because PR #405 took
ctl.detailsPropertyName to be a title when in fact it's a key. This
changes it to ctl.detailsPropertyKey and changes ctl.detailsProperty,
which is the title, to get the plural title if defined.

Question: ctl.detailsProperty isn't actually used anywhere.  Delete it? Or keep it around on theory that if someone wants that title in the future it would be a bother for them to have to figure out how to get at it again?

### DIFF
--- a/web/app/scripts/details/details-tabs-controller.js
+++ b/web/app/scripts/details/details-tabs-controller.js
@@ -23,6 +23,7 @@
             var sorted = _.map(ordering, function(section) {
                 var definition = ctl.recordSchema.schema.definitions[section];
                 definition.propertyName = definition.title;
+                definition.propertyKey = section;
                 return definition;
             });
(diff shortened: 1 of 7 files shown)


flibbertigibbet:
If I understand correctly, `detailsProperty` should be the same as `detailsPropertyName` (the title, if available), so I suppose the less-descriptively named `detailsProperty` might as well go if unused.

@fungjj92 care to take a look? LGTM.

KlaasH:
I went ahead and removed detailsProperty.  And I realized detailsProperty was also being used erroneously in the duplicates list and being set but never used by the dashboard controller, so I cleared it out of those places as well.

PR #428, opened 2016-03-01
Make add/edit view work for interventions

KlaasH:
Makes the add/edit record view work for interventions (i.e. for secondary records).
Some things are generic (it loads the secondary record type from config) but some things are hard-coded: it shows the 'occurred to' date and hides the weather and light fields based on ctl.secondary (a boolean to indicate that we're looking at a secondary record) rather than through any sort of generic config, which would be more proper.

Some notes:
* Makes the driver-details-selectlist field type check for an array and join with comma if that's what's stored.
* Changes the RecordState service to return a promise for getSecondary.
* Changes the date picker(s) to use local controller properties that get initialized on load and copied back to record.occurred_from and _to on save, and changes the timezone correction method accordingly.

You can edit an intervention from the link in the map pop-up.  To add one, go to /addsecondary.  A link to that will be coming in the next feature.

### DIFF
--- a/web/app/scripts/details/details-tabs-partial.html
+++ b/web/app/scripts/details/details-tabs-partial.html
@@ -4,19 +4,19 @@
          ng-init="properties = ctl.sortedProperties(definition.properties)">
 
         <driver-details-single
+            ng-if="!definition.multiple"
             data="::ctl.record.data[definition.propertyKey]"
             properties="::properties"
             record="::ctl.record"
-            definition="::definition"
-            ng-if="!definition.multiple">
+            definition="::definition">
         </driver-details-single>
 
         <driver-details-multiple
+            ng-if="definition.multiple"
             data="::ctl.record.data[definition.propertyKey]"
             properties="::properties"
             record="::ctl.record"
-            definition="::definition"
-            ng-if="definition.multiple">
+            definition="::definition">
         </driver-details-multiple>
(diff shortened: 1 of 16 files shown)


web/app/scripts/details/details-tabs-partial.html
```diff
@@ -4,19 +4,19 @@
          ng-init="data = ctl.record.data[definition.propertyName]; properties = ctl.sortedProperties(definition.properties)">
 
         <driver-details-single
+            ng-if="!definition.multiple"
```
KlaasH: This was just for readability. I felt like putting the ng-if at the end was burying the lede.

ddohler:
Quick initial reaction without having read any code -- I think needing to toggle the weather / light fields is an indication that it might be better for them to live on the schema rather than be Ashlar model fields. I'm not sure that's a feasible refactor at this point (and it's almost certainly outside the scope of these cards) but I wanted to throw it out in case it's a possibility.

ddohler:
The code looks good; in testing locally I had trouble getting it to detect the difference between primary / secondary record types on the View and Edit pages; it would still display the weather and light fields and it wasn't displaying the Occurred To field. The Add page worked well though. +1 after double-checking those.

ddohler:
+1 for the detail view fixes.

PR #430, opened 2016-03-03
Add report aggregation endpoint

ddohler:
Adds a `/crosstabs` endpoint to the `DriverRecordViewSet` that provides
aggregate values over user-configurable grouping parameters, which
currently includes boundaries, time periods, and enumeration fields.

Still TODO: I'd like to add a few tests because this endpoint has multifarious configuration options and behaviors, but since there are still about 2 points left on this card I'm going to pause for the moment to keep those points available for addressing any issues that come up as the front-end is put into place. If there is anything left over I'll add tests.

### DIFF
--- a/app/data/views.py
+++ b/app/data/views.py
@@ -13,6 +13,8 @@ from django.db.models import (Case,
                               When,
                               IntegerField,
                               DateTimeField,
+                              CharField,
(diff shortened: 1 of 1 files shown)


kshepard:
:+1:

PR #431, opened 2016-03-04
Interventions tool tab

KlaasH:
Adds a tool tab on the map for adding and exporting interventions.  The files and controllers and such are called 'interventions' but the code loads the secondary record type and gets its label from that.  So maybe the files should be called 'secondary', but that felt like it would cost more in clarity than it would 

I also moved the two existing tool tabs into a tools/ directory and added this one there.  They're pretty similar and somewhat linked (each needs to close when another opens), so it seemed confusing to have them in three seemingly-unrelated directories.

Since the export for interventions is the same thing as the primary record export, just with a different record type and a slightly different tilekey query, I put the export initialization and polling into a service that's shared between this and the Export tab.


web/app/scripts/resources/recordexports-service.js
```diff
@@ -0,0 +1,78 @@
...
+            var deferred = $q.defer();
+
+            Exports.create({ tilekey: tilekey },
+                function (result) {
+                    var taskID = result.taskid;
+                    pollingInterval = $interval(
+                        function () {
+                            Exports.get({ id: taskID }).$promise.then(function (response) {
+                                switch (response.status) {
+                                    case 'PENDING':
```
flibbertigibbet: Maybe this is excessively pedantic, but an explicit `break` here might be good rather than letting it fall through, in case somebody comes back later and makes changes without recognizing the fall-through behavior.

flibbertigibbet:
Interface looks/works great.

A few of notes on the export:

I added an intervention, and in the exported CSV, the `Type` field in `interventionDetails.csv` appears to have come through as the string representation of a single-item string list (`[u'Bicycle Facilities']`), but the items added by the loader script look correct (just the string).

The file download is a gzipped tarball; I expect the users will have an easier time dealing with a zip file. It also might be nice to give the download file a descriptive name, instead of it being a UUID.

The `records.csv` file include columns for `weather` and `light`. Are those relevant for interventions?

KlaasH:
The Type field on intervention was set up as an array, but I'm not sure that makes sense and it definitely makes handling them more complicated (aside from the current issues, it would make searching or aggregating more difficult).  It was working OK in the front-end because I had already worked around it, but now I've just changed it to not allow multiple selections.

To see that difference, you'll need to re-import interventions.  I also changed the  scripts/load_interventions.py to re-use the RecordType if it exists, which makes reloading the schema slightly less messy.

Re gzipped tarball: I don't disagree, but I took a stab at switching from tarfile to zipfile in export_csv.py and it didn't go that well, so I want to call that a separate task.  Similarly, the presence of weather and light fields in the Interventions export is not great, but it's part of the broader issue of having fields in the "constants" category (i.e. attributes of the Record object rather than inside the schema) that don't really belong there.  And working around it would require hard-coding the divergent behavior into the exporter here: https://github.com/WorldBank-Transport/DRIVER/blob/develop/app/data/tasks/export_csv.py#L144, which seems too messy to be worth it, to me.  Having extra columns in the csv export seems like less of an issue than having extra fields showing up on the front-end.

I did make the filename slightly friendlier, though.

flibbertigibbet:
:+1: 

Maybe open an issue for the tarball-to-zip change.

PR #433, opened 2016-03-07
Feature/region saving2

fungjj92:
Persist region across app

We want to save boundary data between sessions. So I changed localStorageService to local instead of session storage. Previously sessionStorage was chosen -- was this for a particular reason that this change shouldn't quash?
PR #142 

Oddity/bug under investigation: Local storage only works if browser history is cleared.

### DIFF
--- a/app/requirements.txt
+++ b/app/requirements.txt
@@ -3,7 +3,7 @@ djangorestframework-gis==0.9.1
 django-filter==0.11.0
 django-extensions==1.6.1
 django-cors-headers==1.0.0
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
My understanding was that the issue was boundary selection used to persist, but has broken recently. Is this to address a different issue?

fungjj92:
Two things:
- The blackspot map didn't adhere to the set boundary when changed outside the dashboard page
- Boundary selection persists, but not forever. John logged out, waited ~1 hr, and boundaries were lost upon login.

flibbertigibbet:
Ah, ok. Thanks for clearing that up.

flibbertigibbet:
:+1:

PR #434, opened 2016-03-08
Add custom reports modal

kshepard:
custom-reports-1
custom-reports-2
custom-reports-3
custom-reports-4

### DIFF
--- a/web/app/scripts/custom-reports/custom-reports-modal-controller.js
+++ b/web/app/scripts/custom-reports/custom-reports-modal-controller.js
@@ -1,0 +1,169 @@
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function CustomReportsModalController($log, $modalInstance,
+                                          FilterState, GeographyState, QueryBuilder,
+                                          RecordState, RecordSchemaState) {
+        var ctl = this;
+        ctl.closeModal = closeModal;
+        ctl.createReport = createReport;
+
+        // Only the date part is needed when displaying these dates
+        ctl.dateFormat = 'MMM D, YYYY';
+
+        // These are set as the selections are made
+        ctl.colAggSelected = null;
+        ctl.rowAggSelected = null;
+        ctl.geoAggSelected = null;
+
+        // List of aggregations available for rows and columns.
+        // The time-related options defined here (type: `Time`) are constant.
+        // This list is later populated with all available filterable types (type: `Filter`),
+        // and all available geographies (type: `Geography`).
+        ctl.rowColAggs = [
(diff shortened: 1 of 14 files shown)


web/app/scripts/custom-reports/custom-reports-modal-controller.js
```diff
@@ -0,0 +1,169 @@
...
+            {
+                label: 'Month of Year',
+                value: 'month',
+                type: 'Time'
+            },
+            {
+                label: 'Week of Year',
+                value: 'week',
+                type: 'Time'
+            }
```
ddohler: The endpoint also supports years, not sure if we actually want to let users select that though.

web/app/scripts/custom-reports/custom-reports-modal-controller.js
```diff
@@ -0,0 +1,169 @@
...
+                    params = _.extend(params, crosstabsParams);
+
+                    // Ensure the page limit parameter is not set
+                    if (params.limit) {
+                        delete params.limit;
+                    }
+
+                    // The crosstabs endpoint uses `record_type_id` rather than `record_type`.
+                    // Renaming it here, though the name should probably be standardized.
+                    params.record_type_id = params.record_type;
```
ddohler: Yeah, standardizing on `record_type` is probably better; I was thinking in Django terms but from the point of view of the Javascript it should just be `record_type`, not `record_type_id`.

web/app/scripts/custom-reports/custom-reports-modal-controller.js
kshepard: Oh, thanks, I missed that one. I think it's worth adding. If we see some performance issues while testing, we can always strip it out again.

web/app/scripts/custom-reports/custom-reports-modal-controller.js
kshepard: Cool, I kind of figured, but wasn't sure if there was a reason I was missing. I'll remove this and update the backend code.

ddohler:
+1

web/app/scripts/custom-reports/custom-reports-modal-controller.js
kshepard: Oh, I think it's just the docstring that was saying `record_type_id` on the backend -- the param actually being used is `record_type`. Easy fix!

ddohler: Hah, cool, sorry for the confusion!

PR #436, opened 2016-03-08
Allow deleting records from edit page (+ various bugfixes)

kshepard:
Also fixes a problem where archived records were not being properly excluded from queries.

Screenshot:
prs-delete


fungjj92:
It's great!
Only suggestion might be changing the modal question from "Really delete?" to something maybe more commonly  said like "Are you sure?" "Do you really want to delete?" etc.
 +1

kshepard:
Cool, good suggestion, I'll do that!

PR #439, opened 2016-03-10
Various minor blackspot script changes

kshepard:


scripts/generate_black_spots_training_input.py
```diff
@@ -374,12 +394,13 @@ def main():
...
     parser.add_argument('--record-col-occurred', help='Record column: occurred', default='DATETIME')
-    parser.add_argument('--record-col-severe', help='Record column: severe', default='FATAL')
+    parser.add_argument('--record-col-severe', help='(Optional) Record column: severe',
+                        default='FATAL')
     args = parser.parse_args()
 
     logger.info('Reading shapefile from {}'.format(args.roads_shp))
     roads, road_projection = read_roads(args.roads_shp)
-    logger.info('Found {} roads in projection: {}'.format(len(roads), road_projection['init']))
+    logger.info('Found {:,} roads in projection: {}'.format(len(roads), road_projection['init']))
```
ddohler: I didn't know `format()` could do that -- nifty!

ddohler:
I'm not quite clear how the severe records are getting separated out from the non-severe records; does that happen prior to this script?

+1 aside from that.

kshepard:
Yes, that happens prior to this script. The Philly input data we were testing with had a 'FATAL' column, so that was being used as the 'severe' column here. The Philippines CSV I grabbed to test with doesn't have any such attribute, which is why I made it optional.

PR #440, opened 2016-03-11
Custom report view

KlaasH:
Creates a custom report view, which you get to via the custom report modal from the map's Export tab.

I modified the crosstabs view a bit in the course of making this, the main changes being:
* Some format changes for easier usability in controller
* Added row totals
* Added labels for months and days of the week
* Changed the year range to be based on the query rather than pulling the range of all records ever.

A further note on date ranges: the year range should actually be based on the range selected in the filters.  That should also be the case for day, month, and week aggregates, which are not currently supported and need to be added.

Building the tables using ng-repeat was unacceptably slow, so this uses ng-repeat for the headers but builds the bulk of the table HTML by hand in the controller.

### DIFF
--- a/app/data/views.py
+++ b/app/data/views.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import uuid
+import calendar
 
 from dateutil.parser import parse as parse_date
 from datetime import timedelta
@@ -224,6 +225,10 @@ class DriverRecordViewSet(RecordViewSet, mixins.GenerateViewsetQuery):
             records before any aggregation is applied. This may result in some rows / columns /
             tables having zero records.
 
+        Note that the tables are sparse: rows will only appear in 'data' and 'row_totals',
+        and columns will only appear in rows, if there are values returned by the query.
+        'row_labels' and 'col_labels', however, are complete and in order.
+
         Response format:
         {
             "row_labels": [
@@ -244,18 +249,21 @@ class DriverRecordViewSet(RecordViewSet, mixins.GenerateViewsetQuery):
                     "data": {
                         {
                             "row_key1": {
-                                "col_key1": X,
-                                "col_key3": Y,
+                                "col_key1": N,
+                                "col_key3": N,
                             },
                         },
                         ...
-                    ]
+                    },
+                    "row_totals": {
+                        {
+                            "row_key1": N,
+                            "row_key3": N,
+                    }
                 },
                 ...
             ]
         }
-        Note that all combinations of rows / columns are NOT guaranteed to exist within the
-        table `data` dictionaries; in other words, the tables are sparse.
         """
         valid_row_params = set(['row_period_type', 'row_boundary_id', 'row_choices_path'])
         valid_col_params = set(['col_period_type', 'col_boundary_id', 'col_choices_path'])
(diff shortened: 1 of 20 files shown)


web/app/scripts/custom-reports/aggregations-config.js
```diff
@@ -0,0 +1,108 @@
...
+         * enumerated property it finds.
+         *
+         * TODO:  add " || property.format === 'number'" to the condition to enable aggregating
+         * numerical properties if/when that's implemented on the back end.
+         */
+        function loadFilters(recordType) {
+            /* jshint camelcase: false */
+            return RecordSchemaState.get(recordType.current_schema).then(function(schema) {
+            /* jshint camelcase: true */
+                _.forEach(schema.schema.definitions, function(definition, defName) {
```
CloudNiner: Is it intentional that you're returning nothing from the then block in this function? The promise returned by loadFilters will not wait until this 'then' block resolves before running.

Same question for loadGeographies below

web/app/scripts/custom-reports/custom-report-controller.js
```diff
@@ -0,0 +1,88 @@
...
+            ctl.loading = true;
+            ctl.params = $stateParams;
+
+            getCategoryLabels().then(function () {
+                return Records.report(ctl.params).$promise;
+            }).then(function (report) {
+                console.log('report ready');
+                ctl.report = report;
+                composeTables();
+                ctl.loading = false;
```
CloudNiner: ctl.loading = false should go in a finally block, instead of in the then. Consider displaying a small error message if getCategoryLabels fails.

web/app/scripts/custom-reports/custom-reports-modal-controller.js
```diff
@@ -153,16 +90,23 @@
...
-                    $modalInstance.close();
+                    ctl.params = params;
                 }
             );
         }
 
+        function createReport() {
+            // If we were using $resource, Angular would automagically encode params that are
+            // objects. But since we're using $state.href, we have to do it ourselves.
+            _.forEach(ctl.params, function (value, key) {
```
CloudNiner: Glad that worked!

CloudNiner: Consider creating a new object to pass to `$state.href` so that you're not mutating the state of the existing ctl.params object. Certainly doesn't matter now but could be a nasty side effect in the future if more stuff is added to this controller.

app/data/views.py
```diff
@@ -286,64 +305,73 @@ def crosstabs(self, request):
...
-        valid_periods = easy_ranges.keys() + ['year']
-        if period_type not in valid_periods:
-            raise ParseError(detail=('row_/col_period_type must be one of {}; received {}'
-                                     .format(valid_periods, period_type)))
 
-        # Set the range min / maxes based on the period type
-        period_range = None
+        # TODO: This sets year range based on the oldest and youngest records in the filtered
+        # queryset, but it (and possibly other date ranges) should be defined by the request's
+        # occurred_max and occurred_min parameters if available.
```
ddohler: Shouldn't the filtered queryset already be limited by the `occurred_max` and `occurred_min` query params by this point? I agree that determining the other ranges from the query set / query params is probably a good way to boost user friendliness. That way the number of days will be correct if the user does a report for a specific month, for example.

web/app/scripts/custom-reports/custom-report-controller.js
```diff
@@ -0,0 +1,88 @@
...
+        init();
+
+        function init() {
+            ctl.loading = true;
+            ctl.params = $stateParams;
+
+            getCategoryLabels().then(function () {
+                return Records.report(ctl.params).$promise;
+            }).then(function (report) {
+                console.log('report ready');
```
ddohler: This can probably go away, the app is already fairly chatty.

flibbertigibbet: Or if kept, should be angular `$log` (so IE won't barf).

app/data/views.py
KlaasH: It will be, but I think we should use the filter params for all range aggregations so that if there are periods at the beginning or end with no records they'll show up at zeroes rather than not showing up in the report.

web/app/scripts/custom-reports/custom-report-controller.js
KlaasH: Yup, unintentional.

app/data/views.py
ddohler: Ah, gotcha. Yeah, that makes sense, hadn't thought of that case.

ddohler:
+1

PR #442, opened 2016-03-11
Fix provisioning error with Windshaft

ddohler:
Fixes an error that @sharph and I encountered while setting up the Saudi / Training instances.


KlaasH:
+1

PR #443, opened 2016-03-11
Remove stepwise bars for 0 count

fungjj92:
Before:
screen shot 2016-03-11 at 2 47 39 pm

After:
screen shot 2016-03-11 at 2 46 39 pm

Hurrah


KlaasH:
:+1:

PR #444, opened 2016-03-11
Don't revert to default date when loading filters

sharph:


sharph:
@ddohler and I worked on this together. Will merge as soon as it passes tests.

PR #446, opened 2016-03-11
Fix javascript errors from negative heights

KlaasH:
I was hasty and failed to notice that PR #443 introduced javascript errors for negative heights.  This achieves the goal of that PR (making the zero-value bars stop showing up) without making D3 angry.


flibbertigibbet:
Thinking there must be a cleaner way to deal with that (hide the bar altogether or something).

+1 on this fix

PR #447, opened 2016-03-14
Fix cron jobs for './' on manage.py, timezone

KlaasH:
The cron tasks weren't running because they were trying manage.py rather than ./manage.py.  And I adjusted the schedule since they were written as though the server would be in local time but in fact it's in UTC so they'd be running mid-morning.

### DIFF
--- a/deployment/ansible/roles/driver.celery/tasks/main.yml
+++ b/deployment/ansible/roles/driver.celery/tasks/main.yml
@@ -29,10 +29,10 @@
 - name: Ensure Driver celery service is running
   service: name=driver-celery state=started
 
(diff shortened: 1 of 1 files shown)


KlaasH:
Actually, is the calculate_black_spots command good to go, or do we possibly want to keep that from running automatically just yet?

ddohler:
Looks like that command is a placeholder at the moment; there probably wouldn't be any harm in letting it run automatically, but I'd probably lean toward disabling it until it actually does something.

ddohler:
+1 sorry I missed the update

sharph:
:+1:

PR #448, opened 2016-03-14
Pass query parameters to graph functions properly

ddohler:
The `stepwise` and `toddow` functions expect `extraParams` to be an Object, but the order of the parameters was incorrect and it was getting passed as a boolean instead.

### DIFF
--- a/web/app/scripts/views/dashboard/dashboard-controller.js
+++ b/web/app/scripts/views/dashboard/dashboard-controller.js
@@ -65,7 +65,7 @@
             });
             /* jshint camelcase: true */
 
-            RecordAggregates.toddow(false, params).then(function(toddowData) {
+            RecordAggregates.toddow(params, false, false).then(function(toddowData) {
                 ctl.toddow = toddowData;
             });
         }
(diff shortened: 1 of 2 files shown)


KlaasH:
Arg, I was afraid that bug would be fallout from that little QueryBuilder refactor (PR #415).  To make the ToDDoW on the dashboard work we should also change `RecordAggregates.toddow(false, params)` to `RecordAggregates.toddow(params, false, false)` at https://github.com/WorldBank-Transport/DRIVER/blob/develop/web/app/scripts/views/dashboard/dashboard-controller.js#L68.

Kenny mentioned the count widget on the dashboard as well.  In that case it look like the polygon is being passed to the API but isn't affecting the answer.

ddohler:
I've updated the parameters for the ToDDoW on the dashboard as well.

@kshepard : After some discussion with @KlaasH , it looks like the recent event counts on the dashboard have never filtered by the selected boundary correctly. Fixing them would require relatively larger surgery, unfortunately, because we'll need to refactor how some of the viewsets are structured, so we're going to pass on that for now.

KlaasH:
+1

PR #449, opened 2016-03-14
Add sequential options to custom report view

KlaasH:
Renames some of the existing periodic aggregation options ('day' ->
'day_of_month') and adds sequential options ('month', 'week', and 'day').

Still to do: add tests.

### DIFF
--- a/app/data/views.py
+++ b/app/data/views.py
@@ -2,9 +2,10 @@ import json
 import logging
 import uuid
 import calendar
+import datetime
 
 from dateutil.parser import parse as parse_date
-from datetime import timedelta
+from django.template.defaultfilters import date as template_date
 
 from celery import states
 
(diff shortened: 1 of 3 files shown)


app/data/views.py
```diff
@@ -331,57 +332,152 @@ def _query_param_to_case_stmnt(self, param, request, queryset):
...
+                'lookup': lambda (yr, wk, day): {'occurred_from__week': wk,
+                                                 'occurred_from__year': yr,
+                                                 'occurred_from__day': day},
+                'label': lambda (yr, wk, day): template_date(datetime.date(yr, wk, day))
+            },
+        }
+
+        if period_type in easy_ranges.keys():
+            period = easy_ranges[period_type]
+        elif period_type in sequential_ranges.keys():
```
ddohler: Hmm, I feel like this could be a potential booby trap if someone modifies `easy_ranges` without realizing that it has to be disjoint from `sequential_ranges`. Not exactly sure what to do about it though or if it's even worth addressing since there aren't many periodic aggregations outside of what's already here.

app/data/views.py
```diff
@@ -331,57 +332,152 @@ def _query_param_to_case_stmnt(self, param, request, queryset):
...
+            if request.query_params.get('occurred_max') is not None:
+                max_date = parse_date(request.query_params['occurred_max']).date()
+            else:
+                max_date = queryset.order_by('-occurred_from').first().occurred_from.date()
+
+            # Build the relevant range of aggregation periods. 'month' and 'week' depend on'year'
+            # and 'day' depends on 'month'
+            sequential_ranges['year']['range'] = xrange(min_date.year, max_date.year + 1)
+            if period_type != 'year':
+                sequential_ranges['month']['range'] = [
```
ddohler: These are mighty list comprehensions indeed, but I think they're readable and straightforward to think about.

app/data/views.py
```diff
@@ -331,57 +332,152 @@ def _query_param_to_case_stmnt(self, param, request, queryset):
...
+                elif period_type == 'week':
+                    sequential_ranges['week']['range'] = [
+                        (year, week) for year in sequential_ranges['year']['range']
+                        for week in easy_ranges['week_of_year']['range']
+                        if min_date <= datetime.date(year, month, calendar.monthrange(year, month)[1])
+                        and datetime.date(year, month, 1) <= max_date
+                    ]
+            period = sequential_ranges[period_type]
+
+            # # Which is less ugly?
```
ddohler: I prefer the list comprehensions, actually.

app/data/views.py
ddohler: I think I was mentally prepared after our kitchen conversation though, so perhaps a comment would help future readers.

app/data/views.py
```diff
@@ -331,57 +332,152 @@ def _query_param_to_case_stmnt(self, param, request, queryset):
...
-            when_args[when_lookup] = x
+        for x in period['range']:
+            when_args = period['lookup'](x)
+            when_args['then'] = Value(str(x))
             whens.append(When(**when_args))
 
-        labels = [{'key': str(x), 'label': period_labels(x)} for x in period_range]
-        return (Case(*whens, output_field=IntegerField()), labels)
+        labels = [{'key': str(x), 'label': period['label'](x)} for x in period['range']]
+        return (Case(*whens, output_field=CharField()), labels)
```
ddohler: What was the reason for this change?

ddohler:
+1

app/data/views.py
KlaasH: Right now it knows to build up ranges based on which dictionary it finds the key in, so the cost of combining them would be needing a 'type' attribute or at least more logic looking at the param value to figure out what to do.

I'm renaming 'easy_ranges' to 'periodic_ranges', and I'm inclined to leave them separate and hope that future maintainers see that see how things are set up.  Any periodic range should probably be 'something_of_something', so I would think people wouldn't be too inclined to add periodic ranges that clobber sequential ones.

app/data/views.py
KlaasH: The output_field is the key, and formerly those were all numbers, but now a bunch of them are tuples, so I'm casting them to strings (see also the change from `Value(x)` to `Value(str(x))` just above).

KlaasH:
Added some comments and a couple basic tests.

ddohler:
:+1:

PR #452, opened 2016-03-16
Ignore gradle data folder

fungjj92:
I think this is an unnecessary local folder


flibbertigibbet:
This folder does need to exist locally.

fungjj92:
Sorry, to clarify: Necessary to exist locally. But unnecessary to handle on the repo?

PR #453, opened 2016-03-16
Spinner on loading record list

fungjj92:


KlaasH:
The onRecordsLoaded function in this controller is kind of weird.  It doesn't look at records at all.  It should probably be named onSchemaLoaded and be called on initialization and recordType change before records are loaded.  It's currently called unnecessarily on boundarystate and filterbar changes, but not called on pagination requests, which is proper but means that right now ctl.loadingRecords never gets cleared for pagination requests.

If you're not up for the bigger refactor at the moment, I would say just move the `ctl.loadingRecords = false;` into a `.finally()` on the `QueryBuilder.djangoQuery()` call inside `loadRecords`.

fungjj92:
Very helpful comments, thanks! Refactor in other PR.

KlaasH:
:+1:

PR #454, opened 2016-03-16
Refactor list controller

fungjj92:
Load schema labels once.
Remove unnecessary calls and rename function (onRecordsLoaded) so it reflect its use case better.

Follow up to comment on PR #453

### DIFF
--- a/web/app/scripts/views/record/list-controller.js
+++ b/web/app/scripts/views/record/list-controller.js
@@ -38,7 +38,7 @@
             return RecordSchemaState.get(currentSchemaId)
                 .then(function(recordSchema) {
                     ctl.recordSchema = recordSchema;
-                });
+                }).then(onSchemaLoaded);
         }
 
         /*
@@ -75,7 +75,7 @@
             });
         }
 
-        function onRecordsLoaded() {
+        function onSchemaLoaded() {
             var detailsDefinitions = _.filter(ctl.recordSchema.schema.definitions,
                 function(val, key) {
                     if (key.indexOf('Details') > -1) {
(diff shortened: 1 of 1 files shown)


web/app/scripts/views/record/list-controller.js
```diff
@@ -38,7 +38,7 @@
             return RecordSchemaState.get(currentSchemaId)
                 .then(function(recordSchema) {
                     ctl.recordSchema = recordSchema;
-                });
+                }).then(onSchemaLoaded);
```
KlaasH: Andrew pointed out on a PR of mine last week that promises behave differently depending on whether or not they're returning something.  Apparently this setup, where the `.then()` function doesn't return anything, results in that code and the next `.then()` in the chain running concurrently (or at least not being guaranteed to run sequentially).  Probably the dynamics of this setup are such that the problem would never actually arise, but I think it would be more correct/airtight to have the first `then` return recordSchema (still setting it on the controller, just also returning it) and have onSchemaLoaded accept it as an argument.

fungjj92: Ayy good catch, this is true. This is error prone.

web/app/scripts/views/record/list-controller.js
```diff
@@ -37,8 +37,8 @@
 
             return RecordSchemaState.get(currentSchemaId)
                 .then(function(recordSchema) {
-                    ctl.recordSchema = recordSchema;
-                });
+                    return ctl.recordSchema = recordSchema;
```
KlaasH: This works, but it makes jshint nervous:
<pre>app/scripts/views/record/list-controller.js
  line 40  col 46  Did you mean to return a conditional instead of an assignment?

  ⚠  1 warning

Warning: Task "jshint:all" failed. Use --force to continue.

Aborted due to warnings.</pre>

+1 after placating jshint with a separate return statement.

PR #455, opened 2016-03-21
Make recent counts filter by selected boundary

KlaasH:
Moves the recent counts API view to Record (rather than RecordType) and
makes it filter by the selected boundary ID.

### DIFF
--- a/app/data/views.py
+++ b/app/data/views.py
@@ -3,6 +3,7 @@ import logging
 import uuid
 import calendar
 import datetime
+import pytz
 
 from dateutil.parser import parse as parse_date
 from django.template.defaultfilters import date as template_date
@@ -76,6 +77,13 @@ class DriverRecordViewSet(RecordViewSet, mixins.GenerateViewsetQuery):
         if is_admin_or_writer(self.request.user) and not requested_details_only:
             return RecordSerializer
         return DetailsReadOnlyRecordSerializer
+
+    def get_filtered_queryset(self, request):
+        """Return the queryset with the filter backends applied. Handy for aggregations."""
+        queryset = self.get_queryset()
+        for backend in list(self.filter_backends):
+            queryset = backend().filter_queryset(request, queryset, self)
+        return queryset
 
     # Change auditing
     def add_to_audit_log(self, request, instance, action):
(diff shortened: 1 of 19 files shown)


KlaasH:
Now also takes on the broader issue behind why the boundary filter was missing from recent counts--that it was being handled ad hoc in many (but obviously not all) the relevant places--by moving it into QueryBuilder.  See the commit message on b2d8605 for more words on the subject.

app/data/views.py
```diff
@@ -198,6 +202,22 @@ def toddow(self, request):
...
+        now = datetime.datetime.now(tz=pytz.timezone(settings.TIME_ZONE))
+        qs = self.get_filtered_queryset(request).filter(occurred_from__lte=now)
+        durations = {
+            'month': 30,
+            'quarter': 90,
+            'year': 365,
+        }
+
+        counts = {label: qs.filter(occurred_from__gte=(now - datetime.timedelta(days=days))).count()
+                  for label, days in durations.items()}
```
KlaasH: Is there a nice way to do this with one query instead of three?  It seems similar to the other aggregations happening in the other Record views, but I don't have much experience with those tools and my attempts met with failure.

web/app/scripts/custom-reports/custom-reports-modal-controller.js
```diff
@@ -66,7 +66,9 @@
         }
 
         function assembleParams() {
-            return QueryBuilder.assembleParams(0, true, true).then(
+            // TODO: if there's a boundary selected, should that filter be applied when
+            // they haven't selected a geographical aggregation? Currently it's never applied.
```
ddohler: My gut is that it should, but that might be a good question for @jbranigan .

app/data/views.py
ddohler: Hmm, I'm waffling back and forth on this. You can't do the datetime directly in the database because `F` and `Value` objects don't support timedeltas, as far as I've been able to discover. If it was possible, what you'd have to do is pre-compute the 30-, 90- and 365-day cutoff datetimes in Python, and then put those into a `Case` statement. But I'm not sure if that would actually work because I think the "buckets" being passed to the `Case` statement may need to be mutually exclusive in order for the counts to be correct. I'd be happy to poke at this a bit tomorrow if you'd like.

app/data/views.py
```diff
@@ -77,6 +78,13 @@ def get_serializer_class(self):
             return RecordSerializer
         return DetailsReadOnlyRecordSerializer
 
+    def get_filtered_queryset(self, request):
```
ddohler: Good call, this snippet is starting to pop up a lot.

ddohler:
+1 on Travis success.

web/app/scripts/custom-reports/custom-reports-modal-controller.js
jbranigan: It's applied everywhere else, right? It would be confusing to have different behavior here.

KlaasH: Yeah, it should be applied, but there's a small nuance.  See issue #456.

PR #463, opened 2016-03-24
Redirect after login

KlaasH:
Adds a param to the login state, which gets set by the authentication interceptor, to store the page you were trying to reach and send you there once you're logged in.

I went through a process with this, trying to get it to work without a full reload, since that makes the page flash.  I got the navbar working, but then realized that's the easy one, because there are other pieces that initialize from the API and so implicitly rely on authentication but don't actually even check it.  I also tried but failed to find a way to load anything besides '/' in a way that would cause it to load fresh rather than update the state in place.

But it does get you to the page you were trying to get to, and a little flash seems better than always going to the dashboard.


flibbertigibbet:
+1

PR #464, opened 2016-03-25
Add/edit datetime fixes

KlaasH:
Makes both the front-end and the back-end produce an error when trying to save a record if the 'occurred_from' date is in the future.

I put the back-end check in DRIVER, rather than in Ashlar, since it seems more like a project-specific requirement (also changing Ashlar is a pain...), which meant creating a DriverRecordSerializer.

### DIFF
--- a/app/data/serializers.py
+++ b/app/data/serializers.py
@@ -1,6 +1,8 @@
 import re
+import datetime
+import pytz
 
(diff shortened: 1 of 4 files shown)


ddohler:
+1, looks good. I think putting the validation in DRIVER rather than Ashlar is the right call.

PR #465, opened 2016-03-25
Add satellite basemap and base layers service

KlaasH:
Adds Esri World Imagery as a basemap option.  Also, the existing base layer was being recreated in several places, so I consolidated into a BaseLayersService.  Now that we have more than one basemap option, there's a use for layer pickers on the inset maps, so I added them (showing only the basemap options.  Wouldn't really make sense to select different record layers in e.g. the "Black Spots by Severity" inset map.).

### DIFF
--- a/web/app/scripts/black-spots/black-spots-controller.js
+++ b/web/app/scripts/black-spots/black-spots-controller.js
@@ -16,7 +16,15 @@
         }
 
         function addBaseLayers(map) {
-            map.addLayer(BaseLayersService.streets());
+            var baseMaps = BaseLayersService.baseLayers();
+            map.addLayer(baseMaps[0].layer);
+
+            if(!ctl.layerSwitcher){
+                ctl.layerSwitcher = L.control.layers(
+                    _.zipObject(_.map(baseMaps, 'label'), _.map(baseMaps, 'layer'))
+                );
+                ctl.layerSwitcher.addTo(map);
+            }
             return map;
         }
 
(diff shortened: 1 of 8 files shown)


web/app/scripts/black-spots/black-spots-controller.js
```diff
@@ -18,27 +16,23 @@
...
-            TileUrlService.baseLayerUrl().then(function(url) {
-                var streets = new L.tileLayer(url, streetOptions);
-                map.addLayer(streets);
-            });
+            var baseMaps = BaseLayersService.baseLayers();
+            map.addLayer(baseMaps[0].layer);
 
+            if(!ctl.layerSwitcher){
+                ctl.layerSwitcher = L.control.layers(
+                    _.zipObject(_.map(baseMaps, 'label'), _.map(baseMaps, 'layer'))
```
flibbertigibbet: Nice

web/app/scripts/map-layers/baselayers-service.js
```diff
@@ -0,0 +1,48 @@
...
+                }
+            );
+            return layer;
+        }
+
+        function satellite() {
+            var layer = new L.tileLayer(
+                '//server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
+                {
+                    attribution: 'Tiles &copy; Esri &mdash; Source: Esri, i-cubed, USDA, USGS, AEX, GeoEye, Getmapping, Aerogrid, IGN, IGP, UPR-EGP, and the GIS User Community',
```
flibbertigibbet: Should also disable `detectRetina` here and also give it explicit `zIndex` of 1.

flibbertigibbet:
+1

Issue #456, opened 2016-03-21
Need to handle boundary filter for custom reports


KlaasH:
Currently, several different controllers call BoundaryState and save the resulting boundary UUID (if any) to polygon_id and add it to the params they pass to QueryBuilder.djangoQuery() or QueryBuilder.assembleParams().  It would be better if boundary filtering were done entirely inside those functions, controlled by (yet another) toggle argument (e.g. doBoundaryFilter.  Or possibly that whole setup could be refactored to use a config object rather than a long chain of optional booleans).

KlaasH:
The original issue ("Boundary filter should be moved into QueryBuilder") was fixed by PR #455, but one unresolved issue remains: custom reports when a boundary is selected in the navbar.

All the other existing filters are applied, and it makes sense that someone who has selected a state or region would only want a report for that state or region, but the interaction with geographic aggregation gets a little complicated.  Or rather, seems like it's a little complicated to convey to the user.  It's actually simple--apply the boundary filter if they haven't selected any geographic aggregation, don't apply it if they have.  So we could just add an entry to "Active Filters" along the lines of "Region: National Capital Region (only applied if no geographic aggregation is selected)".  Or we could make it appear and disappear depending on whether or not it will be applied.

Here's how it looks right now:
![custom_report_modal](https://cloud.githubusercontent.com/assets/6598836/14017899/e58c1626-f1a0-11e5-8ce4-efdda11c83dd.png)

jbranigan:
I think adding the geographic boundary to the active filter list is a good idea. This would also give users the option to aggregate by subunits of the primary boundary (e.g. by neighborhood or zip). If they try to aggregate by superunits, they'll get a weird report, but I think they'll have to figure out what they did wrong at that point. If we explicitly note the boundary filter, it will help them figure that out.

PR #467, opened 2016-03-28
Add boundary filter to custom report

KlaasH:
Follow-up to PR #455, discussed in issue #456.

Adds the selected boundary (if there is one) to the list of existing filters in the custom report modal, and makes the custom report query filter by boundary.

Notes:
* The text in the "Active Filters" box on the modal is just "Boundary: <state or region name>".  I would have done "State" or "Region" or whatever instead of "Boundary", but we don't have a singular label for geographies. And it seemed a little much to add one for this.
* If they have an active boundary filter and then add a geographic aggregation, they'll get a table in the report for each state/region/whatever, even if it's not within their selected boundary. We would need to add some kind of "exclude" logic to the back-end to change that.


ddohler:
+1, nice.

PR #468, opened 2016-03-28
Rename Accident to Incident thru app

fungjj92:
Lots of ctrl+f
To test, you'll have to upload the schema again and some data. Note the file name changes, as well so loading would be like...
`scripts/load_incidents_v3.py --authz 'Token YOUR_AUTH_TOKEN' /path/to/incident_csvs`


KlaasH:
You renamed pnp_incident_schema.json and pnp_incident_schema_v2.json in scripts/, but load_pnp_accidents.py and load_pnp_accidents_v2.py are still there.

README.md and scripts/README.md should be changed.

They're in comments, but probably better to change the couple occurrences of 'accident' in app/data/views.py.

fungjj92:
Thanks for the thorough search, I missed those

KlaasH:
:+1:

PR #469, opened 2016-03-28
Add endpoint to get API token from mobile SSO

flibbertigibbet:
Backend portion of mobile client SSO authentication.

https://developers.google.com/identity/sign-in/android/backend-auth#send-the-id-token-to-your-server

Adds an endpoint that accepts a single posted `token` parameter, which is the token ID obtained by the mobile client from authorizing an account with Google for use with DRIVER. This token is then validated for the app server-side, using the Google python authentication client; if it is, the endpoint returns the same response as the existing `api-token-auth` endpoint: the user ID and token for DRF authentication of API requests.

### DIFF
--- a/app/driver_auth/views.py
+++ b/app/driver_auth/views.py
@@ -68,9 +68,9 @@ def authz_cb(request):
             # authentication failed
             # return 403 here instead of raising error
             return JsonResponse({'error': 'This login is not valid in this application'},
-                            status=403)
+                            status=status.HTTP_403_FORBIDDEN)
     except OIDCError as err:
-        return JsonResponse({'error': err, 'callback': query}, status=400)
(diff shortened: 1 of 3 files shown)


app/driver_auth/views.py
```diff
@@ -69,6 +78,42 @@ def get_oidc_client_list(request):
...
+class DriverSsoAuthToken(APIView):
+    parser_classes = (JSONParser,)
+    permission_classes = (AllowAny,)
+
+    def post(self, request, format=None):
+        token = request.data.get('token')
+        if token:
+            return validate_oauth_token(token)
+        else:
+            return JsonResponse({'error': 'Token parameter is required'}, status=400)
```
ddohler: We're not always consistent about doing this, but I think it would be good to move these status codes to use the constants in the `rest_framework.status` module.

ddohler:
+1

Issue #462, opened 2016-03-23
Undefined text in saved filters


jbranigan:
Saved filters:
- We were seeing some 'undefined' text in saved filters. Possibly has to do with numerical filters being reset, yet still showing up.

Lknechtli:
fixed by #470

PR #470, opened 2016-03-29
Add text search type to saved-filter-as-html

Lknechtli:


Lknechtli:
Fixes #462

web/app/scripts/saved-filters/saved-filter-as-html.js
```diff
@@ -51,7 +51,11 @@
                         break;
 
                     default:
-                        htmlBlocks.push('Unknown rule type: ' + val._rule_type);
+                        if(key === '__searchText'){
+                            htmlBlocks.push('<strong>Text Search:</strong> '+val);
```
kshepard: Add a space after the `if`, before the `{`, and on both sides of the `+`.

kshepard:
Might as well add a test case to saved-filter-as-html.spec.js. +1 after that.

PR #472, opened 2016-03-29
Use zip instead of tar.gz for export

Lknechtli:
Can you test this for me? I'm having trouble with my celery container.

### DIFF
--- a/app/data/tasks/export_csv.py
+++ b/app/data/tasks/export_csv.py
@@ -1,7 +1,8 @@
 import csv
 import os
-import tarfile
+import zipfile
 import tempfile
+import time
 
 from django.conf import settings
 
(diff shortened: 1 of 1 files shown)


app/data/tasks/export_csv.py
```diff
@@ -55,9 +57,10 @@ def export_csv(query_key):
     # Add a directory for the schema we're outputting
     dirname = 'schema-' + str(schema.pk) + '/'
     for f, name in record_writer.get_files_and_names():
-        info = zipfile.ZipInfo(dirname + name)
+        info = zipfile.ZipInfo(filename=dirname + name, date_time=time.localtime(time.time())[:6])
```
ddohler: It'd be clearer to access the time via `struct_time`'s named parameters; I had to look it up in the documentation to figure out what this meant. https://docs.python.org/2/library/time.html#time.struct_time

app/data/tasks/export_csv.py
```diff
@@ -55,9 +57,10 @@ def export_csv(query_key):
...
         info.external_attr = 0755 << 16L
-        archive.writestr(info, f.read())
+        z = open(f.name, 'r')
```
ddohler: This would be better as a context manager so that we don't have to worry about closing the file afterward, i.e.
```python
with open(f.name, 'r') as z:
    archive.writestr(info, z.read())
```

app/data/tasks/export_csv.py
Lknechtli: so like this? 
```python
        t = time.struct_time(time.localtime(time.time()))
        info = zipfile.ZipInfo(filename=dirname + name, date_time=(
            t.tm_year, t.tm_month, t.tm_day, t.tm_hour, t.tm_min, t.tm_sec
        ))
```

ddohler:
The modification time of the directory into which the files unpack is still 1980 (though the files inside now have correct modification times); it would be nice to clean that up if possible.

+1 otherwise, I tested this and it works great.

Lknechtli:
Ack this zipfile library is a pain :tired_face:

PR #473, opened 2016-03-29
Limit CSV exports to the Details field if the user is read-only

ddohler:
Mashes the *Details info together with the record model information. I just realized that combining the two files has caused the `record_id` column to be repeated, but I think that this is ready otherwise. The way the Details info is combined with the model information is not the most elegant; I'm happy to hear alternative approaches.

Also fixes an import error in views.py that was preventing me from logging in.

### DIFF
--- a/app/data/tasks/export_csv.py
+++ b/app/data/tasks/export_csv.py
@@ -3,23 +3,25 @@ import os
 import zipfile
 import tempfile
 import time
+import StringIO
 
 from django.conf import settings
+from django.contrib.auth.models import User
 
-from celery import current_task
 from celery import shared_task
-from celery import states
 from celery.utils.log import get_task_logger
 
 from django_redis import get_redis_connection
 
 from ashlar.models import Record
 
+from driver_auth.permissions import is_admin_or_writer
+
 logger = get_task_logger(__name__)
 
 
 @shared_task(track_started=True)
-def export_csv(query_key):
+def export_csv(query_key, user_id):
     """Exports a set of records to a series of CSV files and places them in a compressed tarball
     :param query_key: A UUID corresponding to a cached SQL query which will be used to filter
                       which records are returned. This is the same key used to generate filtered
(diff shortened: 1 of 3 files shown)


app/requirements.txt
```diff
@@ -4,6 +4,7 @@ django-filter==0.11.0
 django-extensions==1.6.1
 django-cors-headers==1.0.0
 google-api-python-client==1.5.0
+oauth2client==2.0.1
```
flibbertigibbet: Why this?

ddohler: Without this line I was getting import errors from https://github.com/WorldBank-Transport/DRIVER/blob/develop/app/driver_auth/views.py#L10 when I tried to log in to the admin site, even after reprovisioning a couple times; but aside from that I'm not sure if it needs to be there, there may be a better way to solve the issue.

kshepard: I also just saw the same thing when checking out develop and provisioning (twice). oauth2client import problem that was causing my queries to fail.

flibbertigibbet: Okay. Maybe the `google-api-python-client` requirement adds nothing useful then. The docs confusingly reference the python client library, but then use `oauth2client` in the example, which I expected to be contained or a dependency; I thought I'd simply reprovisioned with the python client line added there.

flibbertigibbet: Apparently `oauth2client` used to be a submodule of `google-api-python-client`, but now isn't even a dependency. Not sure how it got in my docker container without an explicit install then.

ddohler: Okay; shall I remove the `google-api-python-client` install as well then?

flibbertigibbet: Yes, I'm pretty sure that gives us nothing then. Sorry that broke things.

ddohler: No worries! I'll go ahead and open up a separate PR so the fix isn't tied to this larger feature.

ddohler:
This has been updated to resolve the duplicate `record_id` issue.

KlaasH:
I'm seeing the record attributes and details on separate lines.  They're intended to be on the same line, right?
[Incidents-64681754.tar.gz](https://github.com/WorldBank-Transport/DRIVER/files/196380/Incidents-64681754.tar.gz)

ddohler:
Good catch; I had forgotten that Python defaults to Excel-style CSVs by default, which means DOS newlines. Should be fixed now.

KlaasH:
Looks like a new bug got rebased in.  Currently it returns "`'time.struct_time' object has no attribute 'tm_month'`", but if I change line 70 to `t.tm_year, t.tm_mon, t.tm_mday, t.tm_hour, t.tm_min, t.tm_sec` (i.e. change `tm_month`->`tm_mon` and `tm_day`->`tm_mday`), it works.

So +1 with that fix.

ddohler:
Good catch, I updated and tested locally and everything seems happy now.

PR #475, opened 2016-03-30
Fix deleting related content types

### LINKED ISSUE #411: Editor section deletion breaks schema
If a 'related content' section gets deleted from a schema in the editor, the `$ref` key in the schema `properties` section remains, but the referred-to section under `definitions` has been removed, breaking the schema.

ddohler:
The content type needs to get deleted from both the schema's
`definitions` section and its `properties` section.

Closes #411


flibbertigibbet:
:+1:

PR #479, opened 2016-04-08
Use main map state on add-record map

KlaasH:
Makes the embedded map used by the add-record view read the zoom, center, and basemap state as set by the main map, so that the map shown when someone hits the "add record" button looks similar to the one they were just seeing rather than resetting to the streets basemap and wide zoom.

Saves the basemap selection to local storage because otherwise views that spawn a new window and use the embedded map (e.g. Edit) don't get the selection.  Since those views have their own defaults for zoom and location, I didn't add those to local storage.

Other inset maps (blackspots, recent events) still don't pull anything from MapState.

(Kenny, sending this to you because you're the only other person who has PRS in their "doing" box at the moment.)

### DIFF
--- a/web/app/scripts/state/mapstate-service.js
+++ b/web/app/scripts/state/mapstate-service.js
@@ -5,15 +5,20 @@
     'use strict';
 
     /* ngInject */
-    function MapState() {
-        var filterGeoJSON, zoom, location;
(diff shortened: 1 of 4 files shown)


kshepard:
+1, I'm not set up to test at the moment, but the code looks solid.

PR #480, opened 2016-04-10
Style/improve dashboard styles

mtedeschi:
Includes:
•   New partial for the risk areas tool on the map (feel free to rename if we don't use risk areas)
•   Fix for the filter bar styling (tools on the right getting pushed down)
•   Rework media queries on the dashboard (image below) on smaller screens

screen shot 2016-04-09 at 7 36 34 pm


kshepard:
This looks good. I rebased it and will merge.

PR #481, opened 2016-04-11
Amend training input generation script to also write out file for load forecasting

kshepard:
The only difference with the load forecast training output is that it contains an additional column for precipitation, and the segments containing data are written out twice (once for aggregates where it was precipitating and once for aggregates where it wasn't precipitating).

The script has been tested with an extract from the web app (records.csv). The default parameters of the script have been updated to match the values of the extract, since that will be what's used in production.


ddohler:
+1

PR #483, opened 2016-04-13
Add foundation for internationalization support

kshepard:
This commit adds support for angular-translate, and uses the browser's language settings to serve the correct language translation json file. A few translations have been added as an example.


web/Gruntfile.js
```diff
@@ -9,6 +9,10 @@
 
 module.exports = function (grunt) {
 
+  // Add new languages here
+  // Do not put en or exclaim in this list
+  var i18nForeignLanguages = ['app/i18n/ar.json'];
```
ddohler: I don't know if written Arabic is more standardized than spoken Arabic, but my understanding is that there are wide regional variations in spoken Arabic, so this might need to be `ar-sa.json`. Maybe something the WB can answer?

web/app/scripts/app.js
```diff
@@ -31,6 +31,30 @@
...
+        // Enable sanitization of token
+        $translateProvider.useSanitizeValueStrategy('sanitizeParameters');
+
+        // Get the browser's navigator object and use for determining preferred language
+        var navigator = $windowProvider.$get().navigator;
+        var language = (navigator.languages ? navigator.languages[0] :
+                        (navigator.language || navigator.userLanguage)).substring(0,2);
+
+        // Interpret Zulu ('zu') to request dev language ('exclaim')
+        $translateProvider.preferredLanguage(language === 'zu' ? 'exclaim' : language);
```
ddohler: I've never been entirely comfortable with the selection of Zulu as a debug language given that it has ~10 million native speakers. Looking at the list of valid language codes, I think that Volapük (`vo`), Pāli (`pi`), Old Church Slavonic (`cu`), Interlingue (`ie`), or Avestan (`ae`) would be better choices. Most of them are primarily used as liturgical languages, which makes it relatively unlikely that someone would actually want to use the site in that language. Interlingue is a constructed language, similar to Esperanto but significantly less popular.

ddohler:
+1

web/Gruntfile.js
kshepard: Pretty sure you're correct. I'll go ahead and make that change. It'll be easy to rename it back later if we want it to be more general.

web/app/scripts/app.js
kshepard: That's a really good point I hadn't considered. I'll take a look through the supported languages in Chrome, and will switch to a lesser used language.

kshepard:
Actually, I'm going to merge this PR as-is, since there's a separate task for adding a language switcher to the app which will address both these concerns (don't want to make changes now that will shortly be removed).

PR #485, opened 2016-04-14
Add function to download OSM data

ddohler:
This is currently structured as a Celery task. To test, run `./scripts/manage.sh shell_plus` and then

```python
import black_spots.tasks as tasks
tasks.load_road_network()
```

Uses EPSG:3395 for projection by default because `ogr2ogr` apparently projects EPSG:3857 strictly following the EPSG specification, whereas map tiles are projected slightly differently, so if you project to EPSG:3857 the geometries appear shifted when displayed over map tiles.

### DIFF
--- a/app/black_spots/tasks/load_road_network.py
+++ b/app/black_spots/tasks/load_road_network.py
@@ -1,0 +1,54 @@
+import glob
+import os
+import shutil
+import subprocess
+
+from django.conf import settings
+
+from celery import shared_task
+from celery.utils.log import get_task_logger
+
+import requests
+
+
+logger = get_task_logger(__name__)
+
+
+@shared_task
+def load_road_network(output_root=None, output_srid='EPSG:3395'):
+    """Downloads OpenStreetMap data to a Shapefile by using an OSM extract PBF file.
+
+    Args:
+        output_root (string): Absolute path of directory to output to; default current directory
+        output_srid (string): Tranform Shapefile output to this SRID; default 'EPSG:3395'
+    Returns:
+        Path to OSM data as shapefile.
+    """
+    if not output_root:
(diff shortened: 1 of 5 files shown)


app/black_spots/tasks/load_road_network.py
```diff
@@ -0,0 +1,53 @@
...
+
+import requests
+
+
+logger = get_task_logger(__name__)
+
+
+@shared_task
+def load_road_network(output_root=None, output_srid='EPSG:3395'):
+    """Downloads OpenStreetMap data to a Shapefile by inspecting uploaded boundaries
```
kshepard: This should be updated now that grabbing the url via boundary inspection isn't possible.

kshepard:
+1, generated the shapefile succesfully. There were some errors printed to the console: `ERROR 1: fread(2268) failed on DBF file.` -- not sure if that's concerning or not.

ddohler:
I noticed those as well; in my (limited, in comparison to the scope of the data) testing, there didn't seem to be any problems with the generated shapefile. I was able to load it into QGIS and view geometry attributes, so I think the DBF file is largely correct. Some quick searching suggests that this may be related to encoding issues. It looks like there's a way to specify the encoding of the DBF file, so I'm going to give that a shot, but if it doesn't solve the problem I'm inclined to wait and see if this error causes problems for us in practice.

kshepard:
Cool, sounds like a plan.

ddohler:
I updated the subprocess command parameters to include `-lco ENCODING=UTF-8` -- this didn't actually solve the `fread` errors, but it did get rid of some UTF-8 conversion errors that were happening previously, so it seemed like a beneficial addition.

PR #487, opened 2016-04-15
Add RTL styles

jfrankl:
To view RTL mode, add class .rtl on the body element.

image

image


kshepard:
+1, this looks good. I'm going to merge it and will work on optionally adding the .rtl class as a part of the language picker task.

PR #488, opened 2016-04-18
Extract the majority of translation strings

kshepard:


web/app/scripts/custom-reports/aggregations-config.js
```diff
@@ -9,58 +9,62 @@
      */
 
     /* ngInject */
-    function AggregationsConfig($q, RecordState, RecordSchemaState, GeographyState) {
+    function AggregationsConfig($translate, $q, RecordState, RecordSchemaState, GeographyState) {
+        var timeType = $translate.instant('AGG.TIME');
```
ddohler: Is this ever user-visible? I might be mis-remembering but I thought this was passed to the API as a parameter for how to render the report. If that's the case it probably doesn't need to be translated, but if that's not right then feel free to ignore.

ddohler:
+1

web/app/scripts/custom-reports/aggregations-config.js
kshepard: It's visible in the row/column dropdowns as a group header for the time items.

PR #491, opened 2016-04-18
Add language picker

kshepard:


deployment/ansible/group_vars/all.example
```diff
@@ -54,6 +54,12 @@ local_time_zone_id: 'Asia/Manila'
 local_country_code: 'ph'
 local_center_lat_lon: [12.375, 121.5]
 osm_extract_url: 'https://download.geofabrik.de/asia/philippines-latest.osm.pbf'
+languages:
+    - { id: 'en-us', label: 'English (US)', rtl: false  }
```
ddohler: Since it looks like ordering matters, it might be good to add a comment specifying that the first item in the list will be used as the app's default language. Alternatively, a possible refactor might be to turn this into an object with language ids as the keys, and then add a `defaultLanguage` key specifying the code of the default which the LanguageState service would use when starting up.

ddohler:
+1

deployment/ansible/group_vars/all.example
kshepard: Good point, I'll add a note explaining.

PR #493, opened 2016-04-19
Move black spot geometry script into Celery task

ddohler:
This moves the black spot geometry script into a Celery task so that it can create Django BlackSpot objects in the database. It also performs buffering to make sure that geometries end up as polygons.

To test, make sure that `combined_segments.shp` (and friends) and `forecasts.csv` are in your `app` folder (they're on the fileshare). Then open a shell and run:

```python
from black_spots.tasks import load_blackspot_geoms
rtid = RecordType.objects.filter(pk='<pk for Incidents RecordType here>')[0].pk
load_blackspot_geoms('combined_segments.shp', 'forecasts.csv', rtid, 32651)
```

You should then be able to view black spots in Cebu, if your filter timespan includes today.


kshepard:
+1, ran through the test successfully.

PR #494, opened 2016-04-19
Add celery task to download records into a csv

Lknechtli:
Add django-storage, defaulted to local fs

### DIFF
--- a/app/data/tasks/fetch_record_csv.py
+++ b/app/data/tasks/fetch_record_csv.py
@@ -1,0 +1,113 @@
+import csv
+from django.core.files import File
+from collections import namedtuple
+import tempfile
+import re
+
+from celery import shared_task
+
+from ashlar.models import Record
+from black_spots.models import BlackspotRecordsFile
+
+
+# Keys which cannot be usefully exported to csv
+DROPPED_KEYS = ['media']
+# static record fields
+RECORD_FIELDS = ['record_id', 'created', 'modified', 'occurred_from',
+                 'occurred_to', 'lat', 'lon', 'location_text',
+                 'city', 'city_district', 'county', 'neighborhood', 'road',
+                 'state', 'weather', 'light']
+
+
+def FIELD_TRANSFORMS():
+    def date_iso(d):
+        return d.isoformat()
+
+    FieldTransform = namedtuple('FieldTransform', ['field', 'transform'])
+    return {
+        'record_id': FieldTransform(field='uuid', transform=lambda uuid: str(uuid)),
+        'created': FieldTransform(field='created', transform=date_iso),
+        'modified': FieldTransform(field='modified', transform=date_iso),
+        'occurred_from': FieldTransform(field='occurred_from', transform=date_iso),
+        'occurred_to': FieldTransform(field='occurred_to', transform=date_iso),
+        'lat': FieldTransform(field='geom', transform=lambda geom: geom.y),
+        'lon': FieldTransform(field='geom', transform=lambda geom: geom.x),
+    }
+
+
+@shared_task
+def export_records(occurred_min, occurred_max, record_type_id):
+    def to_utf8(s):
+        """Convert to utf8 encoding and strip special whitespace/commas for csv writing"""
+        if isinstance(s, str):
+            return re.sub(r'[\r\n\t]', '', s)
+        elif isinstance(s, unicode):
+            return re.sub(r'[\r\n\t]', '', s).encode('utf-8')
+        elif s is None:
+            return unicode('').encode('utf-8')
+        else:
+            return re.sub(r'[\r\n\t]', '', unicode(s)).encode('utf-8')
+
+    records = Record.objects.filter(
+        occurred_from__gte=occurred_min,
+        occurred_to__lte=occurred_max,
+        schema__record_type_id=record_type_id
+    )
+
+    try:
+        record_type = records[0].schema.record_type
+        schema = record_type.get_current_schema()
+    except IndexError:
+        raise Exception('No records in result set')
+
+    jsonschema = schema.schema
+    details = {
+        key: subschema for key, subschema in jsonschema['definitions'].viewitems()
+        if 'details' in subschema and subschema['details'] is True
+    }
+    details_key = details.keys()[0]
+    record_detail_fields = [
+        key for key, val in details[details_key]['properties'].viewitems()
+        if (
+                'options' not in val or
+                'hidden' not in val['options'] or
+                val['options']['hidden'] is False
+        ) and (
+            key not in DROPPED_KEYS
+        )
+    ]
+    csv_columns = RECORD_FIELDS + record_detail_fields
+    transforms = FIELD_TRANSFORMS()
+    row_dicts = []
+    for record in records:
+        row = dict()
+        for field in RECORD_FIELDS:
+            if field in transforms:
+                ft = transforms[field]
+                row[field] = to_utf8(ft.transform(getattr(record, ft.field)))
+            else:
+                row[field] = to_utf8(getattr(record, field))
+        for field in record_detail_fields:
+            if field in transforms:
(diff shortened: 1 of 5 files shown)


app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,96 @@
...
+        'lon': FieldTransform(field='geom', transform=lambda geom: geom.x),
+    }
+
+
+@shared_task
+def export_records(occurred_min, occurred_max):
+    records = Record.objects.filter(
+        occurred_from__gte=occurred_min,
+        occurred_to__lte=occurred_max,
+        schema__record_type__label=settings.PRIMARY_RECORD_LABEL
```
ddohler: I think it would be good to keep this function generic by passing in a `record_type_id`; that will make it easy for us to run the script against a different record type if necessary. Right now I don't think that's likely to happen but the general trend on this project has been that a lot of features have ended up being used with multiple record types even when we didn't think they would initially.

Lknechtli:
Example output csv
[dd75cb64-2d70-4a17-aaae-ddee7a57fc88.csv.zip](https://github.com/WorldBank-Transport/DRIVER/files/228443/dd75cb64-2d70-4a17-aaae-ddee7a57fc88.csv.zip)

app/data/tasks/fetch_record_csv.py
```diff
@@ -37,11 +37,22 @@ def date_iso(d):
...
+    def to_utf8(s):
+        """Convert to utf8 encoding and strip special whitespace/commas for csv writing"""
+        if isinstance(s, str):
+            return re.sub(r'[\r\n\t,]', '', s)
+        elif isinstance(s, unicode):
+            return re.sub(r'[\r\n\t,]', '', s).encode('utf-8')
+        elif s is None:
+            return unicode('').encode('utf-8')
+        else:
+            return re.sub(r'[\r\n\t,]', '', unicode(s)).encode('utf-8')
```
ddohler: Python's CSV writer should handle commas by wrapping them in quotes, which most CSV readers will correctly understand -- were commas causing problems?

Also, it might be helpful to assign the regex to a variable up top, that way if someone needs to edit it they don't have to change three different places.

app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,111 @@
...
+    ]
+    csv_columns = RECORD_FIELDS + record_detail_fields
+    transforms = FIELD_TRANFORMS()
+    row_dicts = []
+    for record in records:
+        row = dict()
+        for field in RECORD_FIELDS:
+            if field in transforms:
+                ft = transforms[field]
+                row[to_utf8(field)] = to_utf8(ft.transform(getattr(record, ft.field)))
```
ddohler: Wrapping `field` in `to_utf8()` shouldn't be necessary since they're all hard-coded up top and could potentially introduce unexpected behavior since `to_utf8` also deletes certain characters (which might be a reason to split that functionality out, not sure).

app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,111 @@
...
+        'modified': FieldTransform(field='modified', transform=date_iso),
+        'occurred_from': FieldTransform(field='occurred_from', transform=date_iso),
+        'occurred_to': FieldTransform(field='occurred_to', transform=date_iso),
+        'lat': FieldTransform(field='geom', transform=lambda geom: geom.y),
+        'lon': FieldTransform(field='geom', transform=lambda geom: geom.x),
+    }
+
+
+@shared_task
+def export_records(occurred_min, occurred_max, record_label=settings.PRIMARY_RECORD_LABEL):
```
ddohler: Just to re-emphasize, I think this needs to be a `record_type_id` (not label); `RecordType.label` is not guaranteed to be unique (and in fact often isn't -- by default, running the event import script multiple times will create RecordTypes with duplicate labels).

app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,111 @@
...
+# Keys which cannot be usefully exported to csv
+DROPPED_KEYS = ['media']
+# static record fields
+RECORD_FIELDS = ['record_id', 'created', 'modified', 'occurred_from',
+                 'occurred_to', 'lat', 'lon', 'location_text',
+                 'city', 'city_district', 'county', 'neighborhood', 'road',
+                 'state', 'weather', 'light']
+
+
+def FIELD_TRANFORMS():
```
ddohler: Typo, missing 'S' in TRANSFORMS.

app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,111 @@
...
+        for field in RECORD_FIELDS:
+            if field in transforms:
+                ft = transforms[field]
+                row[to_utf8(field)] = to_utf8(ft.transform(getattr(record, ft.field)))
+            else:
+                row[to_utf8(field)] = to_utf8(getattr(record, field))
+        for field in record_detail_fields:
+            if field in transforms:
+                ft = transforms[field]
+                if ft.field in record.data['incidentDetails']:
```
ddohler: Is it possible for this to make use of the `details_key` logic above so that it works for any record type?

app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,111 @@
...
+                    row[to_utf8(field)] = to_utf8(ft.transform(record.data['incidentDetails'][ft.field]))
```
ddohler: Does there need to be an 
```python
else:
    row[field] = ''
```
here as well (like below)?

app/data/tasks/fetch_record_csv.py
```diff
@@ -0,0 +1,111 @@
...
+                else:
+                    row[to_utf8(field)] = ''
+        row_dicts.append(row)
+
+    with tempfile.NamedTemporaryFile() as csvfile:
+        writer = csv.DictWriter(csvfile, fieldnames=csv_columns)
+        writer.writeheader()
+        writer.writerows(row_dicts)
+        store = BlackspotRecordsFile()
+        csvfile.seek(0, 0)
```
ddohler: I'm assuming this is so that the `store.csv.save()` call below can read the whole file. Might be worth a comment, this threw me for a second.

app/driver/settings.py
```diff
@@ -110,6 +111,12 @@
     map(int, os.environ.get('DJANGO_POSTGIS_VERSION', '2.1.3').split("."))
 )
 
+# File storage
+DEFAULT_FILE_STORAGE = 'storages.backends.overwrite.OverwriteStorage'
+
+# Primary record label
+PRIMARY_RECORD_LABEL = os.environ.get("DRIVER_PRIMARY_RECORD_LABEL", "Incident")
```
ddohler: If the celery task function is switched to take a `record_type_id` then this and the changes to `deployment/ansible/roles/driver.app/defaults/main.yml` below can be removed.

ddohler:
Looking good so far, just needs a bit of cleanup.

app/driver/settings.py
Lknechtli: yea. I'm just going to remove the defaults and require specifying it.

app/data/tasks/fetch_record_csv.py
Lknechtli: None of the field transform lambdas can return `None` as a result. I could add so that if someone changes a transformer to do so it doesn't act weird.

Lknechtli: wait nvm. see what you mean. I'll add it.

app/data/tasks/fetch_record_csv.py
Lknechtli: good catch. that was fragile.

ddohler:
+1 after squashing cleanup commits.

PR #495, opened 2016-04-20
Add optimizations for intersection discovery

kshepard:
This commit takes the bounds of the input shapefile and divides
it up into equal rectangles (determined by the `tile-max-units` arg),
and then processes each tile in parallel via multiprocessing.


notthatbreezy:
+1, looks good to me

PR #496, opened 2016-04-21
Fix checkbox field aggregation

### LINKED ISSUE #492: Ensure checkbox types work throughout the app
Currently if a field (such as Severity) is changed from a select list to a checkbox, using it to generate a custom report fails with:
`Error: The property at choices_path is missing required 'enum' field`

This may be the only place where it fails, but it'd be good to fix and review other places that could potentially fail.

kshepard:
Closes #492


ddohler:
+1, glad that wasn't too hard.

PR #497, opened 2016-04-21
Order Records list by occurred_from

ddohler:
Overrides the default Ashlar ordering (which is to order by `'-created'`). I initially was planning to put this in Ashlar directly, but I realized that sorting by creation date is probably sane for many uses, so I decided to put the change into DRIVER instead absent a compelling reason to change the default.


KlaasH:
Makes sense to me.  +1

PR #498, opened 2016-04-21
Fix adding related references

### LINKED ISSUE #490: Javascript error when adding a schema with a relationship field
I tried adding a new record type with a schema containing Vehicles and People and creating a relationship field linking Vehicles and People, but there was an error:
`TypeError: Cannot read property 'properties' of undefined`

ddohler:
Fixes #490. The reference fields in the editor hadn't been updated to use the new camelcased keys yet.


kshepard:
+1

PR #500, opened 2016-04-25
Add blackspot identification script

jerheff:
This R script builds a statistical model of accident activity and then uses it
to create forecasts for future activity.


kshepard:
+1, merging this in and we'll work on installing requirements and integrating it into the workflow.

PR #502, opened 2016-04-27
Extract more translation strings

kshepard:
This should be all the text strings. It doesn't yet account for
datetime localization, which will be handled in another task.
Some TODOs have been added for pieces that will require datetime
localization.

Also, some minor cleanup has been done to remove jshint warnings.


web/app/scripts/custom-reports/custom-report-partial.html
```diff
@@ -7,14 +7,20 @@
...
             <h3>
-                Incidents
-                <span ng-if="ctl.params.aggregation_boundary"> for {{ ::ctl.report.table_labels[table.tablekey] }}</span>
-                by {{ ::ctl.rowCategoryLabel }} and {{ ::ctl.colCategoryLabel }}
+                {{ ctl.recordType.plural_label }}
+                <span ng-if="ctl.params.aggregation_boundary">
+                  {{ 'COMMON.FOR' | translate }}
+                  {{ ::ctl.report.table_labels[table.tablekey] }}
+                </span>
+                {{ 'COMMON.BY' | translate }}
```
ddohler: This may need to be revisited; "by" is a weird word that has a lot of different meanings in different contexts. I'm not sure how Arabic handles things, but at least in the languages I know, "by" tends not to have a clean 1-to-1 mapping with another word (and it doesn't always map to a word at all, sometimes it's a prefix / suffix to another word).

kshepard: Yeah, I definitely agree, already discussed that (and other such terms) with Tyler. We'll have to discuss with a native speaker and figure out the best way to proceed.

ddohler: I talked to Arianna and the solution we agreed on is that we should change these to be more specific keys, something like `CUSTOM_REPORT.AREA_FOR`, `CUSTOM_REPORT.CATEGORY_BY`, and `CUSTOM_REPORT.CATEGORY_AND`. She said there may still be some issues if the word `by` changes depending on which type of categories we have, but we can work that out with the translators.

ddohler:
+1 after updating custom report keys.

web/app/scripts/custom-reports/custom-report-partial.html
kshepard: Nice, thanks for looking into that.

PR #505, opened 2016-04-28
Create Docker image for running R scripts

ddohler:
Also sets up docker inside of the app container so that Celery tasks can
run docker images via a `subprocess` call. For example (from inside
Python shell in celery container):
```python
import subprocess

cmd = ['docker',
    'run',
    '--rm',
    '--volumes-from',
    'driver-celery',
    'quay.io/azavea/driver-analysis',
    'ls',
    '/opt/app'
]
subprocess.check_call(cmd)
```

Result:
```
black_spots
data
dev-requirements.txt
Dockerfile.base
Dockerfile.development
driver
driver_auth
manage.py
requirements.txt
user_filters
```

The `--volumes-from` parameter is important; it guarantees that the R
process has access to the same volumes as the celery process.  This
allows the celery task to pass absolute file paths referenced from
within its own file system directly to the R script as parameters and
ensures that any files created by the R container will be persistent
from the perspective of the celery container.

@hectcastro may also want to take a look.


ddohler:
The best way to test is probably:
0. `vagrant provision celery`
1. `vagrant ssh celery` 
2. `sudo docker exec -ti driver-celery ./manage.py shell_plus`
3. Run the below Python:
```python
import subprocess

cmd = ['docker',
    'run',
    '--volumes-from',
    'driver-celery',
    'quay.io/azavea/driver-analysis',
    'Rscript',
    'blackspots_model.R',
    '--help'
]
subprocess.check_call(cmd)
```

app/Dockerfile.base
```diff
@@ -16,6 +11,16 @@ RUN apt-get update && apt-get install -y \
     libspatialindex-dev \
     libgdal1-dev
 
+# Dependencies for installing docker locally
+RUN apt-get update && apt-get install -y \
+    curl
+RUN curl -L https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 > /usr/bin/docker && chmod +x /usr/bin/docker
```
hectcastro: It may be useful to pull the `1.10.3` bit out into an environment variable so that it is easier to update later.

analysis_tasks/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN R --slave --silent --vanilla \
+    -e 'install.packages(c("optparse", "caret", "plyr", "gbm", "doParallel"), repos="http://cran.rstudio.com")'
+
+RUN mkdir -p /opt/analysis_tasks
+WORKDIR /opt/analysis_tasks
+
+COPY . /opt/analysis_tasks
+
+# Provide access to the R environment by default if run without parameters
+CMD ["R", "--vanilla"]
```
hectcastro: Changing this to use `ENTRYPOINT` and `CMD` might make simpler to run the container later. Say if it was always meant to be launched with `Rscript`, then all you'd need to supply to run would be the name of a file.

.travis.yml
```diff
@@ -21,6 +21,7 @@ install:
   ./web
 - docker build -f ./gradle/Dockerfile -t "quay.io/azavea/driver-gradle:${TRAVIS_COMMIT:0:7}"
   ./gradle
+- docker build -f ./analysis_tasks/Dockerfile -t "quay.io/azavea/driver-analysis:${TRAVIS_COMMIT:0:7}"
```
hectcastro: I think that a path is missing on the end. The others seem to have the path on a new line.

hectcastro:
I like the conciseness of `--volumes-from`, but worry that it'll mount the volumes read/write. Main concern there is that something in the R container will muck with the data in the volumes, which could compromise things happening inside the Celery container.

I also think using `--rm` in the run command would be useful as a form of automatic cleanup.

analysis_tasks/Dockerfile
ddohler: I initially had it that way, but my understanding of the way `ENTRYPOINT` works is that it would make it more difficult to run the container with `R` because it would run `Rscript R <args>` by default, so I was thinking it would be easier to use for both debugging and non-interactive tasks if it didn't define a custom entrypoint. It'll also make it a bit clearer what's happening when a container is run since the full command won't be split across two files and there are at least two reasonable default choices in this case.

.travis.yml
ddohler: Good catch, I missed that.

analysis_tasks/Dockerfile
hectcastro: Sounds fair. 👍

ddohler:
Good call about `--rm`; I've updated the two example commands to include `--rm`. Agreed about the chance for the R script to muck with the Celery container. I think the right solution would be to create a data-only container that can be shared between the two containers, but I'm sensitive to the time that would take to set up.

hectcastro:
👍 It'll be interesting to see how this plays out in practice.

kshepard:
+1, nice. Tested successfully.

PR #506, opened 2016-04-28
Fix race condition with translations

kshepard:
The translations performed in javascript with `$translate.instant`
need to ensure the translation file is ready at the time of translation.
`InitialState` is an object that many controllers rely on in order
to wait for all basic resources to be available before initialization.
`InitialState` has been modified to also check for translations being
ready. This covers the majority of potential translation race conditions.

There are a couple modules that aren't using `InitialState` and also
rely on `$translate.instant`. In these cases, `$translate.onReady` is
being used directly. It's hard to test these locally, since translation
files are loaded so quickly, the race condition is not apparent. I'll
test this once it makes its way to staging and will fix any additional
sections that show themselves, if there are any.

### DIFF
--- a/web/app/scripts/custom-reports/aggregations-config.js
+++ b/web/app/scripts/custom-reports/aggregations-config.js
@@ -80,7 +80,8 @@
                     return RecordSchemaState.get(recordType.current_schema);
                     /* jshint camelcase: false */
                 });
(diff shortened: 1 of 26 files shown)


ddohler:
+1

PR #510, opened 2016-05-03
Add health organization API use case doc

kshepard:
@jbranigan may also want to take a look at this

### DIFF
--- a/doc/api-use-case-health-organization.md
+++ b/doc/api-use-case-health-organization.md
@@ -1,0 +1,312 @@
+# API Use Case &mdash; Health Organization
+
+This document illustrates a use case for the DRIVER API where a health organization needs to interact with the API in order to perform a couple different tasks:
+
+1. Obtain a list of recent accidents where a person involved was fatally injured. The results of this query may then be used to populate another system with this fatality information.
+2. Search accidents for people in order to update their injury status within DRIVER. This is useful if a fatality was not observed at the scene of the accident, but occurred later within a hospital.
+
+
+## Obtaining an API key
+
+The first step in interacting with the API is to obtain an API key. The tasks discussed in this document require a user with a minimum of `analyst` credentials, which allow viewing the full set of record details (including information about involved people), as well as submitting an updated record back to the application.
+
+Once logged into the web application, click your user email in the top navigation bar. Select `My Account` from the drop-down and the `Token` will be displayed when the page loads. Copy this token down, and use it in place of the `TOKEN` placeholder in the following examples.
+
+
+## Schema assumptions
+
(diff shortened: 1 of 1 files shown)


doc/api-use-case-health-organization.md
```diff
@@ -0,0 +1,276 @@
...
+
+1. Obtain a list of recent accidents where a person involved was fatally injured. The results of this query may then be used to populate another system with this fatality information.
+2. Search accidents for people in order to update their injury status within DRIVER. This is useful if a fatality was not observed at the scene of the accident, but occurred later within a hospital.
+
+
+## Obtaining an API key
+
+The first step in interacting with the API is to obtain an API key. The tasks discussed in this document require a user with a minimum of `analyst` credentials, which allow viewing the full set of record details (including information about involved people), as well as submitting an updated record back to the application.
+
+Once logged into the web application, click your user email in the top navigation bar. Select `My Account` from the drop-down and the `Token` will be displayed when the page loads. Copy this token down, and use it in place of the `TOKEN` placeholder in the proceeding examples.
```
ddohler: Perhaps proceeding -> following?

doc/api-use-case-health-organization.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Schema assumptions
+
+It is important to note that the DRIVER schema is dynamic and fields may change depending on how the system administrator has configured the schema. This document makes the following assumptions about a schema that may be used to enter accident information, but actual queries may need to be updated depending on the exact schema configuration:
+
+1. The record type is named `Incident`.
+2. There are two relevant sections of information within the `Incident` type: `incidentDetails`, and `person` (which is an array type containing one entry for each person involved).
+3. The `incidentDetails` section contains a field for `Severity` &mdash; this field signifies the overall severity level of the incident and has an option for `Fatal`.
+4. The `person` section contains a field for `Injury` which has an option for `Fatal`. It also has some other text fields for identifying a person: `Name`, `Address`, `License Number`, and `Hospital`.
+
```
ddohler: It might be good to add a short overview of the task at about this point, so that readers have a high-level sense of the sub-tasks involved, something like

1. Determine record type ID
2. View schema configuration (optional)
3. Query for records with severity FATAL
4. etc.

doc/api-use-case-health-organization.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+If the `PATCH` was successful, the updated record will be returned in the response, and it should contain the updates that have been made.
```
ddohler: There could also be a mention of status codes here as a way to determine update success.

ddohler:
+1, looks good to me with maybe a few changes for user-friendliness. I also think a troubleshooting section would be good, although I don't think we have a good sense of the most common errors at this point; maybe we can just update the document after we see what the real problems are.

kshepard:
Thanks for the feedback, I'll make those changes, and will add a short section for troubleshooting that we can flesh out as things come up.

PR #511, opened 2016-05-03
Feature/finish segment breakout

Lknechtli:


app/black_spots/models.py
```diff
@@ -47,3 +49,16 @@ class RoadSegmentsShapefile(AshlarModel):
...
+
+class BlackSpotTrainingCsv(AshlarModel):
+    """Model to track blackspot training csvs"""
+
+    #: Path to csvs
+    csv = models.FileField(upload_to='training/blackspot/')
+
+
+class ForecastTrainingCsv(AshlarModel):
+    """Model to track forcast training csvs"""
```
ddohler: Typo: forecast

app/black_spots/tasks/generate_training_input.py
```diff
@@ -0,0 +1,355 @@
...
+
+from black_spots.models import (
+    BlackSpotRecordsFile, RoadSegmentsShapefile, BlackSpotTrainingCsv, ForecastTrainingCsv
+)
+
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
+logger = logging.getLogger()
+logger = get_task_logger(__name__)
+
+INTERSECTION_BUFFER_UNITS = int(os.getenv('INTERSECTION_BUFFER_UNITS', '5'))
```
ddohler: It looks to me like a few of these are never used. My linter isn't complaining but a search for `INTERSECTION_BUFFER_UNITS` for example, only turns up the two matches on this line. Same for a few of these other constants.

app/black_spots/tasks/generate_training_input.py
```diff
@@ -0,0 +1,355 @@
...
+@shared_task
+def get_training_input(segments_shp_uuid, records_csv_uuid):
+    """Generate training input
+
+    Output to training csv to django-storages
+
+    Args:
+    :param segments_tar_uuid: (str) UUID for the RoadSegmentsShapefile to use
+    :param records_csv_uuid: (str) UUID for the BlackSpotRecordsFile
+    """
```
ddohler: This docstring could use a `Returns:` section.

app/black_spots/tasks/generate_training_input.py
Lknechtli: I largely copied and pasted the constants from get_segments. I suspect that we'll want to to move them all to django.settings once we have all the scripts working

app/black_spots/tasks/generate_training_input.py
```diff
@@ -0,0 +1,361 @@
...
+        schema, segments_with_data = get_segments_with_data(
+            seg_shapes, segments_with_records, min_occurred, max_occurred
+        )
+        _, segments_with_data_precip = get_segments_with_data(
+            seg_shapes, segments_with_records_precip, min_occurred, max_occurred
+        )
+        _, segments_with_data_no_precip = get_segments_with_data(
+            seg_shapes, segments_with_records_no_precip, min_occurred, max_occurred
+        )
+    except:
```
ddohler: I think you can omit the `except` in this case and just have `finally`.

app/black_spots/models.py
```diff
@@ -47,3 +49,16 @@ class RoadSegmentsShapefile(AshlarModel):
...
+
+
+class BlackSpotTrainingCsv(AshlarModel):
+    """Model to track blackspot training csvs"""
+
+    #: Path to csvs
+    csv = models.FileField(upload_to='training/blackspot/')
+
+
+class ForecastTrainingCsv(AshlarModel):
```
ddohler: It might be good to make this `LoadForecastTrainingCsv`; I was a bit confused about the meaning since the blackspot script itself outputs a `forecasts.csv` by default.

app/black_spots/tasks/generate_training_input.py
Lknechtli: If I do that, it'll just throw another exception down the road on line 116 since `segments_with_data` won't be defined. Better to have it raise the initial exception for debugging purposes.

ddohler: I think the original exception will still be raised if the `except` is omitted:
```
In [4]: try:
    foo = 1/0
finally:
    print("Ran finally")
   ...:     
Ran finally
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-4-9d726c13a117> in <module>()
      1 try:
----> 2     foo = 1/0
      3 finally:
      4     print("Ran finally")
      5 

ZeroDivisionError: integer division or modulo by zero
```

app/black_spots/models.py
```diff
@@ -47,3 +49,16 @@ class RoadSegmentsShapefile(AshlarModel):
 
     #: Path to gzipped shapefile
     shp_tgz = models.FileField(upload_to='road_segments/')
+
+
+class BlackSpotTrainingCsv(AshlarModel):
```
ddohler: Looks like this PR is missing migrations for these two models.

ddohler:
Looks good to me after adding migrations; @kshepard can you also give this a look over since you're more familiar with the code mapping events to segments?

app/black_spots/tasks/generate_training_input.py
Lknechtli: right you are. I'll do that then.

app/black_spots/tasks/generate_training_input.py
```diff
@@ -0,0 +1,361 @@
...
+                    data['precip'] = precip_val
+                    csv_writer.writerow(data)
+
+        write_data(segments_with_data_precip, 1)
+        write_data(segments_with_data_no_precip, 0)
+
+        # Reset file pointer to start so it can be read into a ForecastTrainingCsv
+        csv_file.seek(0, 0)
+        training_csv = ForecastTrainingCsv.objects.create()
+        training_csv.csv.save(str(training_csv.uuid)+'.csv', File(csv_file))
```
kshepard: Use `.format` to create this string.

app/black_spots/tasks/generate_training_input.py
```diff
@@ -0,0 +1,361 @@
...
+            RECORD_COL_PRECIP, RECORD_COL_PRECIP_VALS
+        )
+    records_with_precip = [record for record in records if record['precip']]
+    records_without_precip = [record for record in records if not record['precip']]
+
+    # segments shapefile
+    tar_output_dir = tempfile.mkdtemp()
+    try:
+        shp_tar = RoadSegmentsShapefile.objects.get(uuid=segments_shp_uuid).shp_tgz.path
+        with tarfile.open(shp_tar, "r:gz") as tar:
```
kshepard: Probably switch this to single quotes, since every other string in the file uses single quotes.

kshepard:
+1, looks good to me too. I'm not in a good place to test it at the moment, but it looks like all the necessary logic was transferred over to the new task correctly.

Lknechtli:
tested after refactor, works fine still

ddohler:
+1 for refactor changes.

PR #512, opened 2016-05-04
Ensure batch cron jobs are not run multiple times

kshepard:
This commit adds an initial heimdall configuration. Heimdall is a program
that wraps an executable program inside of an exclusive lock provided by
PostgreSQL's `pg_try_advisory_lock`. This results in the ability to wrap
cron jobs in heimdall across multiple worker machines and have only one
instance of the job run.

Note: this is only the initial configuration to set up heimdall. When the final 
commands for calculating black spots are finalized, this will need to be 
modified to account for them.


ddohler:
+1, looks good -- good to wrap the duplicate detection job as well.

kshepard:
Cool, going to merge this, ensure the initial setup works on staging, and the actual cron commands can be implemented in a separate PR when they're finalized.

PR #513, opened 2016-05-05
Add UI feature flags to web-config.js

ddohler:
Adds feature flags to the WebConfig object, which can be used to hide / show features in the UI.


kshepard:
+1

PR #514, opened 2016-05-05
Add translation strings for json-editor components

kshepard:
Translations have been added for the four buttons that are displayed
in the json-editor. The json-editor version was updated as part of
this in order to be able to use some of the new features such as
translation string overrides. Also, some new configuration objects
were added for button display, so some of them have been removed,
and the text has been switched to glyphicons, in order to declutter
the interface a little.

Screenshot:
json-editor-translations


ddohler:
+1, nice, glad this ended up being so easy.

PR #516, opened 2016-05-05
Don't display heatmap, blackspots, interventions if disabled

ddohler:
If you haven't reprovisioned since #513 was merged, it will be necessary to update your `group_vars/all` with the feature flags added in that PR, and then provision before testing this branch. 

To test, run `./scripts/grunt.sh web serve` and then edit `web/app/scripts/config.js` and change the `visible` flags for blackspots, interventions, and heatmaps to `false`. Note that in order to disable Interventions from showing up in the Map layer picker, it will also be necessary to make the following change:
```js
        recordType: {                                                           
            visible: false,                                                     
            primaryLabel: 'Accident',                                           
            secondaryLabel: '' //'Intervention'                                 
        },                                                                      
```

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -4,7 +4,7 @@
     /* ngInject */
     function DriverLayersController(
         $q, $filter, $log, $scope, $rootScope, $timeout, $translate, $compile,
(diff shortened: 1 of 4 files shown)


kshepard:
+1, nice.

PR #517, opened 2016-05-06
Fix graph labels when in right-to-left mode

kshepard:
The labels were being drawn off-screen in RTL-mode.
This adjusts their placement slightly so they show up.
It doesn't change anything else about placement or
alignment, as there are open implementation questions.

graphs-ltr-stepwise

graphs-rtl-stepwise

graphs-ltr-toddow

graphs-rtl-toddow


ddohler:
+1, cool!

PR #519, opened 2016-05-09
Draw chart data & labels right-to-left in RTL-mode

kshepard:
Left-to-right:
charts-ltr

Right-to-left:
charts-rtl


ddohler:
+1, looks good

PR #523, opened 2016-05-13
Create script to run through all blackspot phases

ddohler:
Lots of miscellaneous fixes, optimizations, and hole-plugging to get the blackspots process to run seamlessly from the beginning.

Currently, this script cannot complete on the full Philippines data set,
under the memory constraints that we have; more optimization is needed
in the training output generation in order to ensure that it can
complete successfuly with only 8GB of RAM. I believe that this script
can successfully run up to that point.


app/black_spots/tasks/calculate_black_spots.py
```diff
@@ -1,13 +1,69 @@
...
+        occurred_min (datetime.datetime): Earliest event to include
+        occurred_max (datetime.datetime): Latest event to include
+    """
+    # Get the parameters we'll use to filter down the records we want
+    now = timezone.now()
+    oldest = now - datetime.timedelta(days=4 * 365 + 1)
+    # Note that this assumes that there is only one RecordType with this label; there's
+    # no straightforward way to auto-detect this otherwise.
+    record_type = RecordType.objects.filter(label=settings.BLACKSPOT_RECORD_TYPE_LABEL).first()
+    roads_srid = 3395
```
kshepard: Should this be configurable?

app/black_spots/tasks/calculate_black_spots.py
```diff
@@ -1,13 +1,69 @@
...
-
+    """Integrates all black spot tasks into a pipeline
+    Args:
+        record_type_id (UUID): The record type to train black spots for
+        occurred_min (datetime.datetime): Earliest event to include
+        occurred_max (datetime.datetime): Latest event to include
+    """
+    # Get the parameters we'll use to filter down the records we want
+    now = timezone.now()
+    oldest = now - datetime.timedelta(days=4 * 365 + 1)
```
kshepard: Should probably make this configurable as well.

app/black_spots/tasks/get_segments.py
```diff
@@ -301,9 +337,11 @@ def write_segments_shp(segments_shp_path, road_projection, segments):
         }
     }
 
+    count = 0
+    print("Opening shapefile for output")
```
kshepard: These should for `logger` statements.

kshepard:
+1. Unfortunately I don't have the time or free memory to take it for a spin at the moment, but it looks good to me.

ddohler:
Updated to clean up after comments.

PR #525, opened 2016-05-17
Allow for using `production` group_vars/inventory

kshepard:
This allows using group_vars and inventory files with the name
of `production`. Using a `production` file previously wouldn't
work as expected, because all the required tasks weren't run.
The alternative to this was hijacking the `staging` group_vars
and inventory files for a production deploy, which is error-prone.


ddohler:
+1 nice.

PR #527, opened 2016-05-18
Add monit service monitoring

ddohler:
This installs monit on all VMs and sets it up to monitor all the services that we run. Although there are a lot of files here, the general gist is that I added a role to install monit, and then added configuration files to the various service roles that enable monit to monitor the services activated by each role. Monit doesn't attempt to try to restart services if they fail since we usually handle that within upstart, so this is strictly to provide a relatively straightforward interface that can be used by admins to view service status and basic system statistics.

Unless / until we get SSL going, it's not safe to open monit to the outside world since the authentication passwords are sent in plaintext, so I've limited monit to be accessible from localhost only by default; additional access sources can be added by group_vars.

@hectcastro may also want to take a quick look although I think the primary thing that's interesting here is the strategy for determining if a docker container is running from monit. Docker's equivalent of pidfiles don't really seem to be usable as such because they don't automatically disappear when a container is stopped.

### DIFF
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -79,3 +79,14 @@ keystore_password: 'driverkeystore'
 
 ## Django allowed hosts setting. Override when not in DEBUG mode.
 allowed_host: ''
+
+## Access settings for Monit admin interface. Note that this is
+## insecure; ideally, SSL should be used (which Monit supports).
+## An intermediate step would be to restrict the IPs from which
+## monit can be accessed via monit_allow_hosts
+## Username and Password MUST be defined in order for monit to work
+## but monit is restricted to requests coming from localhost by default
+monit_allow_hosts: []
+monit_allow_user: 'test'
+monit_allow_password: 'test'
+monit_enable_web_access: yes
--- a/deployment/ansible/roles/driver.database/templates/monit-postgresql.cfg.j2
+++ b/deployment/ansible/roles/driver.database/templates/monit-postgresql.cfg.j2
@@ -1,0 +1,3 @@
+check process postgresql with pidfile "/var/run/postgresql/{{ postgresql_version }}-main.pid"
+  start program = "/etc/init.d/postgresql start"
+  stop program = "/etc/init.d/postgresql stop"
--- a/deployment/ansible/roles/driver.monit/tasks/main.yml
+++ b/deployment/ansible/roles/driver.monit/tasks/main.yml
@@ -1,0 +1,49 @@
+---
+- name: Download monit
+  get_url: >
+    url=https://mmonit.com/monit/dist/binary/{{ monit_version }}/monit-{{ monit_version }}-linux-x64.tar.gz
+    dest=/usr/local/src/monit-{{ monit_version }}-linux-x64.tar.gz
+
+- name: Extract monit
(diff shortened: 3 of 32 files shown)


deployment/ansible/roles/driver.database/templates/monit-postgresql.cfg.j2
```diff
@@ -0,0 +1,3 @@
+check process postgresql with pidfile "/var/run/postgresql/9.4-main.pid"
```
kshepard: Probably want to use `{{ postgresql_version }}` here to make future version upgrades less error-prone.

deployment/ansible/group_vars/all.example
```diff
@@ -79,3 +79,14 @@ keystore_password: 'driverkeystore'
...
 ## Django allowed hosts setting. Override when not in DEBUG mode.
 allowed_host: ''
+
+## Access settings for Monit admin interface. Note that this is
+## insecure; ideally, SSL should be used (which Monit supports).
+## An intermediate step would be to restrict the IPs from which
+## monit can be accessed via monit_allow_hosts
+## Username and Password MUST be defined in order for monit to work
+## but monit is restricted to requests coming from localhost by default
+monit_allow_hosts: []
```
kshepard: Cool, I like how this is made configurable. Setting it to the couple IPs of the small number of people who will potentially access monit will be a pretty good interim solution to start with.

kshepard:
+1, though I'm not familiar with `docker top` and don't know if there are any edge cases we need to worry about.

deployment/ansible/roles/driver.monit/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
+---
+- name: Download monit
+  get_url: >
+    url=https://mmonit.com/monit/dist/binary/{{ monit_version }}/monit-{{ monit_version }}-linux-x64.tar.gz
```
hectcastro: Just wanted to ask why we'd use this vs. the version in the Ubuntu repositories?

```bash
vagrant@vagrant-ubuntu-trusty-64:~$ sudo apt-cache policy monit
monit:
  Installed: (none)
  Candidate: 1:5.6-2
  Version table:
     1:5.6-2 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/universe amd64 Packages
```

ddohler: The version in the Ubuntu repositories doesn't allow passing parameters inside of `check program with path`, which was introduced in 5.7. So for the docker containers that get checked with `docker top <container-name>` we would have had to create a separate wrapper script for checking each docker container with the container name prepopulated.

hectcastro: Ah, I see. Curious about the choice of using `docker top`? Not sure I have a great alternative that wouldn't require application level health checks.

ddohler: Yeah, the goal was to find a program that will return 0 if a container is running, non-zero otherwise, ideally without piping to `grep` and friends, which is why `docker ps` didn't seem to fit the bill. It does seem like this is sort of a side-effect of `docker top` so I'm open to other suggestions.

hectcastro:
👍

PR #529, opened 2016-05-19
Fix cronjob logging, add logging docs

ddohler:
This adds documentation for the location of log files, and also ensures that logging for Celery tasks is written to a specified location on the Celery host.


kshepard:
+1, glad it wasn't too bad! The documentation will be very helpful.

PR #531, opened 2016-05-20
Improve RTL styling in several places

kshepard:
* Proper alignment of username/password labels on login page
 * Proper alignment of json-editor buttons (collapse/etc.)
 * Better dashboard layout
 * Use DRIVER title in navbar

Screenshots:

rtl-dashboard

rtl-login

rtl-json-editor

### DIFF
--- a/web/app/scripts/views/dashboard/dashboard-partial.html
+++ b/web/app/scripts/views/dashboard/dashboard-partial.html
@@ -39,6 +39,16 @@
     </div>
 
     <div ng-if="!ctl.showBlackSpots" class="blackspots-disabled">
+        <div class="block block-2x1 quick-map">
+            <div class="block-inner">
+                <div class="map" leaflet-map recent-events zoom-to-boundary></div>
+                <h3><span>{{ ctl.recordType.plural_label }}: {{ 'DASHBOARD.RECENT_EVENTS_PERIOD' | translate }}</span>
+                    <a ui-sref="map" class="small pull-right">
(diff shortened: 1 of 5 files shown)


web/app/scripts/views/dashboard/dashboard-partial.html
```diff
@@ -39,6 +39,16 @@
     </div>
 
     <div ng-if="!ctl.showBlackSpots" class="blackspots-disabled">
+        <div class="block block-2x1 quick-map">
+            <div class="block-inner">
+                <div class="map" leaflet-map recent-events zoom-to-boundary></div>
+                <h3><span>{{ ctl.recordType.plural_label }}: {{ 'DASHBOARD.RECENT_EVENTS_PERIOD' | translate }}</span>
+                    <a ui-sref="map" class="small pull-right">
+                        {{ 'DASHBOARD.VIEW_ALL' | translate }} {{ ctl.recordType.plural_label | lowercase }} »
```
ddohler: Just double-checking, does this chevron need to be escaped?

ddohler:
+1, I like the new dashboard layout.

web/app/scripts/views/dashboard/dashboard-partial.html
kshepard: I don't think so. It's been like this since the begininning and we haven't seen any problems with browsers we've tested with. Though I wouldn't doubt if an old browser might have an issue with it.

PR #537, opened 2016-05-20
Fix exporting CSV with non-ASCII headings

### LINKED ISSUE #534: CSV Exports don't support non-ASCII field names
Attempting to export a schema which contains non-ASCII field names ~~(and possibly values)~~ returns an encoding error.

ddohler:
Closes #534


kshepard:
+1

PR #538, opened 2016-05-20
Allow reports with non-ASCII row/column headings

### LINKED ISSUE #533: Custom reports don't support non-ASCII field names
Attempting to create a custom report based upon a field with a non-ASCII name fails with a 500 error.

ddohler:
Fixes #533


kshepard:
+1

Issue #543, opened 2016-05-22
SA - Editing Related Content Type Creates Duplicates


tgilcrest:
When trying to make some changes to a Related Content type, we saved the changes and then returned to the View Related Content.  In the list of related content, the type had been duplicated.  However, in the app, only one of the types of Related Content was still visible (for example, it didn't show two "Photos" tabs when viewing an incident).

When we deleted of the duplicate types, *some* of the tabs were also deleted in the app incident details.  From their appearance in the schema editor they are identical, so we couldn't tell which one was actually connected to the app (until you changed / deleted one). 

This also happened for Incident details, which was also duplicated. Because it is required, it can't be deleted.  This has also created duplicate filters in the record list.

ddohler:
This is caused specifically by editing the Plural Title for the Related Content Type; this should be avoided until we've got a fix. Other edits should be fine.

PR #547, opened 2016-05-23
Fix editing plural titles of related content

### LINKED ISSUE #544: SA - Headers on Incident List Disappeared
After changing some fields in incident details, headers on the incident list disappeared.  This may be because the data no longer matches the fields in the schema.

In general, how do changes in the schema editor change affect how the data is stored and retrieved? Does a database conversion need to happen every time the schema editor is changed?

ddohler:
This form was not using the camelCasing code, so it inserted the key in
the wrong place, causing a "duplicate" schema field to be generated.

Closes #544, #543


kshepard:
+1

PR #549, opened 2016-05-24
Add BlackSpotConfig model and API endpoints

ddohler:
The API can only be used to update the config, not create or delete
configs, because the BlackSpotConfig is meant to be treated as a
singleton.

A data migration is included to generate a default BlackSpotConfig,
although the API list endpoint will also generate one if it doesn't
already exist.


kshepard:
+1. I do wonder a bit if it should be more generic, for storing other similar admin-defined settings. But we don't have any others planned at the moment, so maybe it's good to keep it simple.

ddohler:
I thought about genericizing it; I considered combining it with the social cost configuration model that I'm working on, but since they have slightly different behaviors it seemed like it would keep things simpler to leave them separate, yeah.

kshepard:
Cool, sounds good.

PR #550, opened 2016-05-24
Cost Config Model

ddohler:
Creates a RecordCostConfig model which can be used for storing a configuration that allows calculating the social cost of records. This should be ready to merge on its own, but I'm going to hold off until I've started writing some of the aggregation endpoint, which is the other half of this on the backend. If that requires major surgery to this model then I may supersede this PR.

I also noticed that the BlackSpotConfig class didn't inherit from AshlarModel, so I fixed that as well so that it's more consistent with the rest of the API.

This can be tested by navigating to `/api/recordcosts/` and interacting with the API. The browsable API doesn't handle the HStore field correctly unless you use the raw data entry field to put in JSON manually, unfortunately.

### DIFF
--- a/app/black_spots/migrations/0007_auto_20160525_0320.py
+++ b/app/black_spots/migrations/0007_auto_20160525_0320.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import datetime
+from django.utils.timezone import utc
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('black_spots', '0006_auto_20160525_0111'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='blackspotconfig',
+            name='id',
+        ),
+        migrations.AddField(
+            model_name='blackspotconfig',
+            name='created',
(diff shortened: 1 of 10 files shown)


kshepard:
I tested this out, so far so good. I have a couple of thoughts since it's potentially still kind of in progress. I think it'd probably be good to do some validation of the fields being inserted to ensure they actually reference appropriate values on the active schema of the selected record type. Also, I worry about not allowing checkbox fields. It would make it easier from a programming perspective, but in practice, the Philippines users do desire the Severity field to be a checkbox type to be able to specify that there was a fatality, as well as an injury, as well as property damage, etc. So maybe give some thought into if it would be do-able to also allow checkbox fields, and use something like the max or the sum to determine the social cost of each record.

ddohler:
Those are both good ideas; I deliberately left out validation to speed up development but it wouldn't be hard to add.

The reason I initially opted not to allow checkbox fields was because it allows us to mostly piggyback off of the existing aggregation code that we developed for custom reports if we can assume that each field has only one value. However, there's no enforcement in the code at this stage that it can't be a checkbox, so it's possible that it might output useful results even with checkbox fields. My worry is that it may choose a random value from among the selected checkbox values, but I'll experiment once I've got it working to see what that behavior is; it's possible it may work or there may be a way we can tweak it.

kshepard:
:+1: sounds good.

ddohler:
Superseded by #555; I added validation and the ability to handle checkbox values in that PR.

PR #554, opened 2016-05-25
Fix boundary labels on custom report

### LINKED ISSUE #553: SA - Labels for Geography Not Displaying on Custom Report
When using a Geography layer for either rows or columns in the custom report, the labels aren't being displayed.  This happens for both Provinces and Maintenance Contracts, so doesn't seem to be an Arabic issue.  Labels _do_ work when using the Geographic Aggregation option, when non-geographic rows / columns are selected.

image

kshepard:
This also adds sorting for better display of the labels, and
removes items from the list that have no display label.

Closes #553 

screenshot from 2016-05-25 13 00 01

### DIFF
--- a/app/data/views.py
+++ b/app/data/views.py
@@ -561,8 +561,20 @@ class DriverRecordViewSet(RecordViewSet, mixins.GenerateViewsetQuery):
         """
         boundary = Boundary.objects.get(pk=boundary_id)
         polygons = BoundaryPolygon.objects.filter(boundary=boundary)
-        labels = [{'key': str(poly.pk), 'label': poly.data[boundary.display_field]}
-                  for poly in polygons]
+
+        # Sort the polygons by display_field and remove any items that have an empty label
+        polygons = sorted([p for p in polygons if p.data[boundary.display_field]],
+                          key=lambda p: p.data[boundary.display_field])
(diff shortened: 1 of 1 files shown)


app/data/views.py
```diff
@@ -561,8 +561,23 @@ def _make_boundary_case(self, boundary_id):
...
         boundary = Boundary.objects.get(pk=boundary_id)
         polygons = BoundaryPolygon.objects.filter(boundary=boundary)
-        labels = [{'key': str(poly.pk), 'label': poly.data[boundary.display_field]}
-                  for poly in polygons]
+
+        # Sort the polygons by display_field
+        polygons = sorted(polygons, key=lambda p: p.data[boundary.display_field])
+
+        # Remove any items that have an empty label
+        polygons = [p for p in polygons if p.data[boundary.display_field]]
```
ddohler: If you wanted to save a line you could probably move this comprehension up inside the `sorted()` call, but it's fine the way it is, just thought I'd mention it.

ddohler:
+1

app/data/views.py
kshepard: Thanks, that looks a little better, I'll go with that.

PR #555, opened 2016-05-25
Add an endpoint to perform record cost aggregation

ddohler:
This supersedes #550.

Changes:
- Add a `RecordCostConfig` Django model for configuring how the cost of `Records` should be calculated, as well as API endpoints for interacting with RecordCostConfigs. The API endpoints perform validation to ensure that the `RecordCostConfig` is valid at the time it is generated, although it may become outdated if the schema is modified later.
- Make it easier to launch an interactive Django `runserver` session, e.g. if you need to interactively debug views.
- Add a new endpoint, `/api/records/costs` for aggregating the social cost of records.

Testing:
1. Checkout and run migrations
2. Use the API to construct a new RecordCostConfig linked with the active RecordType for your app. The request will need to look something like the following, although it needs to match the active schema, so this may not be exactly accurate: 
```json
{
    "content_type_key": "accidentDetails",
    "property_key": "Severity",
    "enum_costs": {
        "Fatal": "13",
         "Injury": "7",
         "Property": "3"
     },
    "record_type": "75903939-0ab0-4de8-9e21-dc92631975d5"
}
```
3. <-- actually 3. Visit `/api/records/costs/`. As with other endpoints on the `/records/` ViewSet, you can specify filter parameters to reduce the scope of the events that are returned. The `record_type` parameter is required, however, and must match the `record_type` parameter used when creating the RecordCostConfig in step 2. A full URL might be `http://localhost:7000/api/records/costs/?archived=False&details_only=True&jsonb=%7B%22accidentDetails%22:%7B%22Num+driver+casualties%22:%7B%22_rule_type%22:%22intrange%22,%22min%22:null,%22max%22:null%7D%7D%7D&limit=50&occurred_max=2016-10-15T15:59:59.000Z&occurred_min=2015-09-30T16:00:00.000Z&record_type=75903939-0ab0-4de8-9e21-dc92631975d5`. You should received a response that looks something like:
```json
{
    "total": 171,
    "outdated_cost_config": false,
    "subtotals": {
        "Fatal": 0,
        "Injury": 63,
        "Property": 108
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ Both Angular apps can be run in development mode via:
 The frontend app will be available on port 7002 and the schema editor will be available on port
 7001. Both will reload automatically as changes are made.
 
+To make requests to a Django runserver directly (for example, to perform interactive debugging in
+the request-response cycle), run `./scripts/manage.sh runserver 0.0.0.0:8000`. You should then be
+able to access the Django runserver on port 3001 of the `app` VM.
+
 #### Updating existing translation files
 New angular translation tokens should be added to i18n/exclaim.json with a value of "!<english>".
 The English translation (en-us.json) is automatically built from exclaim.json. New tokens are also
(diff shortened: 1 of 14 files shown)


kshepard:
+1

PR #558, opened 2016-05-27
Add black spot severity threshold config UI

kshepard:
A settings section was added to the schema editor app to contain
this setting, as well as any more that we may add in the future.

### DIFF
--- a/schema_editor/app/scripts/views/settings/list-controller.js
+++ b/schema_editor/app/scripts/views/settings/list-controller.js
@@ -1,0 +1,60 @@
+(function() {
+    'use strict';
+
+    /**
+     * @ngInject
+     */
+    function SettingsListController ($log, $scope, BlackSpotConfig, Notifications, Utils) {
+
+        var ctl = this;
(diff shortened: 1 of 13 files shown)


schema_editor/app/scripts/views/settings/list-controller.js
```diff
@@ -0,0 +1,60 @@
...
+
+        function blackSpotSeverityUpdateClicked() {
+            /* jshint camelcase: false */
+            var threshold = ctl.blackSpotConfig.severity_percentile_threshold;
+            /* jshint camelcase: true */
+
+            // Validate the black spot severity threshold value.
+            // The threshold will be undefined if it didn't meet the criteria of the input field.
+            // The range check has been added as well, just in case, but probably isn't needed.
+            if (threshold === undefined || threshold < 0 || threshold > 1) {
```
ddohler: There's server-side validation for this as well but it's good to have it client-side too so that users receive quicker feedback.

ddohler:
+1

PR #559, opened 2016-05-27
Add dashboard social costs widget

ddohler:
Replaces the perennially unimplemented proportions widget with a social costs widget that displays calculated social and economic harm caused by accidents over the past 90 days.

Adds `cost_prefix` and `cost_suffix` fields to the `RecordCostConfig` model; these are primarily designed to display currency information.

Displays a configuration warning if there are any errors / warnings retrieving costs.

Fixes a couple bugs with the `costs` endpoint:
- If no Records at all were found by the query, the cost calculation would throw a 500 error; it now correctly returns 0.
- A 500 would be thrown if a RecordCostConfig was not found; it now correctly returns a 404.

philippinescosts (not real numbers).


kshepard:
About to review this. From the screenshot, I'm wondering if we should add a little more interactivity to the widget to spice it up a little. Maybe something like the 30/60/90 day buttons on the event counts widget.

ddohler:
To spice this up, the easiest thing would probably be to allow the user to toggle between total and subtotal views. That will probably have to happen as part of the map card anyway, so I'll see if I can roll it into that card.

kshepard:
Cool, that sounds good to me.

kshepard:
+1

PR #560, opened 2016-05-27
Optimize black spot calculation script

kshepard:
The script has been optimized to first load records and use the
buffers to filter out all roads that will never be associated with
any records. This speeds up all subsequent operations of the script
and it is able to complete easily with the full Philippines boundary.

Also a few bug fixes were made regarding behavior of UUIDs vs. objects,
and cleanup of temp files.


kshepard:
Here are some black spots in Manila that were generated:

manila-blackspots

ddohler:
+1

PR #563, opened 2016-06-01
Societal costs tool on map

ddohler:
- Adds a societal cost tool to the map that updates as the map filter parameters are updated. It can be clicked to expand and see subtotals.
societalcost1
societalcost2
- Adds the subtotal functionality to the dashboard widget as well; it can be clicked to toggle between viewing totals and subtotals.

Also fixes a recent issue I've had where provisioning will change the permissions of all my local static files, causing them to clutter up `git status`. This was caused by a mismatch between those files' permissions as set by Ansible during provisioning and their permissions in the repository. The mismatch isn't new, as far as I can tell, so I'm not sure why it suddenly started causing problems now, but it seemed like something that should have been fixed, so I went ahead and committed the permissions changes that Ansible makes so that the repository aligns with what the deployment code enforces.

Testing should be as simple as following steps 1 and 2 of #555 although it may be useful to load the dashboard and map once before completing step 2 to see how it reacts to a misconfiguration situation.


web/app/scripts/social-costs/social-costs-directive.js
```diff
@@ -5,11 +5,35 @@
     function SocialCosts() {
         var module = {
             restrict: 'EA',
-            templateUrl: 'scripts/social-costs/social-costs.html',
+            templateUrl: function(elem, attrs) {
+                // Note that use of templateUrls is discouraged in most cases
```
KlaasH: Since they're used in different places and not set dynamically, what about just defining two directives, e.g. `driver-social-costs` and `driver-social-costs-tool`?

ddohler: Yeah, that's a good point -- my initial reasoning was that they have very similar behavior, so doing it this way would avoid the boilerplate that would accompany defining a second directive. Now that I've got it all set up, there are a couple places where the functionality is slightly different, so I think this is probably a good change to make.

KlaasH: This way is fine, too.  It would mean you could get rid of the big comment and the attributes, but at the cost, as you say, of more duplicated directive boilerplate.

KlaasH:
The dashboard version is sitting too high in its containing box (for me, on Linux Chromium):

cost too high

Other than that this looks good, :+1:.

ddohler:
Updated with fixes to tests as well as the header positioning on the dashboard. I also switched to one-way data bindings within the directive to try and follow our Angular best practices more closely.

I decided to forgo a refactor into separate directives in the interest of time, but I put in a note about the possibility of splitting them in case the functionality diverges further in the future.

PR #565, opened 2016-06-07
Make style adjustments to the stepwise chart

Lknechtli:
### Testing
The bar chart appears on the dashboard and on the map under the charts tab

To test, verify that for resolutions >= 1280x720 the chart renders correctly

Verify that the chart looks good in both left to right and right to left modes (this requires having a translation available for Arabic)

Dashboard render @ 1280x720:
image

image

Dashboard render @ maximixed screen (1920x1080):
image

image

Note: slight scaling issue at <1280 px width, which needs to be fixed. Should probably adjust text size at smaller resolutions so everything fits

### DIFF
--- a/web/app/scripts/stepwise/stepwise-directive.js
+++ b/web/app/scripts/stepwise/stepwise-directive.js
@@ -4,61 +4,214 @@
  *  the property 'occurred_from'.
  */
 
-(function () {
+(function() {
     'use strict';
 
     /* ngInject */
-    function Stepwise($translate, LanguageState) {
+    function Stepwise($translate, $window, LanguageState) {
         var module = {
             restrict: 'E',
             scope: {
-              chartData: '=',
-              minDate: '=',
-              maxDate: '='
+                chartData: '<',
+                minDate: '<',
+                maxDate: '<',
+                inDashboard: '<'
             },
-            template: '<svg></svg>',
-            link: function(scope, elem) {
-                var margin = {top: 6, right: 20, bottom: 58, left: 30},
-                    width = 660 - margin.left - margin.right,
-                    height = 230 - margin.top - margin.bottom;
-
-                // GLOBALS
-                var svg, xAxis, yAxis;  // GLOBAL
-                // scales -- set in init
-                var xRange;
-                var tScale;
-                var x;
-                var y;
-
+            link: function(scope, element) {
                 // Need to set after initialization to guarantee translations are available
-                var tooltip = null;
-
                 $translate.onReady(init);
 
                 /**
-                 * Initialize graph to make clean updates possible in a separate, 'updateChart',
-                 *  function. Grab the SVG, manipulate it, generate tooltips, gather rectangle
-                 *  elements together under `rect`.
+                 * Set up listeners for data and draw area size
                  */
                 function init() {
                     var isRightToLeft = LanguageState.getSelected().rtl;
+                    var svg = d3.select(element[0])
+                        .append('svg')
+                        .attr('width', '100%')
+                        .attr('height', '100%');
 
-                    xRange = isRightToLeft ? [width, 0] : [0, width];
-                    tScale = d3.time.scale().range(xRange);
-                    x = d3.scale.ordinal().rangeBands(xRange, 0.05);
-                    y = d3.scale.linear().range([height, 0]);
-
-                    /**
-                     * Watch for changes to chartData and redraw
-                     */
-                    scope.$watch('chartData', function(newVal) {
-                        if (newVal) {
-                            var data = formatData(newVal);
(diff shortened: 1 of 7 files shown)


Lknechtli:
@jbranigan any thoughts on the orientation of the y axis label in Arabic? Not sure whether it should read top down (as it does not) or bottom up

Lknechtli:
Holding off on this - I'l going to put the cost widget in this since my work on it fixes some issues in this pr

Lknechtli:
- [ ] To do - fix RTL layout

Lknechtli:
With cost widget:
image

image

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -4,154 +4,302 @@
...
-    function Stepwise($translate, LanguageState) {
+    function Stepwise($translate, $window, LanguageState) {
+        var inMapMode = false;
         var module = {
             restrict: 'E',
             scope: {
-              chartData: '=',
-              minDate: '=',
-              maxDate: '='
+                chartData: '=',
```
ddohler: Recent change to our preferred style, but it would be good to switch these to one-way bindings if possible.

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -4,154 +4,302 @@
...
-                            d.dt.toLocaleDateString() +
-                            '</br><strong>' +
-                            $translate.instant('RECORD.EVENT_COUNT') +
-                            ':</strong> <span>' +
-                            d.count +
-                            '</span>';
-                    }).offset([-20,-15]);
+                /**
+                 * Render the chart.
+                 * While updating the chart incrimentally may yield better
```
ddohler: Typo: Incrementally

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -4,154 +4,302 @@
...
+                        .attr('x', (isRightToLeft ? 5 : -5))
                         .attr('dy', 0);
+                    yAxisContainer
+                        .append('text')
+                        .attr('text-anchor', 'middle')
+                        .attr('transform',
+                            'translate(' +
+                            (isRightToLeft ? offsets.yAxisText.x : -offsets.yAxisText.x) +
+                            ',' + (dimensions.height / 2 - 2 * offsets.yAxis.y) + ')rotate(-90)')
+                        .text($translate.instant('DASHBOARD.STEPWISE_TITLE'));
```
ddohler: This shouldn't appear on the map since the filter period there isn't guaranteed to be 90 days. Currently it does.

ddohler:
I'm noticing some issues with the axis labels on the map as compared with staging.
This branch:
graph1
Staging:
graph2

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -4,154 +4,302 @@
...
+                    var d3Container = d3.select(element[0])[0][0];
 
-                    var rawSvg = elem.find('svg')[0];
-                    var bufferedWidth = width + margin.left + margin.right;
-                    var bufferedHeight = height + margin.top + margin.bottom;
+                    var dimensions = {
+                        width: d3Container.offsetWidth - 40,
+                        height: d3Container.offsetHeight - 15
+                    };
+                    if (dimensions.width <= 0 || inMapMode) {
```
ddohler: Rather than having this happen implicitly, it would be better to add an explicit configuration setting to the directive. That way it won't behave unexpectedly if this graph is moved to a new location or the display logic is reconfigured on the dashboard or map.

web/app/scripts/stepwise/stepwise-directive.js
```diff
@@ -4,154 +4,302 @@
...
+                            'count': d.count
+                        };
+                    });
+                    var allWeeks = weekRange(formattedData.moment);
+                    var maxCount = _.max(data, function(d) {
+                        return d.count;
+                    }).count || 10;
+
+                    // Set the domain for our X scale - it NEEDS to be at a lower precision than
+                    //  seconds to prevent behavioral abnormalities
```
ddohler: What specifically happens if it's at a resolution of seconds?

ddohler:
Increasing the amount of space between graph bars was part of this card but it doesn't look like this affects the width of the bars as compared with staging. I'm not totally convinced it'll look better that way but we should at least give it a try.

web/app/scripts/stepwise/stepwise-directive.js
Lknechtli: not sure, that comment is from the original code. I can test it out though

ddohler:
+1

PR #568, opened 2016-06-09
Add date translation service and translations

Lknechtli:
Add translations for stepwise graph
Add date translations to list and detail views

Service contains translations for en-us, ar-sa, and exclaim (any unknown language will show en-us dates with an exclaim in front)

### DIFF
--- a/deployment/ansible/roles/driver.web/templates/index.html.j2
+++ b/deployment/ansible/roles/driver.web/templates/index.html.j2
@@ -174,6 +174,7 @@
 
     <script src="scripts/localization/module.js"></script>
     <script src="scripts/localization/local-date-time.js"></script>
+    <script src="scripts/localization/date-utils.js"></script>
 
     <script src="scripts/details/module.js"></script>
     <script src="scripts/details/details-field-controller.js"></script>
--- a/web/app/scripts/details/details-constants-partial.html
+++ b/web/app/scripts/details/details-constants-partial.html
@@ -3,9 +3,9 @@
         {{ 'RECORD.OCCURRED' | translate }}
     </label>
     <div class="value constant date occurred">
-        {{ ::ctl.record.occurred_from | localDateTime : ctl.dateFormat }}
+        {{ ::ctl.record.occurred_from | localizeRecordDate: 'long':'time'}}
         <span ng-if="::ctl.record.isSecondary && ctl.record.occurred_to">
-         - {{ ::ctl.record.occurred_to | localDateTime : ctl.dateFormat }}
+         - {{ ::ctl.record.occurred_to | localizeRecordDate: 'long':'time'}}
         </span>
     </div>
 </div>
@@ -14,7 +14,7 @@
         {{ 'RECORD.CREATED' | translate }}
     </label>
     <div class="value constant date created">
-        {{ ::ctl.record.created | localDateTime: ctl.dateFormat }}
+        {{ ::ctl.record.created | localizeRecordDate: 'long':'time'}}
     </div>
 </div>
 <div class="col-sm-12">
--- a/web/app/scripts/localization/date-utils.js
+++ b/web/app/scripts/localization/date-utils.js
@@ -1,0 +1,106 @@
+(function() {
+    'use strict';
+
+    /* ngInject */
+    function DateUtils(LanguageState, WebConfig) {
+        /* jshint quotmark: false */
+        var languageMap = {
+            'ar-sa': {
+                language: 'ar',
+                calendar: 'ummalqura',
+                formats: {
+                    'short': 'M Y',
+                    'longNoTime': 'd MM, Y',
+                    'long': 'd MM, Y',
+                    'numeric': 'dd/mm/yyyy'
+                }
+            },
+            'en-us': {
+                language: 'en',
+                calendar: 'gregorian',
+                formats: {
+                    'short': 'M Y',
+                    'long': 'MM d, Y',
+                    'numeric': 'mm/dd/yyyy'
+                }
+            },
+            'exclaim': {
+                language: 'en',
+                calendar: 'gregorian',
+                formats: {
+                    'short': "'!'M Y",
+                    'longNoTime': "'!'MM d, Y",
+                    'long': "'!'MM d,Y",
+                    'numeric': "'!'mm/dd/yyyy"
+                }
+            }
+        };
+        /* jshint quotmark: single */
+
+        var module = {
+            getLocalizedDateString: getLocalizedDateString
+        };
+        return module;
+
+        /*
+         * Get a localized date string for a date, using the LanguageState to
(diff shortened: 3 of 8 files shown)


Lknechtli:
image

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+        };
+
+        var module = {
+            getLocalizedDateString: getLocalizedDateString
+        };
+        return module;
+
+        function getLocalizedDateString(date, format, includeTime) {
```
ddohler: This could use a docstring since I expect it is going to be pretty heavily utilized as we add further localization.

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,87 @@
...
+            var language = LanguageState.getSelected().id;
+            if (!language) {
+                language = 'exclaim';
+            }
+            var isRtl = LanguageState.getSelected().rtl;
+            var conversion = languageMap[language];
+            var parsedDate = $.calendars.instance()
+                .parseDate('mm/dd/yyyyy', date.toLocaleDateString()).toJSDate();
```
ddohler: I'm not sure this is going to work in all cases; the docs for `toLocaleDateString` specify that it returns a "language-sensitive" date representation, so if a user's browser is in a language other than `en-us` I would expect this to return a format other than `mm/dd/yyyy`, which would cause parsing to fail.

In any event, I think this line may be superfluous; since Javascript Date objects are already in the Gregorian calendar, `date` can be used directly in line 50:
`var convertedDate = $.calendars.instance(conversion.calendar, conversion.language).fromJSDate(date);`

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,87 @@
...
+            var convertedDate = $.calendars.instance(conversion.calendar, conversion.language)
+                .fromJSDate(parsedDate);
+            var datestring = convertedDate.formatDate(conversion.formats[format], convertedDate);
+            if(includeTime) {
+                var minutes = date.getMinutes() + '';
```
ddohler: Could this whole block be replaced with a call to something like `datestring = datestring + date.toLocaleTimeString(language)`? (I don't totally expect this to work the first time but it should be close; see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString)

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,87 @@
...
+                var hours = date.getHours() + '';
+                if (hours.length < 2) {
+                    hours = '0' + hours;
+                }
+                var seconds = date.getSeconds() + '';
+                if (seconds.length < 2) {
+                    seconds = '0' + seconds;
+                }
+                if (isRtl && format !== 'long') {
+                    datestring = hours + ':' + minutes + ':' + seconds + '  ' + datestring;
```
ddohler: Huh, I'm surprised this is necessary, I would have thought that switching to RTL-mode would take care of the ordering, i.e. that "<Date> <Time>" would become "<Time> <Date>" when displayed RTL, but it doesn't?

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+            return datestring;
+        }
+    }
+
+    /* ngInject */
+    function LocalizeDateFilter(DateUtils) {
+        function module(d, format, includeTime) {
+            return DateUtils.getLocalizedDateString(new Date(d), format, includeTime === 'time');
```
ddohler: We've had problems with the `Date()` constructor on HunchLab failing on older browsers. It might be a good idea to use a shim: https://github.com/csnover/js-iso8601/tree/lax

web/app/scripts/utils/date-utils.js
Lknechtli: Concluded that while it's more concise, it isn't accurate (12hr clock vs 24hr, roman vs arabic numerals)

web/app/scripts/utils/date-utils.js
Lknechtli: we need the time on the right, even in RTL languages

web/app/scripts/utils/date-utils.js
Lknechtli: done

ddohler: I think the line above can be deleted now.

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,152 @@
...
+                    'longNoTime': 'MM d, Y',
+                    'long': 'MM d, Y - ',
+                    'numeric': 'mm/dd/yyyy - '
+                }
+            },
+            'exclaim': {
+                language: 'en',
+                calendar: 'gregorian',
+                formats: {
+                    'short': '\'!\'M Y',
```
ddohler: Using double-quotes here could make this string easier to read.

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,152 @@
...
+                timestamp = origParse ? origParse(date) : NaN;
+            }
+
+            return timestamp;
+        };
+    }(Date));
+
+    /* ngInject */
+    function DateUtils(LanguageState) {
+        var languageMap = {
```
ddohler: As a thought, would it make sense to move this into `config.js` inside the objects that hold the rest of the language data? That way all the language-specific data would be packaged up together and it would be easy to add new date formats by just editing that config.

ddohler:
+1 after a bit more cleanup; consider moving the date format definitions into https://github.com/WorldBank-Transport/DRIVER/blob/develop/deployment/ansible/group_vars/all.example#L66

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,154 @@
+(function() {
```
ddohler: Now that I've done some work on date localization myself, I think that this file would be more appropriate in `scripts/localization` where there is already some Date-related localization code.

web/app/scripts/utils/date-utils.js
```diff
@@ -0,0 +1,154 @@
...
+    function DateUtils(LanguageState) {
+        /* jshint quotmark: false */
+        var languageMap = {
+            'ar-sa': {
+                language: 'ar',
+                calendar: 'ummalqura',
+                formats: {
+                    'short': 'M Y',
+                    'longNoTime': 'd MM, Y',
+                    'long': ' - d MM, Y',
```
ddohler: Looking at this after having worked on the calendar, I think that these formats should include only the date formatting, and leave it up to whatever code uses this format to decide the placement of hyphens to separate the date from the time, if any.

ddohler:
+1

PR #571, opened 2016-06-13
Enable black spot calculation cron job

kshepard:
Note: I tested this successfully on my dev machine, but it is very unlikely to complete successfully in staging unless we upgrade the celery worker VM to a machine type that has more memory.

### DIFF
--- a/deployment/ansible/roles/driver.celery/tasks/main.yml
+++ b/deployment/ansible/roles/driver.celery/tasks/main.yml
@@ -37,17 +37,17 @@
 - name: Create a directory for scheduled task logs
   file: path={{ driver_cron_log_dir }} state=directory
 
-# TODO: uncomment and modify this when the final script is ready
-# - name: Add black spot calculation cronjob
-#   cron: name="black spot calculation"
-#         job="/usr/local/bin/heimdall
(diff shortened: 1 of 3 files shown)


deployment/ansible/roles/driver.celery/tasks/main.yml
```diff
@@ -37,17 +37,17 @@
...
-#         >> {{ driver_cron_log_dir }}/calculate_black_spots.log 2>&1"
-#         hour=19 minute=10  # 3:10am in UTC+8
+- name: Add black spot calculation cronjob
+  when: web_js_blackspots_visible
+  cron: name="black spot calculation"
+        job="/usr/local/bin/heimdall
+        --database '{{ heimdall_db_connection }}'
+        --namespace records
+        --name calculate_black_spots
+        --timeout 300
```
ddohler: This timeout should be increase since the blackspots job can take more than 5 minutes to complete.

ddohler:
+1

deployment/ansible/roles/driver.celery/tasks/main.yml
kshepard: Good call.

PR #576, opened 2016-06-15
Localized date range filter on map

ddohler:
This creates a localized datepicker for the map date range filter that updates based on the interface language. Dates are stored in Gregorian internally so that filtering and saved filters still work.

There is a known issue with click events; clicking on the calendar will cause everything to disappear; this is the result of a fix for a related issue, which would cause the parent dropdown to disappear if the map was clicked, leaving the calendar floating above the map by itself. There are also some styling issues related to the length of weekday names when the calendar is in Arabic.
calendar1
calendar2


web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -2,54 +2,115 @@
...
+                    calendar = $.calendars.instance(dateConfig.calendar, dateConfig.language);
+                    $.calendarsPicker.setDefaults($.calendarsPicker.regionalOptions['']);
+                    scope.calendarOptions = angular.extend({
+                        calendar: calendar,
+                        dateFormat: dateConfig.formats.numeric,
+                        showAnim: '',
+                    }, $.calendarsPicker.regionalOptions[dateConfig.language]);
 
+                    // Make sure that the datepicker hides when the parent dropdown does
+                    elem.on('hide.bs.dropdown', function() {
```
ddohler: @CloudNiner This is related to the visibility issue we've been discussing; without this event listener, the parent drop down with the text fields sometimes disappears without notifying the child calendar element, leaving the calendar floating above the map.

However, the parent dropdown hides itself when the calendar is clicked, so with this fix in place, interacting with the calendar causes both the parent dropdown and the calendar itself to disappear.

The overall issue is that the parent dropdown shouldn't disappear when the user clicks on the child calendar element, but does.

CloudNiner: Seems like you just need to call `stopPropagation` on the datepicker element or its wrapper (if there is one). I'm not familiar with this API, but that should prevent the click bubbling up to the dropdown. The best way to do that would be to add a separate click handler here that just calls stopPropagation.

This fix seems ok as is, just a bit brittle because it depends on a particular dom ID. Could do a `$document.find('.datepicker-class')` using a class you add to all datepicker elements, rather than calling the same method on a bunch of id elements here.

kshepard:
Curious: was there a reason for changing the modes of those files?

web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -2,54 +2,115 @@
...
                     if (filter.label === filterLabel) {
-                        scope.dtMin = filter.value.min;
-                        scope.dtMax = filter.value.max;
+                        // The restored date will be an ISO-8601 string, so we need to convert
+                        // that to a Javascript Date, and then convert that to a localized CDate,
+                        // and then store the formatted string.
+                        dtRange.min = filter.value.min;
+                        dtRange.max = filter.value.max;
+                        if (dtRange.min) {
+                            var jsMin = moment(dtRange.min, moment.ISO_8601).toDate();
```
kshepard: Weird, I'd expect jshint to complain about the underscore.

web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -2,54 +2,115 @@
...
+                 * Sets up the calendar picker based on the current language.
+                 */
+                function configureDatePicker() {
+                    var dateConfig = DateLocalization.currentDateFormats();
+                    calendar = $.calendars.instance(dateConfig.calendar, dateConfig.language);
+                    $.calendarsPicker.setDefaults($.calendarsPicker.regionalOptions['']);
+                    scope.calendarOptions = angular.extend({
+                        calendar: calendar,
+                        dateFormat: dateConfig.formats.numeric,
+                        showAnim: '',
```
kshepard: Trailing comma.

web/app/scripts/filterbar/date-range-directive.js
ddohler: Will the `stopPropagation` strategy work even if the element doesn't exist at the time the call is made? The jQuery plugin inserts the datepicker at the end of the `body` element when its associated `input` tag gains focus, but there's no wrapper element that's available from the point of view of the directive during page load, so I'm worried that there'd be nothing to attach the handler to at that point. But maybe Angular's wrapping code takes care of that?

I like the second idea -- the DOM id reference is what was used here previously, and it seems appropriate when specifically dealing with the min or max, but doing a single call on a class is definitely cleaner for this case.

web/app/scripts/views/map/layers-controller.js
```diff
@@ -328,7 +328,7 @@
 
             if (polygon) {
               return BlackspotSets.query({
-                'efffective_at': FilterState.getDateFilter().maxDate,
+                'effective_at': FilterState.getDateFilter().maxDate,
```
kshepard: Nice catch.

kshepard:
+1, This looks good to me. We should probably hold off on merging it until the click event bug is solved.

web/app/scripts/filterbar/date-range-directive.js
CloudNiner: No I don't think so because event handlers are bound to dom elements. The datepicker should throw a 'created' event or have some sort of callback available to it, you'll need to add the stoppropagation to an click handler created when the element is initialized. jQuery _should_ handle automatically removing the click handler when the datepicker element is destroyed, but double check.

web/app/scripts/filterbar/date-range-directive.js
ddohler: Yeah, me too. It also didn't complain about that trailing comma, but it complained about some other stuff so it may not be configured quite correctly.

web/app/scripts/filterbar/date-range-directive.js
ddohler: Okay, progress: there's a poorly-documented `onShow` event that gets sent and it looks like that includes the proper `<div>` so I think we should be in business.

ddohler: Sweet, that was easy once I tracked down the right event to listen for. Thanks!

ddohler:
The file modes are because they get changed to a certain mode when provisioning, but that mode doesn't match the default system mode, so they show up as changed in git. It started happening fairly recently, so I'm sure there's some underlying cause, but I haven't had the time to figure out exactly why. Matching the repo permissions to their state after provisioning seemed harmless enough, so I've been doing that but if it keeps happening we may want to look into it.

I've got the click events issue mostly sorted thanks to help from @CloudNiner; the click events for some of the controls still propagate for some reason, but I'm hopeful I can get those worked out pretty quickly tomorrow.

web/app/scripts/filterbar/date-range-directive.js
CloudNiner: A possible explanation could be that it ignores underscores in vars that don't use lowercase letters, to allow for a separator in all caps var names. Or I could be throwing darts in the dark.

web/app/scripts/filterbar/date-range-directive.js
CloudNiner: Great, glad that worked out. 

You might want to watch out, though. If the div isn't destroyed and recreated with each 'show', you'll end up with multiple handlers bound. It's fine for now if you're just calling stopPropagation, but it would expose a bug if you added any other logic in that handler. Probably worth taking a second to double check.

ddohler: Yup, I checked and we're good, there's only ever one click handler, even after opening and closing the calendar a bunch of times.

PR #578, opened 2016-06-15
Feature/gradle unicode builds

flibbertigibbet:
Modifications to the Android model jar-building process to support Unicode.


ddohler:
+1

PR #583, opened 2016-06-17
Add function to add property order before saving

### LINKED ISSUE #573: propertyOrder missing for schema sections
Only the primary details section has `propertyOrder` defined.

notthatbreezy:
Closes #573

Adds property order to related schema definitions before saving to any
related definitions that do not have one.

Also adds property order to the development schema.

## Testing
Add a new related schema to incidents in the editor. After saving, look at the schema returned from the API and the related schema definitions should now all have `propertyOrder` defined.

### DIFF
--- a/scripts/incident_schema_v3.json
+++ b/scripts/incident_schema_v3.json
@@ -12,6 +12,7 @@
                 "Picture"
             ],
             "plural_title": "Photos",
+            "propertyOrder": 3,
             "definitions": {},
             "type": "object",
             "properties": {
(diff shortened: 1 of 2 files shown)


scripts/incident_schema_v3.json
```diff
@@ -12,6 +12,7 @@
                 "Picture"
             ],
             "plural_title": "Photos",
+            "propertyOrder": 1,
```
kshepard: Since `propertyOrder` will be used to determine order for showing these sections, here's the desired order in which people should be entering information:
 * Incident Details [0]
 * Vehicles [1]
 * People [2]
 * Photos [3]

PR #584, opened 2016-06-17
Add language param to custom report

Lknechtli:
Add language param to crosstab api endpoint
language param is read, but does nothing for now


app/data/views.py
```diff
@@ -334,6 +334,7 @@ def crosstabs(self, request):
             As you might expect, these operate identically to the row_* parameters above, but the
             results are used as columns instead.
         - record_type: The UUID of the record type which should be aggregated
+        - calendar: the calendar to use for the report to (Ex: 'gregorian' or 'hijri')
```
ddohler: The second example should be `ummalqura`; we don't use `hijri` as a calendar specifier anywhere in the app currently.

ddohler:
+1 after squashing commits.

PR #585, opened 2016-06-17
Only convert key to camelCase

### LINKED ISSUE #570: URL encode schema ref but not referent
JSON schema `ref` fields are URIs and so are encoded. The referent field name under the `definitions` section and elsewhere (such as watch targets) should not be encoded, but are.

notthatbreezy:
Closes #570

This commit switches the behavior of the `generateFieldName` function
slightly by only removing whitespace and converting to camelCase (except
for unicode). When the key/name is used in a URI path it is still
encoded properly.

## Testing

Add a new record type or related record type where the title is unicode. After generation, the field property name should not be encoded, but its URI component should be.

See screenshot for an example with `доллар`

encoding

### DIFF
--- a/schema_editor/app/scripts/schemas/schemas-service.js
+++ b/schema_editor/app/scripts/schemas/schemas-service.js
@@ -131,21 +131,20 @@
         return module;
 
         /**
-         * Convert field titles into field names that are intially lower case,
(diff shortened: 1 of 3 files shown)


kshepard:
+1

PR #586, opened 2016-06-17
Use date-utils.js instead of local-datetime.js

jisantuc:
All references to `localDateTimeFilter` have been replaced with
`localizeRecordDateFilter` in js files, `localDateTime` has been
replaced with `localizeRecordDate`, and `local-date-time.js` and
associated tests have been removed and removed from `<script>`
imports because no longer needed.


kshepard:
Does anything more need to happen with this or is it good to review and merge?

Lknechtli:
lgtm

PR #587, opened 2016-06-20
Add endpoint for enforcer assignments

kshepard:
ToDDoW scaling and tests will be added in a future PR.

### DIFF
--- a/app/black_spots/filters.py
+++ b/app/black_spots/filters.py
@@ -5,12 +5,31 @@ from dateutil.parser import parse
 from django.db.models import Q
 
 from ashlar.exceptions import QueryParameterException
+from ashlar.models import BoundaryPolygon
 
 from black_spots.models import (BlackSpot, BlackSpotSet)
 
 from django.contrib.gis.geos import GEOSGeometry
 
-from rest_framework.exceptions import ParseError
+from rest_framework.exceptions import ParseError, NotFound
+from rest_framework_gis.filterset import GeoFilterSet
+
+
+def parse_and_validate_dt(dt_str, param_name):
+    """
+    Helper for parsing a datetime string and ensuring it's time-zone aware
+    :param dt_str:  datetime string
+    :param param_name:  name of the parameter -- used for creating useful exception messages
+    """
+    try:
+        dt = parse(dt_str)
+    except:
+        raise QueryParameterException(param_name, 'ISO 8601 formatted.')
+
+    if not dt.tzinfo:
+        raise QueryParameterException(param_name, 'timezone aware.')
+
+    return dt
 
 
 class BlackSpotFilter(django_filters.FilterSet):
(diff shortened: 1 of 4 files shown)


app/black_spots/views.py
```diff
@@ -74,3 +76,136 @@ def get_queryset(self):
...
+    instead based direcly on black spots, so this is a list-only ViewSet. This was created
+    as its own ViewSet to make it easier if we ever decide to create a db model for
+    enforcer assignments.
+    """
+    queryset = BlackSpot.objects.all()
+    serializer_class = EnforcerAssignmentSerializer
+    filter_class = EnforcerAssignmentFilter
+    permission_classes = (IsAdminOrReadOnly, )
+
+    def get_required_param(self, key, request):
```
ddohler: We've run into this before; it's really awkward to use different serializers for input and output. I _think_ that it should be possible to do something like this: http://stackoverflow.com/a/17552724 , but I seem to recall that I've tried that and it didn't work for me. But this case might be different than whatever case I was working on, so the approach might work here.

app/black_spots/views.py
```diff
@@ -74,3 +76,136 @@ def get_queryset(self):
...
+        :param assignments: filtered queryset of assignments (black spots)
+        :param shift_start: start dt of the shift
+        :param shift_end: end dt of the shift
+        """
+
+        # TODO: use time of day/day of week to generate the appropriate scaling factor
+        scaling_factor = 1 / (7 * 24.0)
+
+        # Scale the severity score by the scaling factor
+        for assignment in assignments:
```
ddohler: This could perhaps be shortened slightly by replacing it with a call to `map` but it's not critical.

app/black_spots/filters.py
```diff
@@ -56,3 +70,49 @@ def filter_effective_at(self, queryset, effective_at_str):
...
+        except ValueError as e:
+            raise ParseError(e)
+        except BoundaryPolygon.DoesNotExist as e:
+            raise NotFound(e)
+
+    def filter_record_type(self, queryset, rt_id):
+        """Method filter for record type"""
+        return queryset.filter(black_spot_set__record_type=rt_id)
+
+    def filter_shift_end(self, queryset, shift_end_str):
```
ddohler: Is this missing a `filter_shift_start` method?

ddohler:
+1

app/black_spots/filters.py
kshepard: `shift_start` isn't being used for filtering. `shift_end` determines which set of black spots to use, and then the combination of the two will be used for applying the ToDDoW logic. So I don't think it's needed in here, but correct me if I'm wrong.

app/black_spots/views.py
kshepard: Thanks, that worked nicely.

PR #589, opened 2016-06-21
Scale enforcer assignments by ToDDoW distribution

kshepard:


app/black_spots/views.py
```diff
@@ -78,14 +81,12 @@ def get_queryset(self):
 
 
 class EnforcerAssignmentViewSet(drf_mixins.ListModelMixin, viewsets.GenericViewSet):
-    """
-    ViewSet for enforcer assignments
+    """ViewSet for enforcer assignments"""
+    # Enforcer assignments do not currently have an associated model in the database, and are
```
ddohler: Out of curiosity, what was the reasoning behind switching from a docstring to inline comments?

ddohler:
+1

app/black_spots/views.py
kshepard: Good question, glad you asked. DRF adds any ViewSet docstrings to the endpoint description in the browseable API. Since this comment is specific to the code implementation, it doesn't really belong in API docs.

PR #590, opened 2016-06-21
Fix color alignment in incident record detail

jisantuc:
Right-to-left successfully flips all the text on the page,
but SVG coordinates are still in right-to-left, so vehicle color
rectangles wound up left-aligned in their enclosing SVGs, which was
way out of place.

This fix conditions a shift to the rectangles' x attributes that
conditions on right-to-left-ness and the size of the rectangle to
appropriately right-align the SVG elements.


kshepard:
+1

PR #592, opened 2016-06-21
Use localized calendar for add/edit report

Lknechtli:
Create directive for localized calendar

note: didn't include directive in the right file. fixing after I review another PR

### DIFF
--- a/web/app/scripts/localization/date-picker-directive.js
+++ b/web/app/scripts/localization/date-picker-directive.js
@@ -1,0 +1,75 @@
+(function() {
+    'use strict';
+
+    /* ngInject */
+    /*
+     Scope args:
+         datetime {Date} - normal javascript datetime object
+         placeholder {string} - placeholder text for the text field
+         on-change {function} - function to call when datetime is changed
+     */
+    function DatePicker(DateLocalization) {
+        var internalCalendar;
+        var dateConfig;
+        var module = {
+            restrict: 'E',
+            templateUrl: 'scripts/localization/date-picker-partial.html',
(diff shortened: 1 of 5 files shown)


web/app/scripts/localization/date-picker-directive.js
```diff
@@ -0,0 +1,95 @@
...
+                    return;
+                }
+                $($element).calendarsPicker(
+                    'setDate', ctl.internalCalendar.fromJSDate(date)
+                );
+            });
+        }
+
+        function updateDate(date) {
+            ctl.internalDateTime = ctl.internalCalendar.formatDate(ctl.calendarOptions.dateFormat, date);
```
ddohler: I'm not sure I understand how this field is being used; it looks like it gets updated by `ng-model` and `updateDate` but I don't see where the data ever gets read back out or used to calculate anything.

web/app/scripts/localization/date-picker-directive.js
```diff
@@ -0,0 +1,94 @@
...
+        var ctl = this;
+
+        ctl.$onInit = initialize;
+
+        function initialize() {
+            // functions
+            ctl.updateDate = updateDate;
+
+            // data
+            ctl.dateConfig = DateLocalization.currentDateFormats();
```
ddohler: I'm feeling like a good chunk of this logic would be more appropriate in the `link` function of the directive, since a lot of it is the directive talking to itself rather than communicating with the outside world.

ddohler:
+1 after squashing the update commits.

PR #593, opened 2016-06-21
Add basic firewall rules

kshepard:
This commit restricts inbound connections to only the ports
(and where applicable, IPs) that are needed. Here are some details:
 * App -- http[s] is allowed, as are development ports when in develop mode
 * Celery -- http[s] is allowed only from App servers (e.g. for CSV downloads).
 * DB -- PostgreSQL and redis connections are allowed from App and Celery servers.
 * SSH is allowed
 * Everything else is blocked

There is room for improvement if we want to also restrict outbound access,
but this should be a good start.

### DIFF
--- a/deployment/ansible/group_vars/all.example
+++ b/deployment/ansible/group_vars/all.example
@@ -90,3 +90,9 @@ monit_allow_hosts: []
 monit_allow_user: 'test'
 monit_allow_password: 'test'
 monit_enable_web_access: yes
+
+## IP addresses used to configure the firewall
+ip_addresses_app:
+    - 192.168.12.102
+ip_addresses_celery:
+    - 192.168.12.103
--- a/deployment/ansible/roles/driver.app/tasks/firewall.yml
+++ b/deployment/ansible/roles/driver.app/tasks/firewall.yml
@@ -1,0 +1,39 @@
+---
+- name: Enable firewall
+  ufw: state=enabled policy=reject
+
+- name: Allow ssh access
+  ufw: rule=allow port=22
+
+- name: Allow monit access
+  ufw: rule=allow port=2812
+
+- name: Allow http access
+  ufw: rule=allow port=80
+
+- name: Allow https access
+  ufw: rule=allow port=443
+
+- name: Allow Runserver access
+  ufw: rule=allow port=4000
+  when: developing
+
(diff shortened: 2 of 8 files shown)


deployment/ansible/group_vars/all.example
```diff
@@ -90,3 +90,9 @@ monit_allow_hosts: []
 monit_allow_user: 'test'
 monit_allow_password: 'test'
 monit_enable_web_access: yes
+
+## IP addresses used to configure the firewall
+ip_addresses_app:
```
ddohler: Shouldn't it be possible to determine these values from the inventory for a given run?

deployment/ansible/roles/driver.app/tasks/firewall.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Enable firewall
+  ufw: state=enabled policy=reject
+
+- name: Allow ssh access
```
ddohler: I was going to say this could be written more succinctly using `with_items` but then I realized that we'd lose the output of each port that gets opened, which I kind of like. But you might have different preferences, I think either way would be fine.

deployment/ansible/group_vars/all.example
kshepard: That'd be great -- I'll look into it.

ddohler:
+1, nice. Giant Ansible win making this so easy.

deployment/ansible/group_vars/all.example
kshepard: Oh yeah, apparently it's really easy.

deployment/ansible/roles/driver.app/tasks/firewall.yml
kshepard: Yeah, I considered that as well. It's a bit of a toss up. I think I'd prefer to keep them separated out so they're more explicit (plus the nice output), and individual items can potentially be modified more easily if needed. But it definitely does add some repetition.

deployment/ansible/group_vars/all.example
kshepard: Turns out to not be so easy. I tried several options, none of which worked in this context. I'm going to pass on it for now and merge this to test in staging.

PR #596, opened 2016-06-22
Enforcer assignments modal

ddohler:
Adds a UI to allow users to generate enforcement assignments:
enforcer-tool
enforcer popup

Testing from a `./scripts/grunt.sh web serve` should be straightforward on the latest `develop`.

### DIFF
--- a/web/app/scripts/enforcers/enforcers-modal-controller.js
+++ b/web/app/scripts/enforcers/enforcers-modal-controller.js
@@ -1,0 +1,94 @@
+(function () {
+    'use strict';
+
+    /* ngInject */
+    function EnforcersModalController($log, $modalInstance, $state, $translate, $window,
+                                      $timeout, FilterState, BoundaryState, GeographyState,
+                                      RecordState, MapState) {
+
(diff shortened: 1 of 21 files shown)


web/app/scripts/enforcers/enforcers-modal-controller.js
```diff
@@ -0,0 +1,94 @@
...
+
+            ctl.polygon = MapState.getFilterGeoJSON();
+        }
+
+        function closeModal() {
+            $modalInstance.close();
+        }
+
+        function onParamChanged() {
+            // TODO: Without this, changes to the datepicker take two digest cycles to appear on
```
kshepard: Do you think this means other areas in the app with datepickers will also have this same issue?

ddohler: I don't think so because the other spot with a datepicker doesn't use the custom directive I made here, but it'd be a nice refactor if it did, and if we made that change then it would presumably have the same problem.

kshepard:
+1. I think we'll need a designer task to help out with the datetime picker styling to make it more compact.

Issue #577, opened 2016-06-15
Numeric fields on schema are given type text


flibbertigibbet:
See WorldBank-Transport/DRIVER-Android#48.

The `type` of numeric fields should be `integer` or `number`. As it is, number fields are noted using the non-standard `format` `number`.

kshepard:
This was fixed in #598

PR #598, opened 2016-06-23
Add a number type to schema-builder

### LINKED ISSUE #562: SA - Android App Not Respecting Schema Restrictions
It seems like data type restrictions are not being respected by the android app.  For example, Number of Deaths is defined as a number in the schema editor.  However, I could enter 'gh' for that value and it would upload to the database.  When I checked that record in the record list, it had 'gh' as the value for that field.

The example record's Date & Time: 6/1/16 11:48 AM

jisantuc:
Validation on the Android app depends on a better number type
than text formatted as a number. This commit adds that number
type and and renderers for displaying that information in
detail views.

Fixes #562.

### DIFF
--- a/schema_editor/app/builder-schemas/related.json
+++ b/schema_editor/app/builder-schemas/related.json
@@ -9,7 +9,8 @@
             { "$ref": "#/definitions/textField" },
             { "$ref": "#/definitions/selectList" },
             { "$ref": "#/definitions/imageUploader" },
-            { "$ref": "#/definitions/localReference" }
+            { "$ref": "#/definitions/localReference" },
+            { "$ref": "#/definitions/numberField" }
         ]
     },
     "definitions": {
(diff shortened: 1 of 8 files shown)


flibbertigibbet:
Should fix #577.

This should also fix subtle issues with min/max value checks in the web app, if such restrictions are present.

schema_editor/app/builder-schemas/related.json
```diff
@@ -89,6 +90,36 @@
...
+                "maximum": {
+                    "title": "Maximum Value",
+                    "type": "number"
+                },
+                "fieldType": {
+                    "options": {
+                        "hidden": false
+                    },
+                    "type": "string",
+                    "enum": ["number"]
```
flibbertigibbet: Also [integer](http://json-schema.org/latest/json-schema-core.html)?

jisantuc: Sure, that makes sense.

kshepard:
Also worth updating `incident_schema_v3.json` to use number types for "Num passenger casualties" and "Num driver casualties".

jisantuc:
@kshepard is `incident_schema_v3.json` just used in loading the data?

kshepard:
Yep, it's just used in development to get the schema and data set up.

kshepard:
This looks very good. I did notice a bug while testing: the `integer` type doesn't seem to work correctly. I added one via the schema editor, but it doesn't save due to a javascript error.

jisantuc:
@kshepard @flibbertigibbet there's one piece of odd behavior left -- json-editor doesn't invalidate floats in string fields, it just takes out the periods and parses what remains as a number.

So...
image

becomes

image

I'd expect it to warn instead of strip periods and parse, but so it goes.

kshepard:
Good to know. I think that behavior is maybe not completely ideal, but it still does the job of not allowing invalid input and providing a warning, so that's good. +1 - going to merge this.

kshepard:
@jisantuc - there seems to be a problem with rendering number fields on the record detail page. The ones defined as 'number' don't seem to show up, and the ones defined as 'integer' give an error: "No renderer defined for field type:".

jisantuc:
So uh remember when you told me a bit ago that `web/app/index.html` is .gitignored? ...fixing

PR #602, opened 2016-06-24
Send `null` instead of "null" for null polygons

### LINKED ISSUE #601: Enforcer assignments UI: don't send "null" polygon_id to the server
If there is no polygon selected, it's currently sending "null" as the polygon_id, which is causing an error.

jisantuc:
Fixes #601


kshepard:
+1

PR #603, opened 2016-06-24
Configure SSL in staging and production 

kshepard:
Also switches to using https for external resources, such as basemaps and fonts.

Note: there is a timing issue with nginx configurations that
requires some manual intervention when initially installing
the certificate. I will create an issue to fix this.


flibbertigibbet:
+1

yay letsencrypt!

PR #605, opened 2016-06-24
Render date range on map page load

### LINKED ISSUE #581: Invalid Gregorian Date on map initial page load
This is probably due to a null or undefined value being passed into the calendar `setDate` function.

jisantuc:
Fixes #581


kshepard:
+1

PR #606, opened 2016-06-27
Ensure that fieldnames are valid java identifiers

### LINKED ISSUE #575: Do Not Allow Control Characters in Schema Field Names
Do not allow forward slashes, or anything that's not a valid Java identifier.

notthatbreezy:
Closes #575 

This commit changes the field name generation to strip any control
characters and prevent the use of reserved words in generated field
names.

Rather than use a complicated regular expression that would be difficult
to match unicode characters with, field name generation now strips out
non-ascii characters, prepends 'driver' to the field name, then appends
an '_<random string>' to generated identifiers.

valid-identifiers

### DIFF
--- a/scripts/incident_schema_v3.json
+++ b/scripts/incident_schema_v3.json
@@ -3,7 +3,7 @@
     "title": "",
     "plural_title": "",
     "definitions": {
-        "photo": {
+        "driverPhoto": {
             "multiple": true,
             "description": "Pictures of the incident",
             "title": "Photo",
@@ -41,7 +41,7 @@
                 }
             }
         },
-        "vehicle": {
+        "driverVehicle": {
             "multiple": true,
             "description": "A vehicle involved in the incident",
             "title": "Vehicle",
@@ -202,7 +202,7 @@
                 }
             }
         },
-        "incidentDetails": {
+        "driverIncidentDetails": {
             "multiple": false,
             "description": "Details for Incident",
             "title": "Incident Details",
@@ -356,7 +356,7 @@
                 }
             }
         },
-        "person": {
+        "driverPerson": {
             "multiple": true,
             "description": "A person involved in the incident",
             "title": "Person",
@@ -508,9 +508,9 @@
     "$schema": "http://json-schema.org/draft-04/schema#",
     "type": "object",
     "properties": {
-        "photo": {
+        "driverPhoto": {
             "items": {
-                "$ref": "#/definitions/photo"
+                "$ref": "#/definitions/driverPhoto"
             },
             "type": "array",
             "options": {
@@ -518,9 +518,9 @@
             },
             "propertyOrder": 3
         },
-        "vehicle": {
+        "driverVehicle": {
             "items": {
-                "$ref": "#/definitions/vehicle"
+                "$ref": "#/definitions/driverVehicle"
             },
             "type": "array",
             "options": {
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
Non-ASCII characters will need to be allowed for support of Arabic-only identifiers.

flibbertigibbet:
I'd rather there not be any more character appending than necessary happening, as that will require further changes to the class file generator and mobile app to support.

flibbertigibbet:
With the randomly generated field name suffixes, what will happen if the schema changes slightly and a record is uploaded under the old schema? Will the field names still match?

notthatbreezy:
> Non-ASCII characters will need to be allowed for support of Arabic-only identifiers.

That's what the adding of random strings is supposed to handle, the attached screenshot shows what that will look like.

> I'd rather there not be any more character appending than necessary happening, as that will require further changes to the class file generator and mobile app to support.

Hmm, there needs to be some to deal with the first issue. Matching non-printable characters and stripping them is possible, but I have a feeling it will be more error prone than the method attached here which also captures cases where reserved words are used without bubbling the issue up to the user.

My understanding from @kshepard was that the field names were designed to be similar to the labels for debugging purposes, but if it simplifies things and keeps them similar for most debugging purposes then it would be ok to change it.

> With the randomly generated field name suffixes, what will happen if the schema changes slightly and a record is uploaded under the old schema? Will the field names still match?

Fieldnames will still match as far as my testing shows, existing ones remain unchanged.

old-fields-unchanged

scripts/incident_schema_v3.json
```diff
@@ -3,7 +3,7 @@
     "title": "",
     "plural_title": "",
     "definitions": {
-        "photo": {
+        "photo_zu3df": {
```
flibbertigibbet: Shouldn't these be prepended with 'driver' to match what the generator does now?

flibbertigibbet:
Some effort went into supporting Unicode identifiers and file names in the generated Java models with the expectation that we wanted to keep those as similar to the labels as possible.

Is there anything to handle potential collisions in the randomized/cleaned identifiers?

scripts/incident_schema_v3.json
notthatbreezy: Oops, good catch - added these changes before the `driver` prepending.

notthatbreezy:
No, it will be _very_ rare and they only have to be unique when they're generated - that functionality hasn't changed from what already existed.

I think in the ideal case this cleaning/verification should be handled on the back-end, but it seems like we're kind of beyond that at this point. I can increase the length of the random string attached if you're worried about name-collision.

flibbertigibbet:
I understand collisions are unlikely, but as uniqueness is a requirement, there should be something somewhere to ensure that, especially if Unicode-only names are allowed and so will have fully generated names. Will a schema with name conflicts fail to save?

notthatbreezy:
Updated the PR -- took out the randomness to preserve the current name checking behavior for clashes (which is to log to the console and not do anything) and replaced it with a regular expression that will strip control characters. 

I preserved the prepending of `driver` to the fieldname to protect against whitelisted words and to ensure that the fieldname meets the additional requirements of java identifiers.

Doesn't protect against cases where a client will interact with the API directly still.

scripts/incident_schema_v3.json
```diff
@@ -3,7 +3,7 @@
     "title": "",
     "plural_title": "",
     "definitions": {
-        "photo": {
+        "drivePhoto": {
```
kshepard: drivePhoto -> driverPhoto, a few places in this file.

kshepard:
This looks good to me from the web perspective, but I don't have much knowledge of the Java code that will be consuming this. @flibbertigibbet: do you still foresee this causing any issues with the Android library, or have your concerns been alleviated and this can be merged in?

flibbertigibbet:
I guess this should work. If there's really nothing in place to ensure uniqueness of field names, that seems deserving of an issue.

kshepard:
@notthatbreezy: not sure this strips all the necessary characters. I just tried adding a new section that had a '/' in it, and that didn't end up getting stripped out. The text I used to test it was of the form: "<some_arabic_text> / <some_english_text>".

PR #610, opened 2016-06-27
Send non-null min and max dates from filterbar

### LINKED ISSUE #607: Problem with time range on graph queries
To reproduce:
 * Open a the app fresh in an incognito browser tab
 * Browse to the map page
 * Update the start date time
 * See that there's a failed query for the stepwise endpoint because `occurred_min` is set, but `occurred_max` is not. The `occurred_max` is also not sent to the toddow endpoint, yet there's no error message.

jisantuc:
Fixes #607


kshepard:
+1

PR #612, opened 2016-06-28
Add print styles

jfrankl:
Print view:

image

Web view:

image


kshepard:
+1, looks great! There are some plain-text strings that I'll extract to the translation files before merging.

Issue #569, opened 2016-06-10
SA - Custom reports using Hijri calendar dates for period aggregation


ddohler:
I did some research and experimentation, and I think I came up with a way of calculating custom reports that will allow us to generate reports with Hijri dates in most cases. The basic idea is to replace the use of the built-in Postgres date extraction operators with equivalent sets of `Q` expressions on the `occurred_from` field. The boundaries of these `Q` expressions can then be adapted depending on the calendar system in which we want the results to be.

Below is a proof-of-concept for calculating the number of events occurring in the month of Ramadan over the past three years:
```python
from django.db.models import Case, When, IntegerField, Value, Count, Sum, Q
import datetime
# Define Gregorian cutoffs for the Islamic month of Ramadan in 2016, 2015, and                      
# 2014; in "real life" these would need to be calculated for each month of the                         
# Islamic year, over the years for which we are generating the report.
rs16 = datetime.datetime(2016, 6, 5, 17)
re16 = datetime.datetime(2016, 7, 5, 17)
rs15 = datetime.datetime(2015, 6, 17, 17)
re15 = datetime.datetime(2015, 7, 17, 17)
rs14 = datetime.datetime(2014, 6, 28, 17)
re14 = datetime.datetime(2014, 7, 28, 17)
# Define Q operators to determine if a record falls within the month of Ramadan                     
# during any of the past three years. These logical groupings will need to be                        
# generated for each month, grouped by year. 
is_r16 = (Q(occurred_from__gte=rs16)&Q(occurred_from__lt=re16))
is_r15 = (Q(occurred_from__gte=rs15)&Q(occurred_from__lt=re15))
is_r14 = (Q(occurred_from__gte=rs14)&Q(occurred_from__lt=re14))
is_ramadan = is_r16|is_r15|is_r14
# Define Case statement for whether an event falls into a particular month.                         
# I used a binary Ramadan / not-Ramadan condition here, but these will need to                      
# be implemented with a different number for each month, which is what the                          
# existing code does.
ramadan_case = Case(When(is_ramadan, then=Value(1)), default=Value(0),
                    output_field=IntegerField())
# This should slot in directly with our existing aggregation code.
recs = Record.objects.all().annotate(in_ramadan=ramadan_case)
recs.aggregate(Sum('in_ramadan'))
{'in_ramadan__sum': 379}
```

[...]

PR #613, opened 2016-06-28
Add date_utils for date range localization

Lknechtli:
Add ummalqura calendar support to custom reports date aggregation
Fix gregorian sequential weekly aggregation

### DIFF
--- a/app/data/localization/date_utils.py
+++ b/app/data/localization/date_utils.py
@@ -1,0 +1,368 @@
+from umalqurra.hijri_date import HijriDate
+from datetime import datetime, timedelta
+from collections import namedtuple
+import pytz
+
+from django.conf import settings
+
+
+DateRange = namedtuple('DateRange', ['start', 'end'])
+# ranges: array of DateRange's for the date setting
+HijriDateSet = namedtuple('HijriDateSet', ['key', 'label', 'ranges'])
+local_tz = pytz.timezone(settings.TIME_ZONE)
+
+
+def hijri_to_gregorian(umalqurra_date):
+    """Convert a HijriDate to a python datetime
+
+    Args:
+        umalqurra_date (HijriDate): date to convert
+    Return:
+        datetime.datetime
+    """
+    day = umalqurra_date.day_gr
+    month = umalqurra_date.month_gr
+    year = umalqurra_date.year_gr
+    gregorian_date = datetime(int(year), int(month), int(day))
+    return gregorian_date
+
+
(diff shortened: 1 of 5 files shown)


Lknechtli:
Refactoring the crosstabs route out into a separate view - it's up to 800ish lines now.

Lknechtli:
I may have the umalqurra library added as a requirement incorrectly - when re-provisioning, it wasn't added, and I had to do it manually.  Not sure why.

app/data/localization/date_utils.py
```diff
@@ -0,0 +1,384 @@
...
+            end_month = month + 1
+            if end_month >= 12:
+                end_year = year+1
+                end_month = 1
+
+            end_date_um = HijriDate(end_year, end_month, 1)
+            end_year_gr = int(end_date_um.year_gr)
+            end_month_gr = int(end_date_um.month_gr)
+            end_day_gr = int(end_date_um.day_gr)
+            end_date_gr = datetime(end_year_gr, end_month_gr, end_day_gr)
```
ddohler: Could this be generated using the `hijri_to_gregorian` function above?

Lknechtli: yup

kshepard:
@Lknechtli: I also had an umalqurra import issue after provisioning, but restarting the driver-app service fixed it. So it doesn't look like an issue with how the requirement is added.

app/data/views.py
```diff
@@ -649,6 +668,160 @@ def _make_period_case(self, period_type, request, queryset):
...
+            'month_of_year': {
+                'type': 'derived',
+                'query': hijri_month_range,
+            },
+            'week_of_year': {
+                'type': 'derived',
+                'query': hijri_week_range
+            },
+            'day_of_week': {
+                'type': 'source',
```
ddohler: I agree that these could be better named; perhaps `builtin` and `generated`?

kshepard:
It seems the 'day' aggregation in gregorian doesn't work (returns all zeros), yet the ummalqura version is fine. Think you'd be able to take a look at that while your head is still deep in this code?

ddohler:
+1 after squashing commits; it's a bit tough to tell if everything is working perfectly, but the Umm al-Qura aggregations seem to give roughly the same order of magnitude of events as their Gregorian counterparts.

kshepard:
+1 from me too. Code looks good and numbers seem to be in the right ballpark. Nice work!

Lknechtli:
closes #569

PR #615, opened 2016-06-29
Include rendering for numbers and integers

jisantuc:
In #598, I didn't include the directives and controllers for
rendering number and integer fields in the deployment index
file. Naturally, they didn't render, even though with my
non-deployment index.html they did. This fixes that and breaks
out the controllers/directives for number and integer types
in case we want to do things differently with them in the
future.


jisantuc:
@kshepard here's the fix for #598

kshepard:
+1

PR #619, opened 2016-07-13
Add propertyOrder on properties field

ddohler:
Previously the `propertyOrder` was being added to `definitions`, but other parts
of the UI and the Android app assume that it is on `properties`, so this
moves it there.

This can be tested by adding a related content type in the Ashlar editor to any existing schema; after that, all of the `properties` on that schema will have a defined `propertyOrder`.

We will need to manually update the schema on the Saudi demo instance with their preferred ordering once this is deployed.

### DIFF
--- a/schema_editor/app/scripts/views/recordtype/related-add-directive.js
+++ b/schema_editor/app/scripts/views/recordtype/related-add-directive.js
@@ -21,21 +21,21 @@
 
         /**
          * Updates related definitions to include a property order attribute
-         * @param {object} definitions to check for propertyOrder attribute
+         * @param {object} properties to update propertyOrder attribute
          */
-        function addPropertyOrdertoDefinitions(definitions) {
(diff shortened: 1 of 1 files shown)


kshepard:
LGTM. @flibbertigibbet - would you be able to verify that it's definitely the `properties` that the Android app is using for sorting, and not `definitions`?

flibbertigibbet:
Yes, the Java JSON schema parser looks for `propertyOrder` under the definitions, which worked with the way our schemas were generated originally. Also the Java parser project expects `propertyOrder` to be in that section as part of its base annotator class, which would seem to suggest that it's the generally expected place to define property ordering with JSON schemas.

https://github.com/WorldBank-Transport/jsonschema2pojo/blob/master/jsonschema2pojo-core/src/main/java/org/jsonschema2pojo/JsonEditorAnnotator.java#L39

PR #621, opened 2016-08-26
Remove invalid docker tag option to fix deployment

kshepard:
The `-f` option was marked as deprecated a while ago, but it appears
it's now officially phased out and will not run if specified.


kshepard:
Trivial, going to merge this to get a deployment to staging.

PR #618, opened 2016-07-07
Check for additional 'items' in schema path

kshepard:
Checkbox display types are similar to select list types except
for one additional 'items' component. This uses the 'items'
component if it exists when calculating costs.

Fixes #616

PR #622, opened 2016-10-13
Fix checkbox filtering

kshepard:


ddohler:
+1

KlaasH:
Looks good to me.  I see the front-end version of this checkbox/droplist issue was taken care of in #618.  And I have a vague memory of @flibbertigibbet mentioning having encountered and handled it in the Android app a while ago, so hopefully this is the last of it...

flibbertigibbet:
If the issue is limited to cross-field filtering, that should not affect the Android app (which does no record querying or filtering). If the structure of the JSON representation of select lists or checkboxes has changed or should be interchangeable after schema modification, that would affect the Android app.

PR #623, opened 2016-10-24
Some various bugfixes

kshepard:


ddohler:
I can't test at the moment but this all looks good; have you tested the weekday fix in Arabic?

kshepard:
The fix also works in Arabic. Thanks for the note about testing it though: I ended up finding and fixing another Arabic bug that was lurking. Also updated the letsencrypt role, since a url to the licensing agreement is no longer valid which was causing it to fail. Should be good to go now. I'm going to merge and deploy.

PR #625, opened 2017-02-23
Add Mapillary viewer

cbeddow:
Mapillary viewer has been added to the popup builders in the layer control. 

-May need adjusting for iframe dimensions of Mapillary viewer, to meet the popup dimensions
-May want to add it to the str variable in a different location (I added it at top, before other popup content)
-Need to ensure that the ajax call is working properly, so the mly variable is updated with the correct image key before being added to the popup content


cbeddow:
Had pull request from knikel shouldn't be included, will do a new pull request

PR #626, opened 2017-02-23
Add Mapillary viewer to blackspot popup builder

cbeddow:
Need to ensure this looks correct, if not could be problems with ajax call, or location of the iframes inside the popup builder

### DIFF
--- a/web/app/scripts/views/map/layers-controller.js
+++ b/web/app/scripts/views/map/layers-controller.js
@@ -692,7 +692,20 @@
          */
         ctl.buildBlackspotPopup = function(blackspot) {
             /* jshint camelcase: false */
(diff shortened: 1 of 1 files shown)


shreshthkhilani:
Hi @cbeddow, Azavea has started work on this feature for DRIVER. The client ID you're using here is invalid. Is there another one we could use?

cbeddow:
Hello,

The developers will need to register a new client ID at
https://Mapillary.com/developer. It will not by default include access to
the traffic sign API, because it is a closed API. You will need to send a
message to sales@mapillary.com (even for non-sales approval) to request
this access formally.

This client ID was a temporary one so is not reusable.

PR #636, opened 2017-04-17
Fix Vietnamese translation for "EMPTY"

kshepard:
Trivial change. Having a null value for a translation was causing the rendering code to blow up and print a console error.


ddohler:
Oops, my bad, I wrote a script to copy over the translations and it looks like I missed this. :+1:

PR #637, opened 2017-06-05
Add translations for Chinese and Spanish

kshepard:


kshepard:
Trivial/low-risk change that's hard to test due to a dependency change in azavea/calendars. Going to merge this in order to test it out on the demo server.

PR #639, opened 2017-08-24
Add Bengali translations

kshepard:


kshepard:
This is a low-risk change that's hard to test due to a dependency change in azavea/calendars. Going to merge this in order to test it out on the demo server.

PR #642, opened 2017-10-10
Feature/add lao translations

tgilcrest:
Adding `lo.json` file to provide Lao translations


kshepard:
+1, tested it out and it looks good. I also snuck a Lao calendar configuration commit into here, since it's related.

PR #645, opened 2017-12-05
Make provisioning from scratch work

jisantuc:
Overview
-----

This PR makes several small changes to make provisioning from scratch cooperate given changes that have occurred in the state of dependencies since this project's dev environment was setup.

Specifically, it explicitly sets `bower`'s registry, separately installs `bower`, changes the `libpq` version to something that's in the postgresql PPA, upgrades the base images for some containers to `node:4-slim` from our old `static-core` containers.

Testing
-----

- `vagrant provision`
- it should succeed

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,17 @@ telling it to communicate with the Docker daemon on the `app` virtual machine.
 
 ## Testing Data
 
-A CSV of historical data can be downloaded from the project /data folder (incidents_and_sites.csv).
+### Boundaries
+
+To load boundaries, upload the `regions.zip` and `states.zip` files to Ashlar.
+Ashlar is runs on localhost:7001. For each file, first upload the file, then select
+`name` as the display field, then hit save. Either refresh the page or
+navigate somewhere else in between any two uploads.
+
+### Records
+
+A CSV of historical data can be downloaded from the project /data folder.
+Good files are `<city or agency>_traffic.csv`.
 
 Once the app has been built, this data can be loaded.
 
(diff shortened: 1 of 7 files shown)


deployment/ansible/group_vars/all.example
```diff
@@ -19,7 +19,7 @@ postgresql_port: 5432
 postgresql_version: "9.4"
 postgresql_package_version: "9.4.*.pgdg14.04+1"
 postgresql_support_repository_channel: "main"
-postgresql_support_libpq_version: "9.5.*.pgdg14.04+1"
+postgresql_support_libpq_version: "9.3.4-1"
```
ddohler: I'd feel more comfortable if we find a way to make this at least `9.4`; JSONB fields weren't introduced until Postgres 9.4, so it'd be nice if all the versions can match up. I'm not sure whether we're actually _using_ libpq though, so if it's not breaking anything for you locally we can punt and make a separate issue for it.

jisantuc: These are the versions available in the services VM:

```bash
$ sudo apt-cache show libpq-dev | grep Version
Version: 10.1-1.pgdg14.04+1
Version: 9.3.20-0ubuntu0.14.04
Version: 9.3.4-1
```

It might make sense to open a separate (probably small) issue

schema_editor/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/azavea/static-core:0.1.0
+FROM node:4-slim
```
ddohler: Once this is merged I think we can let ops know that they can remove this image from Quay; I assume it's not being used anywhere else.

deployment/ansible/group_vars/all.example
ddohler: Yeah, we're relying on the `azavea.postgresql-support` role to install `libpq-dev`. The `azavea.postgresql` role seems to have support for installing a wide range of postgres versions, but the `azavea.postgresql-support` role doesn't seem to have been updated quite as much. If using 9.4 for the version doesn't work we'll probably need an ops issue to add support to that role.

README.md
```diff
@@ -97,6 +107,17 @@ To load mock interventions, run `python scripts/load_interventions.py --authz 'T
 
 To generate black spot and load forecast training inputs, run `python scripts/generate_training_input.py /path/to/roads.shp /path/to/records.csv`.
 
+### Costs
+
+You can't request records with associated costs successfully until you configure some costs.
```
ddohler: Wait, are you saying that all record requests fail if this isn't configured? We should probably make an issue for that if so, it's not supposed to impact the rest of the app.

jisantuc: Not _all_ records requests (like, you can hit `/api/records/` successfully), but you can't hit `/api/records/costs/`, which is required for a lot of the app. I _think_ you can view the records list but I don't remember.

PR #652, opened 2017-12-15
Feature/add thai translations

tgilcrest:
Adding translations for the Thai language to json file.


tgilcrest:
Still need to add "Modified" and "Modified by" since those are new with the username / email auditing feature we developed.

PR #656, opened 2017-12-19
Support tagged releases in Travis, Ansible

ddohler:
# Overview
This configures Travis to deploy Docker images with tags matching
the built tag on the master branch. It also adds an Ansible variable
called `app_version` that can be used to specify which Docker image
gets pulled onto host machines.

The general scheme here is as follows: Previously, merges to `develop` would deploy Docker images with the `latest` tag, and all deployments of DRIVER currently are hard-coded to pull and run the `latest` image. Merges to `master` had essentially no effect. Now, merges to `develop` continue to behave the same way, but adding new tags to `master` also deploys images which are tagged to match the git tag. In this way, we can have deployments that are pinned to specific versions which will be unaffected by subsequent merges to `master`.

Tagged images can be pulled and run by our deployment scripts by specifying the `app_version` parameter in `group_vars`. If `app_version` is `development`, then the `latest` docker tag will be pulled. Otherwise, the value of `app_version` will be used. `app_version` is ignored in Vagrant because there we build Docker images based on the application source, rather than pulling from Quay.

Because the behavior of merging to `develop` is unchanged, this should be backwards-compatible with existing deployments (although I don't think we were relying on this specific behavior anyway).

# Testing instructions
- This isn't easily testable locally since it affects production deployments, but local provisioning should still work.
- In order to fully test this, we will need to:
  - Merge this
  - Cut a 1.0 (or some other number) release to `master`
  - Confirm that we get properly tagged images in quay.
  - Attempt to deploy or update one of our instances with the new version number and confirm that it succeeds.

# Notes
- This still doesn't support forks because only tags to `master` result in new docker images. This is by design in order to be conservative about how much fragmentation we encourage in the ecosystem until we have more people using this in production. If it becomes necessary in the future, it's easy to remove this restriction; it should be as simple as changing `branch: master` in `.travis.yml` to `all_branches: true`. This would then deploy images for all tags, regardless of the branch, which would in theory allow us to run several parallel branches with very different features (e.g. `master-br`, `master-ph`, `master-sa`, etc.).
- The `1.0.0` placeholder in `all.example` should be changed to the actual first version number we want to use before we merge to `master`.
- I think that these Travis changes should be adaptable to the other deployment options we have been considering, such as whether to switch to a `docker-compose` based setup for easier local development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,20 @@ before_deploy:
 - export PATH="$PATH:$HOME/.rvm/bin"
 - export PHANTOMJS_CDNURL=http://cnpmjs.org/downloads
 deploy:
-  provider: script
-  script: ".travis/deploy.sh"
-  skip_cleanup: true
-  on:
-    repo: WorldBank-Transport/DRIVER
-    branch: develop
+  - provider: script
+    script: ".travis/deploy_staging.sh"
+    skip_cleanup: true
+    on:
+      repo: WorldBank-Transport/DRIVER
+      branch: develop
+  - provider: script
+    script: ".travis/deploy.sh"
+    skip_cleanup: true
+    on:
+      repo: WorldBank-Transport/DRIVER
+      branch: master
+      tags: true
+
 after_deploy:
 - rm deployment/driver.pem
 - rm gradle/data/driver.keystore
--- a/.travis/deploy.sh
+++ b/.travis/deploy.sh
@@ -5,35 +5,7 @@ set -x
 
 for image in app editor web gradle analysis;
 do
-  docker push "quay.io/azavea/driver-${image}:${TRAVIS_COMMIT:0:7}"
-  docker tag "quay.io/azavea/driver-${image}:${TRAVIS_COMMIT:0:7}" "quay.io/azavea/driver-${image}:latest"
-  docker push "quay.io/azavea/driver-${image}:latest"
+  docker push "quay.io/azavea/driver-${image}:${TRAVIS_TAG}"
+  docker tag "quay.io/azavea/driver-${image}:${TRAVIS_TAG}" "quay.io/azavea/driver-${image}:${TRAVIS_TAG}"
+  docker push "quay.io/azavea/driver-${image}:${TRAVIS_TAG}"
 done
-
-# TODO: The below was for deploying to our staging environment, which we no longer have.
-# If staging is ever resurrected, we can re-enable this.
-# Set up group vars
-#echo "setting up group vars..."
-#set +x
-#grep -v nominatim_key deployment/ansible/group_vars/all.example \
-#    > deployment/ansible/group_vars/all
-#echo "web_js_nominatim_key: \"${NOMINATIM_API_KEY}\"" \
-#    >> deployment/ansible/group_vars/all
-#echo "oauth_client_id: \"${OAUTH_CLIENT_ID}\"" \
-#    >> deployment/ansible/group_vars/all
-#echo "oauth_client_secret: \"${OAUTH_CLIENT_SECRET}\"" \
-#    >> deployment/ansible/group_vars/all
-#echo "forecast_io_api_key: \"${FORECAST_IO_API_KEY}\"" \
-#    >> deployment/ansible/group_vars/all
-#echo "keystore_password: \"${DRIVER_KEYSTORE_PASSWORD}\"" \
-#    >> deployment/ansible/group_vars/all
-#set -x
-
-#ansible-galaxy install -f -r deployment/ansible/roles.yml -p deployment/ansible/roles
-
-#chmod 0600 deployment/driver.pem
-
-#ANSIBLE_HOST_KEY_CHECKING=False ansible-playbook \
-#  -i deployment/ansible/inventory/staging \
-#  --private-key="${TRAVIS_BUILD_DIR}/deployment/driver.pem" \
-#  deployment/ansible/database.yml deployment/ansible/app.yml deployment/ansible/celery.yml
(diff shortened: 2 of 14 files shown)


deployment/ansible/roles/driver.app/tasks/main.yml
```diff
@@ -17,15 +17,7 @@
...
     docker build
       -f {{ root_app_dir }}/Dockerfile.base
-      -t quay.io/azavea/driver-app:latest
-      {{ root_app_dir }}
-  when: developing
-
-- name: Build application Docker image
-  command: >
-    docker build
-      -f {{ root_app_dir }}/Dockerfile.development
```
kshepard: Is `Dockerfile.development` no longer used? If so, we should remove it.

kshepard: Also, I see a reference to `driver-app:latest` a few lines down (line 25), that may need to be modified, but I'm not certain.

.travis.yml
```diff
@@ -33,12 +33,20 @@ before_deploy:
...
+    on:
+      repo: WorldBank-Transport/DRIVER
+      branch: develop
+  - provider: script
+    script: ".travis/deploy.sh"
+    skip_cleanup: true
+    on:
+      repo: WorldBank-Transport/DRIVER
+      branch: master
+      tags: true
```
hectcastro: This may lead to two deployments that race, at least if we use `git-flow`. I think that the completion of a release branch will trigger a build for the merge into `master`, but also a build for the tagged commit.

.travis/deploy.sh
```diff
@@ -5,35 +5,7 @@ set -x
 
 for image in app editor web gradle analysis;
 do
-  docker push "quay.io/azavea/driver-${image}:${TRAVIS_COMMIT:0:7}"
-  docker tag "quay.io/azavea/driver-${image}:${TRAVIS_COMMIT:0:7}" "quay.io/azavea/driver-${image}:latest"
-  docker push "quay.io/azavea/driver-${image}:latest"
+  docker push "quay.io/azavea/driver-${image}:${TRAVIS_TAG}"
```
hectcastro: I think that if we remove `on: master` above, this will work. Otherwise, I think the `TRAVIS_TAG` environment variable will be unset when a build is triggered due to changes to `master`.

doc/system-administration.md
```diff
@@ -28,9 +28,11 @@ Each of these servers also includes:
...
+Deploying updates to production is done using [Ansible](https://www.ansible.com/) (version must be at minimum 1.8). An `ansible-playbook` command is run, which uses the local configuration of the application files in the `deployment` directory to deploy updates to the remote servers. It is not necessary to have the application running locally, but it is necessary to have the source code for the version you want to deploy, so make sure to pull down the latest version via `git pull` and then `git checkout tags/<version>. In addition to the latest source code, there are four files not checked in to the repository that are needed in order to successfully deploy.  These files are as follows
```
hectcastro: I think that there is a missing backtick toward the bottom of this paragraph:

> ...so make sure to pull down the latest version via `git pull` and then `git checkout tags/<version>`. In a...

deployment/ansible/roles/driver.app/tasks/main.yml
hectcastro: Yeah, I think it needs to reference `docker_image_tag`.

deployment/ansible/group_vars/all.example
```diff
@@ -9,6 +12,8 @@ production: "{{ 'production' in group_names }}"
 docker_version: "1.10.*"
 docker_py_version: "1.2.3"
 docker_options: "--storage-driver=aufs"
+# meta-variable used to translate app_version into docker container tag
+docker_image_tag: "{% if app_version == 'development' %}latest{% else %}{{app_version}}{% endif %}"
```
hectcastro: Is it expected that `app_version` will always be `development` when you're developing locally?

.travis.yml
ddohler: My understanding from the docs is that `branch` and `tags` are AND-ed together such that a build is only kicked off if it satisfies both conditions:
> When _all_ conditions specified in the on: section are met, your build will deploy.

https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A 

But I have no experience with this aspect of travis so it's definitely possible that that's not quite how it works.

.travis/deploy.sh
ddohler: See above; my understanding from the docs is that `tags: true` will prevent this from running if `TRAVIS_TAG` is unset.

deployment/ansible/group_vars/all.example
ddohler: No; `app_version` is ignored completely if developing locally. I'll make that clearer in the comment up top.

.travis.yml
hectcastro: Hm. This may be true then. I was basing my understanding off of visually seeing builds get triggered for `master` and tagged commits whenever a release gets finalized (merges into `master`, `develop`, and tagged commit). It is possible that at the end of a build it ANDs these conditions together to make a decision.

.travis/deploy.sh
hectcastro: Yeah, this seems to back that up too:

> tags: true: deployment is triggered if and only if $TRAVIS_TAG is set. Depending on your workflow, you may set $TRAVIS_TAG explicitly, even if this is a non-tag build when it was initiated. This causes the branch condition to be ignored.

ddohler:
Thanks for the comments; they made me realize that I hadn't fully thought through how this would interact with local development (e.g. `scripts/grunt.sh`). I've modified things so that `app_version` truly has no effect in development, which allows things like `scripts.grunt.sh` and the Docker image `Makefile` to remain unchanged. The production deployment approach hasn't changed. @kshepard this might be worth another once-over, but should be basically ready to go, I think, so feel free to merge since I'll be out for the next few days.

kshepard:
Cool, this looks good, I'll merge it.

PR #657, opened 2018-01-02
Add CertBot to Ansible

ddohler:
# Overview
Adds Ansible role for management of Let's Encrypt certificates via CertBot.

# Testing instructions
- Log into the AWS demo account and bring up two EC2 machines based on the Ubuntu Trusty AMI, one for the web app and one for the database. They should be placed in the `DriverDemoAppSG` and `DriverDatabaseSG`, respectively. It may make things smoother to bump the boot volume size to 12GB from 8, but otherwise the defaults should be fine. These won't be actually used for anything except testing provisioning, so `t2.small`s should be enough horsepower.
- Attach the free Elastic IP address (there should only be one in the account) to the web app machine.
- Follow the instructions in [doc/system-administration.md](https://github.com/WorldBank-Transport/DRIVER/blob/feature/lets-encrypt-certbot/doc/system-administration.md) to get set up to do a deployment. The best way to get the `production` group_vars and inventory files is probably to grab the ones on the fileshare for the Thai instance. You will need to edit the group_vars file you use to add `app_version: 'latest'`, because that parameter has been added since the last time we launched an instance. You'll also need to edit the IP addresses to match the instances you brought up, and `s/thailand.roadsafety.io/certbot.roadsafety.io/g` (which is already set up in Route 53 to point to that Elastic IP).
- Run the `ansible-playbook` command from the system administration doc to deploy, but omit the `celery.yml` playbook to save time.
- It will fail 😞 on the `Set table-level access privileges for Windshaft DB role` task. The error message will vary depending on your version of Ansible, but the cause is that migrations haven't been run. SSH into the web instance, `sudo docker exec -ti driver-app ./manage.py migrate`. (It may be possible to avoid this step by commenting out line 58 in [roles/driver.app/tasks/main.yml](https://github.com/WorldBank-Transport/DRIVER/blob/feature/lets-encrypt-certbot/deployment/ansible/roles/driver.app/tasks/main.yml#L58) prior to running the deployment, but that's not how I did it). Also, I swear this used to work, so I'm not sure what changed, but there doesn't seem to be an easy way to specific cross-play dependencies in Ansible, so I haven't found an easy fix.
- Run provisioning all over again.
- Once it finishes, navigate to https://certbot.roadsafety.io , and you should get the login screen with a valid security certificate.
- Clean up by deleting the EC2 instances, and you can also get rid of the Elastic IP address and remove the test subdomain.

### DIFF
--- a/deployment/ansible/app.yml
+++ b/deployment/ansible/app.yml
@@ -13,15 +13,6 @@
 
   become: True
 
-  vars:
-    letsencrypt_certs:
-      -
-        name: "driver-ssl-cert"
-        keypath: "{{ letsencrypt_key_dir }}/driver-ssl-cert.key"
-        certpath: "{{ letsencrypt_certs_dir }}/driver-ssl-cert.cert"
-        fullchainedcertpath: "{{ letsencrypt_certs_dir }}/driver-ssl-cert-full.cert"
-        host: "{{ allowed_host }}"
-
   pre_tasks:
     - name: Update APT cache
       apt: update_cache=yes
@@ -33,7 +24,6 @@
     - { role: "driver.windshaft" }
     - { role: "driver.app" }
     - { role: "driver.nginx" }
-    - { role: "azavea.letsencrypt", when: staging or production }
 
 - hosts: database-servers
 
--- a/deployment/ansible/roles.yml
+++ b/deployment/ansible/roles.yml
@@ -14,5 +14,3 @@
   version: 0.1.1
 - src: azavea.redis
   version: 0.1.0
-- src: azavea.letsencrypt
-  version: v11.3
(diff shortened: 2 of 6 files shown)


jisantuc:
A few things I had to do to get testing to work:

- `pip install -U pip` in the database instance, but not the app instance (probably because docker handles everything we care about with pip in the app, but i didn't investigate to make sure)
- perform a series of incantations on the database instance to make provisioning succeed. eventually I changed nothing that should affect state (I `psql`'ed a few times from the instance) and provisioning succeeded instead of failed
- add `docker_image_tag` to `group_vars`
- update the app instance's private IP in firewall settings

But: success!

image

PR #664, opened 2018-01-11
Simplify and document production group_vars

ddohler:
# Overview

This adds a heavily-commented `production.example` file to `deployment/ansible/group_vars`, which is intended to be used as a template for production deployments and make the process easier for users.

# Testing instructions
1. This makes some changes to `group_vars/all.example`, so `cp group_vars/all.example group_vars/all` and make sure provisioning still works.
2. Read through the comments in `production.example` and make sure they are clear and comprehensive enough for someone unfamiliar with DRIVER to get it set up successfully.

# Notes
We also should test this by actually provisioning an instance, but we have at least a couple upcoming chances to do that. So I think it will be more efficient to roll testing this into one of the upcoming deployments and avoid the overhead of spinning up an extra set of testing machines.

### DIFF
--- a/deployment/ansible/group_vars/production.example
+++ b/deployment/ansible/group_vars/production.example
@@ -1,0 +1,241 @@
+---
+## GLOBAL APP SETTINGS
+# Which version of the app to install. Specifically, this designates a tag on quay.io that should be
+# used when pulling docker images. A list of valid tags can be found here:
+# https://quay.io/repository/azavea/driver-app?tag=latest&tab=tags
+# Generally, when setting up a new instance, you should use the highest-number tag available. The
+# 'latest' tag should be avoided when setting up new instances for production usage, because it may
+# contain bugs.
+app_version: "0.0.0"
+# The URL at which users will access the application. Do NOT include 'http' or 'https'; all DRIVER
+# instances are configured to use https.
+app_domain_name: "your.driver-domain.here"
+
+# The *internal* IP address of the hosts for the different components of the web application. DRIVER
+# is currently divided up to run on three servers: A database, a web app server, and a Celery
+# (asynchronous tasks) server. If your servers have only a single IP address, then use that. If they
+# have multiple IP addresses, use the internal, non-public-facing IP address of the database server.
+database_server_ip: "11.11.11.11"
+app_server_ip: "22.22.22.22"
+celery_server_ip: "33.33.33.33"
+
+## POSTGRESQL SETTINGS
+# The username, password, and database name that the main web app will use when connecting to the
+# database server. These do not need to already exist on the database server; they will be added
+# automatically as part of the setup process. You should put in a long (at least 25 characters)
+# random string for the password.
+postgresql_username: driver
+#postgresql_password: "uncomment and change me!"
+postgresql_database: driver
+
+# The username and password that the tile server will use when connecting to the database. This
+# will be added automatically as part of the setup process. This should be different from the
+# previous user/password because the two users have different access privileges, although they both
+# use the same database.
+windshaft_db_username: windshaft
+#windshaft_db_password: "uncomment and change me!"
+
+# Heimdall is a utility used for managing distributed asynchronous processes by using database
+# locks. It needs a username, password, and its own (very small) database in order to run.  This
+# section specifies the account information and database that Heimdall should use when connection to
+# the database. This will be added automatically as part of the setup process. This should be
+# different from the previous user/passwords because the Heimdall user has different access
+# privileges.
+heimdall_db_username: heimdall
+#heimdall_db_password: "uncomment and change me!"
+heimdall_lock_db: lockspace
+
+# UNCOMMENT one of the two examples below.
+# The whitelist of hosts which are allowed to connect to the database, which will be used to modify
+# PostgreSQL's pg_hba.conf file. Generally, this should contain two entries, containing the internal
+# IP addresses of the Celery machine and the Web server machine. For example (using variable
+# substitution), if deploying on bare metal:
+# postgresql_hba_mapping:
+#   - { type: "host", database: "all", user: "all", address: "{{ app_server_ip }}", method: "md5" }
+#   - { type: "host", database: "all", user: "all", address: "{{ celery_server_ip }}", method: "md5" }
+#
+# Alternatively, you can use a subnet range to specify the hosts that will be allowed to connect in a
+# single entry. This should only be used if your hosts have private IP addresses that they use to
+# connect to one another; if you use a range of public addresses it could open your database server
+# to attacks.
+# For example (if deploying into an AWS VPC):
+# postgresql_hba_mapping:
+#   - { type: "host", database: "all", user: "all", address: "172.31.0.0/16", method: "md5" }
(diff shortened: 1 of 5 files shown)


deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
+---
+## GLOBAL APP SETTINGS
+# Which version of the app to install. Specifically, this designates a tag on quay.io that should be
+# used when pulling docker images. A list of valid tags can be found here:
+# https://quay.io/repository/azavea/driver-app?tag=latest&tab=tags
+# Generally, when setting up a new instance, you should use the highest-number tag available. The
+# 'latest' tag should be avoided when setting up new instances for production usage, because it may
```
jisantuc: Should this include a link to the sem ver spec? I'm imagining someone upgrading from 1.1.1 to 1.3.4 and arguing they weren't told they shouldn't do it. Something like "Upgrades where only the last version number has changed should be seamless. Other upgrades may take more effort. For more information, see the semantic versioning spec at https://semver.org/spec/v2.0.0.html"

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# contain bugs.
+app_version: "0.0.0"
+# The URL at which users will access the application. Do NOT include 'http' or 'https'; all DRIVER
+# instances are configured to use https.
+app_domain_name: "your.driver-domain.here"
+
+# The *internal* IP address of the hosts for the different components of the web application. DRIVER
+# is currently divided up to run on three servers: A database, a web app server, and a Celery
```
jisantuc: `s/currently divided up to run on/runs on` -- I don't know how much the potential for future architecture revisions needs to be user-facing

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# (asynchronous tasks) server. If your servers have only a single IP address, then use that. If they
+# have multiple IP addresses, use the internal, non-public-facing IP address of the database server.
+database_server_ip: "11.11.11.11"
+app_server_ip: "22.22.22.22"
+celery_server_ip: "33.33.33.33"
+
+## POSTGRESQL SETTINGS
+# The username, password, and database name that the main web app will use when connecting to the
+# database server. These do not need to already exist on the database server; they will be added
```
jisantuc: Is it fine if they do?

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# connect to one another; if you use a range of public addresses it could open your database server
+# to attacks.
+# For example (if deploying into an AWS VPC):
+# postgresql_hba_mapping:
+#   - { type: "host", database: "all", user: "all", address: "172.31.0.0/16", method: "md5" }
+# More information about pg_hba.conf is available here:
+# https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html
+
+## APP SETTINGS
+# These keys are used for security purposes; the should be set to different, long (50+ characters),
```
jisantuc: `s/the/they/`

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# https://<your domain name>/openid/callback/login
+# https://<your domain name>/openid/callback/login/
+# https://<your domain name>/openid/callback/logout
+# https://<your domain name>/openid/callback/logout/
+oauth_client_id: ""
+oauth_client_secret: ""
+
+## Localization settings
+# The name of the time zone in which you want your DRIVER instance to operate; this should usually
+# be the time zone for your country. If your country has multiple time zones, pick one. A list of
```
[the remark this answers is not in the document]
jisantuc: I also think this is funny

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# The latitude and longitude of where the maps in your installation of DRIVER should center by
+# default. This is given as [lat, lon]
+local_center_lat_lon: [14.5995, 120.984]
+# DRIVER has the ability to calculate "black spots", which are stretches of road or intersections
+# that experience relatively higher crash rates. The calculation process uses extracts of Open
+# Street Map data. This parameter specifies where to download that extract from. GeoFabrik maintains
+# extracts for most countries and regions in the world; for most countries, you should be able to
+# simply search https://download.geofabrik.de until you find your country and then use the link for
+# your country's extract. If you are trying to run DRIVER on a sub-national area, you may want to
+# generate your own extract for efficiency (the black spot calculation will run more quickly on
```
jisantuc: Docs on how to generate their own extract? Like "download the `.osm` and convert to a protobuf for your area of interest"?

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# smaller areas), but the national-level extract can still be used.
+osm_extract_url: "https://download.geofabrik.de/asia/philippines-latest.osm.pbf"
+
+# DRIVER provits a Monit interface for monitoring and inspecting the status of the various
```
jisantuc: `provits`?

deployment/ansible/group_vars/production.example
```diff
@@ -0,0 +1,241 @@
...
+# application components.
+# If you do not plan to use Monit, security can be increased by disabling it completely, so this is
+# the default. Set this to "true" to enable it.
+monit_enable_web_access: "false"
+# Because Monit enables low-level administration of the DRIVER application, access should be
+# restricted only to systems administrators. This can be accomplished by limiting the IP addresses
+# from which the Monit interface will accept connections, as well as via password. For more
+# information about Monit, see https://mmonit.com/monit/
+# You can connect to Monit at port 2812.
+# The hosts from which to allow connections; this should be tightly limited to just the IP address
```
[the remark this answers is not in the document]
jisantuc: Or just drop the adverb entirely

deployment/ansible/group_vars/production.example
ddohler: Yup, should be, I'll add a note about that.

deployment/ansible/group_vars/production.example
ddohler: Thinking more on this, I don't think that asking someone to generate _and host_ their own OSM extracts is very reasonable. Luckily, the OSM wiki has links to a bunch more extract resources that have cities and sub-national extracts, so I'm going to update to include a link to that.

deployment/ansible/group_vars/production.example
ddohler: Yeah, this is a good point. I decided to add some additional suggestions about consulting the changelog and setting up a testing instance.

ddohler:
Yeah, that's a good idea. I didn't make the sections more prominent this time around because it became clear to me while I was doing the comments that the existing section divisions don't really make sense -- some sections only have one item, while others are missing items that are really related but are located elsewhere. I haven't come up with a comprehensive reorganization yet, but I think that'll have to happen as part of the wizard work in order for the order of questions to make any sense, so I'm going to hold off until that happens.

I'm going to go ahead and merge because it looks like the one Travis failure was a transient NPM issue.

PR #667, opened 2018-01-24
Config Wizard #154147803

pcaisse:
## Overview

Add configuration wizard to facilitate deployment.

## Testing Instructions

* Read documentation to see if it gives enough info and is helpful and well-written
* Ensure you have Python 3.6 and pip installed (outside VM)
* Run `./scripts/setup` outside VM and fill in values at prompts.
* Check to see that `deployment/ansible/group_vars/production` and `deployment/ansible/inventory/production` were generated and look correct.
* Check that new files are ignored

# PivotalTracker

https://www.pivotaltracker.com/story/show/154147803


doc/system-administration.md
```diff
@@ -38,9 +42,17 @@ Deploying updates to production is done using [Ansible](https://www.ansible.com/
...
+### Configuration wizard
+
+In addition to a fully manual deployment, you can run the setup wizard at `scripts/setup` which will generate barebones versions of the necessary files for you (`deployment/ansible/group_vars/production`) and (`deployment/ansible/inventory/production`). You will still need to edit these files as per the environment and locale but the wizard can be helpful in getting started.
+
+### Deploy
+
+Once the files are in place, deployment may be performed by opening a terminal, switching to the directory of the DRIVER source code and running the command (NOTE: Make sure to change the `user` argument as appropriate):
 ```
-ansible-playbook -i deployment/ansible/inventory/production \
+ansible-playbook -i deployment/ansible/inventory/production --user=ubuntu \
```
ddohler: This should probably make clear that ubuntu is meant to be changed by the user to a username that has sudo privileges on their servers.

scripts/setup.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@click.command()
+@click.option(
+    '--app_domain_name',
+    prompt='App domain name (without http:// or https://)',
+)
+@click.option(
+    '--database_server_ip',
+    prompt='Database server IP',
```
ddohler: I think these should be clear that these three  are the _internal_ IPs, which the servers will use to talk to one another; they may not necessarily be the same as the IPs used in the inventory file.

deployment/ansible/inventory/production.j2
```diff
@@ -0,0 +1,17 @@
+app ansible_ssh_host={{ app_server_ip }}
```
ddohler: In our most common use-case, which is EC2 instances in a VM, these IP addresses are different from the IP addresses used in the group_vars. For example, for the Mumbai instance, the group_vars IPs look like this:
- `app_server_ip`: `172.31.25.22`
- `celery_server_ip`: `172.31.26.26`
- `database_server_ip`: `172.31.24.80`

But the inventory / public IPs are different; the App Server's public IP is `54.148.118.47`, for example. 

On the other hand, for the Philippines deployment, each machine has only one IP, the public IP.

I think there are a couple ways to handle this. We could add a branch to the wizard, which would prompt the user with a question like "Are you running your servers in a virtual private cloud?" and then additionally prompt for different IPs if so. Alternatively, we could leave this as-is, and just add a note after the wizard finishes saying something like "If your servers have different internal and external IP addresses, you'll need to edit `deployment/ansible/inventory/production` to add the proper external addresses."

I'm leaning toward the second one because there is no guarantee that the instances will have publicly-available IPs at all; a more secure way to administer the app would be not to allocate any public IPs and to just administer everything from a "bastion" instance that has access to the inside of the VPC.

pcaisse: How about if we prompt for public IPs for each server and you can leave it blank if it's n/a? That would work for our most common use case.

I think for now it makes sense to make this wizard work for you and Kenny, the people actually doing the deployments, so we should cater to our common use case without totally mystifying (and hopefully helping) anyone deploying DRIVER for the first time.

ddohler: Yeah, that seems reasonable, we can just note that it's okay to leave those blank.

PR #669, opened 2018-01-29
Add CloudFormation template for demo resources

kshepard:
### Overview

This adds an AWS CloudFormation template that can be used to bring up the three EC2 instances and secure access to them via Security Groups. This does not attempt to do any provisioning -- it simply brings up vanilla Ubuntu 14.04 instances which can then be provisioned as needed.

### Testing

 * Sign into AWS and go to the `CloudFormation` page
 * Launch a new stack with this template
 * Make sure to enter a `KeyName` and a `NameSuffix`
 * Ensure all three instances launch as expected, and check to make sure the Security Groups look accurate.
 * Make sure to delete the stack afterwards

### Notes

I didn't see a good place in the README to slot in a description for how to use this template, but it might be a good idea to add some text for it when we add more deployment-related documentation.

### DIFF
--- a/deployment/demo-cfn-template.yaml
+++ b/deployment/demo-cfn-template.yaml
@@ -1,0 +1,189 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'DRIVER demo stack -- EC2 instances with Security Groups'
+Parameters:
+  KeyName:
+    Description: Name of an existing EC2 KeyPair to enable SSH access to the instance
+    Type: AWS::EC2::KeyPair::KeyName
+    ConstraintDescription: must be the name of an existing EC2 KeyPair.
+  AppInstanceType:
+    Description: Web app EC2 instance type
+    Type: String
+    Default: t2.medium
+    AllowedValues: [t2.nano, t2.micro, t2.small, t2.medium, t2.large, t2.xlarge, t2.2xlarge]
+    ConstraintDescription: must be a valid EC2 instance type in the T2 family.
+  DatabaseInstanceType:
+    Description: Database EC2 instance type
+    Type: String
+    Default: t2.medium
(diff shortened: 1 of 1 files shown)


deployment/demo-cfn-template.yaml
```diff
@@ -0,0 +1,189 @@
...
+  KeyName:
+    Description: Name of an existing EC2 KeyPair to enable SSH access to the instance
+    Type: AWS::EC2::KeyPair::KeyName
+    ConstraintDescription: must be the name of an existing EC2 KeyPair.
+  AppInstanceType:
+    Description: Web app EC2 instance type
+    Type: String
+    Default: t2.medium
+    AllowedValues: [t2.nano, t2.micro, t2.small, t2.medium, t2.large]
+    ConstraintDescription: must be a valid EC2 instance type.
```
ddohler: It looks like there's not a good built-in solution for validating instance types. Could we add change this to `a valid EC2 instance type in the T2 family` and add `t2.xlarge` and `t2.2xlarge` to the list?

deployment/demo-cfn-template.yaml
```diff
@@ -0,0 +1,189 @@
...
+      GroupDescription: Allow inbound traffic for SSH, Monit, and HTTP/S
+      SecurityGroupIngress:
+      - IpProtocol: tcp
+        FromPort: 22
+        ToPort: 22
+        CidrIp: !Ref 'SSHCidr'
+      - IpProtocol: tcp
+        FromPort: 2812
+        ToPort: 2812
+        CidrIp: !Ref 'WebCidr'
```
ddohler: I think `SSHCidr` would be more appropriate here, since monit is more concerned with administrative tasks (same comment for the other places this appears).

kshepard: Both are great points. I've made the changes.

PR #670, opened 2018-01-30
Template out docker repository

shreshthkhilani:
## Overview 

Template out docker repository to help decentralize architecture. 

### Notes

`Makefile` needs to run with `DOCKER_REPOSITORY` environment variable, as does `forecast_segment_incidents.py`.

## Testing

Test build on local development machine. How can we test with a new docker repository?


Makefile
```diff
@@ -1,13 +1,17 @@
 all: app editor web
 
+ifeq ($(DOCKER_REPOSITORY),)
+    DOCKER_REPOSITORY := quay.io/azavea
```
ddohler: I'm not sure that the default is necessary here; this script is only used in local development, so there shouldn't be any need for the prefix in most cases.

deployment/ansible/roles/driver.app/tasks/main.yml
```diff
@@ -17,20 +17,20 @@
   command: >
     docker build
       -f {{ root_app_dir }}/Dockerfile.base
-      -t quay.io/azavea/driver-app:{{ docker_image_tag }}
+      -t driver-app:{{ docker_image_tag }}
```
ddohler: This is inconsistent with the other `docker build ...` commands in other parts of the Ansible provisioning process. Initially I made comments that those other commands should be made to match this one, but I think that substituting a blank variable for `{{docker_repository}}` like you suggested yesterday is a better approach. I think you could update this to keep the `{{docker_repository}}` prefix that the other `docker build ...` commands use and then set the blank variable in the development group_vars (see the comment above), and it should work after that.

deployment/ansible/roles/driver.windshaft/templates/upstart-windshaft.conf.j2
```diff
@@ -16,7 +16,7 @@ pre-start script
   /usr/bin/docker rm windshaft || true
 
   {% if 'development' not in group_names -%}
-  /usr/bin/docker pull quay.io/azavea/windshaft:0.1.0
+  /usr/bin/docker pull {{ docker_repository }}/windshaft:0.1.0
```
ddohler: Hmm, this is an interesting wrinkle, I had forgotten about this. We don't actually build a windshaft image anywhere in the deployment process, we just use an existing Azavea image and use it to run `server.js`. So I think this should actually be hard-coded as `quay.io/azavea` (probably with a comment). Unless someone wanted to like, upgrade to a different version of Windshaft, they should be able to do everything they want by just modifying `server.js` so there'd be no need to use a custom docker image.

scripts/grunt.sh
```diff
@@ -24,15 +24,23 @@ case "$1" in
     PUBLISHED_PORT=9000
     LIVERELOAD_PORT=35731
     VOLUME_ROOT="/opt/schema_editor"
-    IMAGE_NAME="quay.io/azavea/driver-editor:latest"
+    if [[ -z "${DOCKER_REPOSITORY}" ]]; then
+      IMAGE_NAME="quay.io/azavea/driver-editor:latest"
```
ddohler: This script can only get run in development, so I think this can just be `driver-editor:latest` and remove the check for `DOCKER_REPOSITORY` (in both cases here).

deployment/ansible/roles/driver.celery/templates/upstart-celery.conf.j2
```diff
@@ -16,12 +16,13 @@ pre-start script
...
 
   {% if 'development' not in group_names -%}
-  /usr/bin/docker pull quay.io/azavea/driver-app:{{ docker_image_tag }}
+  /usr/bin/docker pull {{ docker_repository }}driver-app:{{ docker_image_tag }}
   {% endif %}
 end script
 
 exec /usr/bin/docker run \
   --name driver-celery \
+  --env DOCKER_REPOSITORY={{ docker_repository }}
```
ddohler: This line is missing a trailing `\`.

shreshthkhilani:
Added to config wizard. Ready for a last look before I merge.

pcaisse:
@shreshthkhilani I would wait for https://github.com/WorldBank-Transport/DRIVER/pull/672 to be merged and then rebase since it has a lot of changes to the wizard.

deployment/ansible/group_vars/production.j2
```diff
@@ -21,6 +21,11 @@ app_version: {{ app_version }}
 # instances are configured to use https.
 app_domain_name: {{ app_domain_name }}
 
+## DOCKER SETTINGS
+# The docker repository you want your app to pull images from. The default is "quay.io/azavea/".
+# The trailing / is required.
+docker_repository: {{ docker_repository }}
```
ddohler: Hey, I chatted with @pcaisse about this and I think we can just have this line be `docker_repository: "quay.io/azavea/"` -- the wizard doesn't need to prompt for it. There are a bunch of things that aren't covered by the wizard that people will need to go into the group_vars to edit manually anyway, so if they ever need to fork the app they'll probably have a pretty good command of the group_vars by that point.

shreshthkhilani:
I made the fix and now that docker_repository isn't prompted for in the wizard, do I still need to wait for #672 to be merged and then rebase? I think the merge will be automatic?

ddohler:
Yeah, you're right, I tested rebasing from that branch locally and it worked fine, so I think this is good to go now.

PR #679, opened 2018-03-05
ToR for batch data loading

shreshthkhilani:
## Overview

This PR adds documentation to help with loading data into a DRIVER instance. 

### Checklist

## Testing Instructions

N/A

### DIFF
--- a/doc/batch-data-loading.md
+++ b/doc/batch-data-loading.md
@@ -1,0 +1,38 @@
+# System administration: Batch data loading
+
+The DRIVER system has provisions to programatically load data. There are two easy ways to do this. Sample python scripts are provided for both methods, but you will likely need to edit these scripts to match your data's format. 
+
+## Loading data using the API
+
+A sample script that loads data using the API can be found at `scripts/load_incidents_v3.py`. It accepts the path to a directory with csv data to be loaded (`incidents_csv_dir`). It also has optional arguments that it accepts with flags:
+
+```
+--schema-path SCHEMA_PATH   Path to JSON file defining schema
+--api-url API_URL           API host / path to target for loading data
+--authz AUTHZ               Authorization header
+```
+
+It does 3 basic things: 
+
+1. Creates a new schema. It does this using the default `scripts/incident_schema_v3.json` or a custom schema specified using the `--schema-path` flag. If you wish to add records to an existing schema, all you need to do is comment out the `create_schema` function call in `main` and assign the schema uuid you want to `schema_id`. 
+
+2. Transforms the data. It transforms a row of the csv file into a json object that is compliant with the schema. It does this using the `transform` method. When you have to load batch data using this script, this is the function that should have the most changes depending on your csv. 
+
+3. Loads the data. It loads the object that it transforms a row of the csv file into using an API call. The `load` function also includes a brief sleep so as to not overload the server with API calls. Remove this at your discretion. 
+
+The script will also inform you of the progress with a log message on loading every 100 data points and on successfully loading a csv file. 
+
+
+## Loading data using Django shell
+
+This functions identically to the script that hits the API except that it runs in the Django shell and is therefore a quicker way to load a huge amount of data. If you have more than 5000 records to load, this is the recommended way to do so. A sample script is included at `scripts/django_batch_loader.py`. This script would have to be run on Django shell as follows: 
+
+```
+$ cp scripts/django_batch_loader.py app
+$ cp <csv-file(s)> app/batch
+$ vagrant ssh app
+$ sudo docker exec -ti driver-app bash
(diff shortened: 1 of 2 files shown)


doc/batch-data-loading.md
```diff
@@ -0,0 +1,28 @@
...
+2. Transforms the data. It transforms a row of the csv file into a json object that is compliant with the schema. It does this using the `transform` method. When you have to load batch data using this script, this is the function that should have the most changes depending on your csv. 
+
+3. Loads the data. It loads the object that it transforms a row of the csv file into using an API call. The `load` function also includes a brief sleep so as to not overload the server with API calls. Remove this at your discretion. 
+
+The script will also inform you of the progress with a log message on loading every 100 data points and on successfully loading a csv file. 
+
+
+## Loading data using Django shell
+
+This functions identically to the script that hits the API except that it runs in the Django shell and is therefore a quicker way to load a huge amount of data. A sample script is included at `scripts/django_batch_loader.py`.
```
ddohler: I'd add a recommendation here saying that if you have to load more than X records, it's probably better to use this script.

doc/batch-data-loading.md
```diff
@@ -0,0 +1,28 @@
+# System administration: Batch data loading
+
+The DRIVER system has provisions to programatically load data. There are two easy ways to do this.
```
ddohler: I would add something like "Sample Python scripts are provided for both methods, but you will likely need to edit these scripts to match your data's format", just so that it's clear up-front that editing Python code will be required.

scripts/django_batch_loader.py
```diff
@@ -0,0 +1,163 @@
...
+        geom=obj['geom'],
+        schema=schema,
+        data=obj['data']
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Load incidents data (v3)')
+    parser.add_argument('incidents_csv_dir', help='Path to directory containing incidents CSVs')
+    parser.add_argument('--api-url', help='API host / path to target for loading data',
```
ddohler: These API-specific parameters can be removed.

doc/batch-data-loading.md
ddohler: Also, I didn't do a good job explaining this, but the way to use this script is to run it from a Django shell, so it needs to be copied into a running app server and then run something like the following (I haven't test this but it should be close).
```
$ cp django_batch_loader.py /opt/app
$ cp <csv-file(s)> /opt/app/mmmm # <-- This is hard-coded in the source, might need to change it to something more general.
$ sudo docker exec -ti driver-app bash
$ ./manage.py shell_plus
>>> from django_batch_loader import main
>>> main()
```

scripts/django_batch_loader.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
+
+"""Loads incidents from MMDA data"""
```
ddohler: MMDA can be changed to CSV here.

scripts/django_batch_loader.py
```diff
@@ -0,0 +1,163 @@
...
+        headers = {'Authorization': args.authz}
+
+    # Do the work
+    schema_id = '95f0e797-04e5-404c-9e41-1c3ea1a5d2de'
+    schema = RecordSchema.objects.get(uuid=schema_id)
+    logger.info("Loading data")
+    count = 1
+
+    # Load all files in the directory, ordered by file size
+    files = sorted(glob.glob('mmmm/*.csv'), key=os.path.getsize)
```
ddohler: Mentioned this above, but this path could be changed to something more general.

shreshthkhilani:
Made the requested changes

scripts/django_batch_loader.py
```diff
@@ -0,0 +1,160 @@
...
+        occurred_to=obj['occurred_to'],
+        geom=obj['geom'],
+        schema=schema,
+        data=obj['data']
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Load incidents data (v3)')
+    parser.add_argument('incidents_csv_dir', help='Path to directory containing incidents CSVs')
```
ddohler: I think this is actually ignored, but that seems like bad UX, so maybe it'd be better to move this to a parameter of main(), substitute it for `batch` on line 145, and update the sample docs.

shreshthkhilani: I don't think I understand what you mean by substitute it for batch. I can just completely remove the parser, right? And this being the Django load, there wouldn't be any update needed to the docs?

ddohler: Ohh, sorry, I missed that you added instructions to copy CSV files into the `batch` folder. Yeah, this can just be deleted then. 👍

PR #685, opened 2018-05-18
Release/1.2.0

ddohler:
## Overview

The motivation for this is to be able to deploy translation updates to the Laos instance, but it seemed like a good opportunity to cut a release for some of the changes and bugfixes we made as part of the most recent project. Because we also added some new functionality, I decided to bump to 1.2.0 rather than 1.1.4.

### Checklist


CHANGELOG.md
```diff
@@ -70,4 +85,7 @@
 
 
 
+\* *This Change Log was automatically generated by [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator)*
```
shreshthkhilani: This footer should be removed since it is already at the bottom of the file on line `91`. Our changelog generator adds this every time it runs even though it already exists.

PR #687, opened 2018-06-07
Improve setup docs and configs based on Jean's onboarding

jeancochrane:
## Overview

It took me two days to get DRIVER fully set up; most of that time was spent getting well enough acquainted with Vagrant/Ansible/VirtualBox/Docker/Angular so that I could debug routine errors in the toolchain, but some of it was genuine issues with the app setup that could be merged back in.

This PR attempts to pull out the useful changes I made during my onboarding and merge them back upstream. 

### Checklist

- [X] PR has a descriptive enough title to be useful in changelogs

### Notes

For posterity, here are a few problems I ran into that turned out to require environment changes (i.e. changes to my setup and not the app itself):

- My home directory lives in an encrypted drive on my Ubuntu 16.04 machine, and it seems that it's not possible to mount shared directories into a VM using NFS when the directories live on this kind of encrypted drive. I had to make sure that the `DRIVER` and `ashlar` repos lived outside of my encrypted drive in order to get the mount to work. The particular error I was getting was something along the lines of `mount.nfs Error: /path/to/shared/folder is not a file or a directory`.
- Provisioning the `app` and `celery` VMs failed a few times due to routine config problems (e.g. I had a port collision on 5432 that I had to resolve by setting `DRIVER_DATABASE_PORT_5432=5433`). I wasn't aware that `vagrant up` and `vagrant reload` don't automatically rerun provisioning tasks; I spent a while trying to figure out why the loopback interface was denying my requests before I realized that large chunks of the app were missing because they still needed to be provisioned after I fixed the config errors. The solution to this is to run `vagrant reload --provision` or do as the docs say and manually retry `vagrant provision <VM>` when you're debugging provisioning problems.

## Testing Instructions

Most of these changes are tiny, so I don't think they need real testing. If you want to verify my work, you could:

1. Load up the Angular apps and check to make sure that disabling HTML5 doesn't have downstream problems (shouldn't affect production anyway I don't think, since the edit is to the example config)
2. Check the styling of the schema editor forms, particularly the geography upload, to make sure that the buttons display appropriately
3. Try running a data import using a virtualenv, the `scripts/requirements.txt` file, and the updated docs in `scripts/README.md`


deployment/ansible/group_vars/all.example
```diff
@@ -49,7 +49,7 @@ redis_bind_address: "redis.service.driver.internal"
 redis_port: 6379
 
 ## WEB SETTINGS
-js_html5mode: "true"
+js_html5mode: "false"
```
jeancochrane: There was some discussion about this in the DRIVER Slack channel -- it seems like Tiago ran into a similar issue. I'm still not sure why it was a problem, but the app was failing to load with HTML5 enabled due to this issue: https://stackoverflow.com/questions/37034140/location-in-html5-mode-requires-a-base

schema_editor/app/scripts/views/geography/add-partial.html
```diff
@@ -56,7 +56,7 @@ <h2>Add New Geography</h2>
                 </select>
             </div>
         </form>
-        <div class="save-area">
+        <div class="save-area col-sm-12">
```
jeancochrane: Without assigned column widths, these buttons didn't appear for me in either Chrome or Firefox.

scripts/README.md
```diff
@@ -1,13 +1,33 @@
-#Loading data for development
+# Loading data for development
```
jeancochrane: I updated the data loading docs to provide a little more background on the requirements. Apologies for the ugly diff.

scripts/requirements.txt
```diff
@@ -0,0 +1,3 @@
+python-dateutil
+pytz
+requests
```
jeancochrane: Ideally these dependencies should be pinned, but I think it would make sense to wait to see if we want to put more effort into containerizing the test data ETL before doing that. My Python environment is substantially different such that I'm not confident the versions of these packages that I installed will generalize to all machines.

scripts/README.md
```diff
@@ -1,13 +1,33 @@
...
-load_interventions.py requires `interventions_sample_pts.geojson`
+```
+pip install -U -r requirements.txt
+```
 
-generate_training_input.py requires `blackspot_training/roads_utm.*` and `blackspot_training/all_crashes_2008-2012.csv`
+## Script and schema dependencies
+
+Certain ETL scripts and schemas in this repo depend on specific data
+sources to run properly. Retrieve these dependencies from the fileshare and
```
ddohler: Outside users won't have access to our fileshare, so I don't think we should include this in the setup instructions. What might be nice is to add some of the files below to a sample_data directory so that they're in the repo. However, the incident data is from the Philippines and I think we should get authorization before including it. Perhaps we could just include these for now:
- `black_spots.json`
- `interventions_sample_pts.geojson`

That will at least allow users to generate a schema with black spots and interventions, and then perhaps manually generate some data using the web UI.

scripts/requirements.txt
ddohler: Hmm yeah, it might actually be best to hold off on this file altogether for the moment. If we add containerization it won't be necessary, and if this was what you needed to get things going for your own local machine, then it's likely to be incomplete because you probably had some stuff installed already. What might be good is to include these in the list of requirements you added to the README; then we can provide better context and explain that the listed requirements are not necessarily complete.

schema_editor/app/scripts/views/geography/add-partial.html
ddohler: Haha, nice, they actually do show up but they're waaaay off to the right. Thanks for fixing this!
screenshot-2018-6-8 ashlar editor

deployment/ansible/group_vars/all.example
ddohler: Yeah, good call; this has tripped up several people recently; I don't know why it broke in the first place but our standard practice has become setting it to false, so it should be that way here too.

schema_editor/app/scripts/views/geography/add-partial.html
jeancochrane: That's where they went! My screen must be too small to see them.

scripts/requirements.txt
jeancochrane: I like that solution -- much more clearly communicates the level of confidence I have in the completeness of the dependencies.

scripts/README.md
jeancochrane: Sounds great! I'll create the `sample_data` directory. I think the dependency table is still helpful if you have access to the fileshare, since it's the only documentation that exists for matching scripts and schemas to the example data that they rely on, but I agree that it's weird to have open docs that reference a private data directory. Is it OK to leave the table here, or would it be better to move it to a README in the fileshare folder?

ddohler: I think a README in the fileshare folder would be perfect, that's a great solution!

PR #701, opened 2018-08-30
Enable filtering records by creation date

rmartz:
## Overview
Adds a second date range filter parallel to the occurrence filter to enable a user to limit records show to ones that were created inside of DRIVER inside a date range, in addition to filtering by when they happened.

### Checklist

### Demo
screen shot 2018-08-30 at 2 46 50 pm

## Testing Instructions
- Create a record and set the Occurred date in the past (But within the past 3 months)
  - The record should be displayed
- Set the Created Date filter to end after the date the record was set to have occurred, but before present day
  - The record should not be displayed
- Adjust the Created Date filter to end in the future, after the current date
  - The record should be displayed again
- Adjust the Occurred Date filter to begin after the filter was set to have occurred, but before present day
  - The record should not be displayed

Closes #159625879


rmartz:
@ddohler Currently both filters only say "Date Range" in the toolbar, which is not very user friendly. Want to keep them short so could do something like "Occurred Range" vs "Created Range", but that feels kind of cumbersome to me. Any preferences on a phrase we could use to help distinguish the two?

tgilcrest:
Maybe not "Range", but I would suggest maybe "Occurred Date" and "Created Date"?

rmartz:
Adjusted, though that showed off a pretty significant problem... it appears like the directive's `scope` is shared between the two instances, so the dynamic settings for one are overwriting the other. There also is some weird behavior in the occurrence date filter, which I suspect might be caused by a common issue.

web/app/scripts/filterbar/date-range-directive.js
```diff
@@ -8,14 +8,21 @@
...
+            scope: true,
             link: function(scope, elem, attrs, ctlArray) {
                 var calendar = null;
-                var filterLabel = '__dateRange';
                 var filterBarCtl = ctlArray[0];
                 var dateRangeCtl = ctlArray[1];
                 var dtRange = {};  // internal min/max date strings, used for API
                 // scope.min and scope.max are localized strings for display
 
+                if(attrs.dateRangeField === '__createdRange') {
```
pcaisse: Might be good to add a space between `if` and `(`

web/app/scripts/custom-reports/custom-reports-modal-controller.js
```diff
@@ -31,7 +31,7 @@
             ctl.geoAggSelected = null;
 
             // Add the active filters for display
-            ctl.nonDateFilters = _.omit(FilterState.filters, '__dateRange');
+            ctl.nonDateFilters = _.omit(FilterState.filters, ['__dateRange', '__createdRange']);
```
pcaisse: Not a huge deal, but is there a constant we could reference instead of the string literals?

rmartz: Do you know if we have an example of sharing a constant like that across multiple files I can run off of? I'm totally down for less magic string literals, but I don't think I've seen a good example of how it'd look in JS.

rmartz: Yeah, I'm a bit worried that given my rustiness with AngularJS this may be something that's fairly easy to get wrong, especially with grunt and all that jazz. Worth keeping in mind in the future, though.

ddohler: The most AngularJs-y way that I can think of to do this would be to add them to the `FilterState` service or a similar service and pull them in as constants that way. Definitely not a big deal though.

PR #707, opened 2018-09-07
Fix loss of polygon filtering caused by Created Date filter

rmartz:
## Overview
When adding the Created Date filter, the implementation inadvertently removed functionality for filtering by polygon. Additionally, the implementation renamed the "Date Range" filter to "Occurrence Date", rather than the intended "Occurred Date" (https://github.com/WorldBank-Transport/DRIVER/pull/701#issuecomment-417433929)

### Checklist

### Demo
screen shot 2018-09-07 at 3 12 59 pm

## Testing Instructions
- In the map view, with some records visible, select the polygon or rectangle tool and draw a shape that excludes some records
  - Records not encompassed within the drawn shape should be hidden


web/app/scripts/filterbar/filterbar.html
```diff
@@ -5,6 +5,8 @@
         </ul>
         <ul class="nav navbar-nav">
             <li date-range-field="__dateRange" class="dropdown"></li>
+        </ul>
+        <ul class="nav navbar-nav">
```
rmartz: Apparently the two date filters were excessively close to each other because they shared the same `<ul>` element... normally each filter has its own `<ul>` with exactly one `<li>` inside.

May be worth at some point investigating the styles and if it's feasible to have a single `<ul>` with each filter represented by a `<li>` inside that shared `<ul>`.

ddohler: Good thought; we used to have a task to address the filter bar styling but it got removed in the latest project iteration. However, I've added this note to the epic where it used to be and if we have extra time at the end of the project we can address it.

PR #711, opened 2018-09-11
Show numeric filter for number and integer filters

rmartz:
## Overview
Fixes a broken expectation that numeric filters can be identified by `format === 'number'`. It appears only checkbox and text filters have a `format` attribute - all others must be identified via their `fieldType` attribute. 

### Checklist

### Demo
screen shot 2018-09-11 at 4 19 15 pm

## Testing Instructions
- In the schema editor, define a number and integer field for a related item of a Record Type, and mark it as "Filterable/Searchable"
  - In the UI map and record views, you should see a filter that corresponds to the fields you created

Closes [PT160325267](https://www.pivotaltracker.com/story/show/160325267)

### DIFF
--- a/web/app/scripts/filterbar/filterbar.html
+++ b/web/app/scripts/filterbar/filterbar.html
@@ -13,7 +13,7 @@
             <li text-search-field="__createdBy"></li>
         </ul>
         <ul class="nav navbar-nav" ng-repeat="(key, value) in filterbar.filterables">
-            <li ng-if="value.format === 'number'"
+            <li ng-if="(value.format === 'number') || (value.fieldType === 'integer') || (value.fieldType === 'number')"
                 numeric-range-field
                 label="key"
                 data="value"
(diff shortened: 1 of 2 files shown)


web/app/scripts/filterbar/filterbar.html
```diff
@@ -13,7 +13,7 @@
             <li text-search-field="__createdBy"></li>
         </ul>
         <ul class="nav navbar-nav" ng-repeat="(key, value) in filterbar.filterables">
-            <li ng-if="value.format === 'number'"
+            <li ng-if="(value.fieldType === 'integer') || (value.fieldType === 'number')"
```
ddohler: I don't remember why this is the case, but there are two routes to defining a number field: you can define a text field with the "number" format, or you can define a number field. I think this may be the behavior of the form editor library we're using. I suspect that `value.format === 'number'` detects the first case, so we should probably leave it in the condition unless / until we change or update the form builder to only have one way to define a numeric field.

rmartz: Good call, I forgot about those. Gave it a try and it looks like you're right, text number fields come out with `"format": "number"` but `"fieldType": "text"`
```json
{  
   "driverIncidentDetails#Wind speed":{  
      "isSearchable":true,
      "propertyOrder":7,
      "type":"string",
      "fieldType":"text",
      "format":"number",
      "multiple":false,
      "propertyName":"Wind speed"
   }
}
```
Added that back in!

PR #714, opened 2018-09-18
Fix docker command in grout playbook

ddohler:
## Overview

This was a change that @pcaisse and I discovered in the course of testing the Grout PR, but I think we got so excited about it actually working that we forgot to, you know, commit it. This fixes that and should fix any problems with migrating to grout.

### Checklist

## Testing Instructions

 * Follow the instructions in https://github.com/WorldBank-Transport/DRIVER/pull/713
 * It should work, rather than failing.

### DIFF
--- a/deployment/ansible/migrate_to_grout.yml
+++ b/deployment/ansible/migrate_to_grout.yml
@@ -13,10 +13,10 @@
   become: true
   tasks:
     - name: Move Ashlar tables to Grout
-      command: docker run --rm {{ docker_repository }}driver-app:{{ docker_image_tag }} ./manage.py migrate data 0011
+      command: docker exec driver-app ./manage.py migrate data 0011
 
     - name: Finish migrating Grout
-      command: docker run --rm {{ docker_repository }}driver-app:{{ docker_image_tag }} ./manage.py migrate grout
+      command: docker exec driver-app ./manage.py migrate grout
 
     - name: Finish migrating DRIVER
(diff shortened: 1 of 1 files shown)


rmartz:
Unfortunately I already bypassed this problem, though this matches the approach I took of using [`./scripts/manage.sh`](https://github.com/WorldBank-Transport/DRIVER/blob/d5192d6a0466b89800ee9254079d42be51e0c8e0/scripts/manage.sh) to run the commands. The only difference I can see is that `manage.sh` uses `exec -ti`, which shouldn't be needed here since we don't expect the command to require input. That said, I believe for safety we _probably_ [should use `--no-input`](https://docs.djangoproject.com/en/2.1/ref/django-admin/#cmdoption-migrate-noinput), just in case one of the systems we run this on has stale content types or some other situation that could cause `./manage.py migrate` try to be interactive.

Side note, I'm not super familiar with docker, and from my reading about this last night it seems like `run` should in theory be superior to `exec` since we don't really need to re-use a pre-existing container. I'm not entirely sure why `run` doesn't work - is it not possible to have more than one container from a given image running? What if we passed in a `--name` flag?

rmartz:
Actually just realized I still have my `pre-grout` snapshot so I can formally test this. One sec!

ddohler:
Just a comment to surface a Slack discussion about this: we determined that we can't easily use `run` because the upstart jobs that start containers set several environment variables that are required in order for Django to run, so the ansible task would have had to replicate those. It didn't seem worth the extra complexity to do this when we can easily piggyback off running containers.

PR #715, opened 2018-09-19
Add sample data for public users

rmartz:
## Overview
In order to help users get a working environment we want to provide sample data that can be used to populate the system. For simplicity this is done as statically defined sample data that can be used with the same import process that real data can be imported by.

### Checklist

### Demo
Data is anonymized from a real-world data set, with identifying information removed, dates randomly assigned, and locations adjusted according to a random normal distribution. This means that records generally follow a realistic geographic distribution but individual points do not necessarily coincide with a road and are not constrained within a fixed radius of their originating location:
screen shot 2018-09-19 at 4 15 05 pm
screen shot 2018-09-19 at 4 15 32 pm

### Notes
Currently, a couple caveats apply:
- The static sample data has a occurrence date directly associated with each record, and so as time progresses this file will become stale.
- The import process creates a new Record Type for the records it imports each time it is run, making it less desirable for use as a refresh process when records begin to cross the 3-month default occurrence window.

## Testing Instructions
- Obtain an authorization token
- Run (You likely will need to install `python-dateutil`, `pytz` and `gdal`, potentially other dependencies as well):
```bash
python scripts/load_incidents_v3.py --authz 'Token [API Token]' scripts/sample_data/
```
- You should have the sample data available 

Closes [PT160318086](https://www.pivotaltracker.com/story/show/160318086)

### DIFF
--- a/scripts/load_incidents_v3.py
+++ b/scripts/load_incidents_v3.py
@@ -32,20 +32,20 @@ def transform(record, schema_id):
 
     Doesn't do anything fancy -- if the schema changes, this needs to change too.
     """
-    details_mapping = {
-        'severity': 'Severity',
-        'description': 'Description',
-        'num_vehicles': 'Num vehicles',
-        'num_driver_casualties': 'Num driver casualties',
-        'num_passenger_casualties': 'Num passenger casualties',
-        'num_pedestrian_casualties': 'Num pedestrian casualties',
-        'traffic_control': 'Traffic control',
-        'collision_type': 'Collision type',
-        'street_lights': 'Street lights',
-        'surface_condition': 'Surface condition',
-        'surface_type': 'Surface type',
-        'main_cause': 'Main cause'
-    }
+    details_mapping = [
+        ('severity', 'Severity', str),
+        ('description', 'Description', str),
+        ('num_vehicles', 'Num vehicles', int),
+        ('num_driver_casualties', 'Num driver casualties', int),
+        ('num_passenger_casualties', 'Num passenger casualties', int),
+        ('num_pedestrian_casualties', 'Num pedestrian casualties', int),
+        ('traffic_control', 'Traffic control', str),
+        ('collision_type', 'Collision type', str),
+        ('street_lights', 'Street lights', str),
+        ('surface_condition', 'Surface condition', str),
+        ('surface_type', 'Surface type', str),
+        ('main_cause', 'Main cause', str)
+    ]
 
     # Calculate value for the occurred_from/to fields in local time
     occurred_date = parser.parse(record['record_date'])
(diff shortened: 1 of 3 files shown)


scripts/sample_data/sample_traffic.csv
```diff
@@ -0,0 +1,978 @@
+surface_type,collision_type,severity,surface_condition,num_driver_casualties,lon,traffic_control,num_vehicles,num_passenger_casualties,num_pedestrian_casualties,record_date,lat,main_cause
+Asphalt,Hit parked vehicle,Property,Dry,0,0,None,1,0,0,06/18/18,0,Human error
+Concrete,Side swipe,Property,Dry,0,-0.00014008239912971835,Traffic lights,2,0,0,06/18/18,0.00020447453828494324,Human error
```
ddohler: These coordinates don't look right -- they're all clustered around (0,0) (off the west coast of Africa), which makes me think that perhaps rather than _adding_ jitter to the coordinates, each coordinate was _replaced_ with the jitter?

scripts/load_incidents_v3.py
```diff
@@ -121,6 +121,7 @@ def create_schema(schema_path, api, headers=None):
                              data={'label': 'Incident',
                                    'plural_label': 'Incidents',
                                    'description': 'Historical incident data',
+                                   'temporal': True,
                                    'active': True},
                              headers=headers)
     response.raise_for_status()
```
ddohler: Nothing you introduced, but I noticed when testing this that the data for the events that are created doesn't quite line up with the schema, which is causing the record list and record editors to show blanks for all the fields. It's a trivial fix though: Lines 51-53 need to have their keys changed to match what is in `incident_schema_v3.json`. This should be as simple as prefixing with `driver` and camel-casing, i.e.:
- `incidentDetails` -> `driverIncidentDetails`
- `vehicle` -> `driverVehicle`
- `person` -> `driverPerson`

scripts/sample_data/sample_traffic.csv
rmartz: Well that's frustrating... I fixed and committed this before I left on vacation, but apparently forgot to push the change to `origin`. This should be fixed there now 😣

scripts/load_incidents_v3.py
rmartz: I changed this, though it appears some attributes under `driverIncidentDetails` fail Grout's type-based validation (e.g., `u'0'` fails `number` validation). I added a typecast function to the `details_mapping` list, which allows that to pass but that revealed that some of the data appears to be legitimately invalid (`u'11' is not one of [u'Head on', u'Rear end', u'Right angle', u'Side swipe', u'Overturned vehicle', u'Hit object in road', u'Hit object off road', u'Hit parked vehicle', u'Hit pedestrian', u'Hit animal']`). I'm going to investigate the origin of this and see if it's feasible to fix or exclude invalid rows, or if I should drop the bad column from the sample data entirely.

rmartz: This should be good now, changed my anonymization script to exclude rows that were missing lat/lon coordinates or had numeric or blank values for required string parameters. Filtered out a good almost 20% of the records, but everything we have now should be schema appropriate.

rmartz:
The source data actually didn't have time for the records - I kind of liked the idea of trying to preserve relative time-of-day (In case we want to do analytics to see if some types of incidents are correlated with evening or midday, for example), but the source only had the `record_date` with no time component, so I wasn't able to keep that 🙁

PR #716, opened 2018-09-19
Add `group_vars/all` to source control

shreshthkhilani:
## Overview

Add `group_vars/all` to source control.

### Checklist

## Testing Instructions

 * Local provision should work
 * Travis should pass

Closes [#160586495](https://www.pivotaltracker.com/story/show/160586495)


.travis/deploy_staging.sh
```diff
@@ -13,8 +13,8 @@ done
 # Set up group vars
 echo "setting up group vars..."
 set +x
-grep -v 'nominatim_key\|app_version:' deployment/ansible/group_vars/all.example \
-    > deployment/ansible/group_vars/all
+grep -v 'nominatim_key\|app_version:' deployment/ansible/group_vars/staging \
```
ddohler: The old version of this was essentially doing the same thing as copying `all.example` -> `all`, but on a per-variable basis. If everything is in source control it may not be necessary to do any copying at all.

ddohler: Oh, I misread the command -- this is literally copying the whole file _except_ `nominatim_key` and `app_version`, and then manually inserting `app_version` as a string literal on the next line (and `nominatim_key` and other secrets below that). The changes we're making should make copying `all.example` unnecessary, and since `group_vars/staging` can just include `app_version: latest` directly within the file, the second one shouldn't be necessary either. 

So I think these first two lines can just be deleted as long as `group_vars/staging` has `app_version` defined and doesn't define `web_js_nominatim_key`. The ones that insert secrets will need to remain though, in order to avoid committing the secrets.

README.md
```diff
@@ -31,14 +31,9 @@ DRIVER - Data for Road Incident Visualization, Evaluation, and Reporting
       (If you just want to install the DRIVER web interface, do this. You can add Android integration later.)
 
     - To build schema model JAR files for the Android app, copy the signing keystore to `gradle/data/driver.keystore`
-    and set the password for the keystore under `keystore_password` in `deployment/ansible/group_vars/all`.
+    and set the password for the keystore under `keystore_password` in `deployment/ansible/group_vars/development`.
```
ddohler: It might be helpful to add a note in the README suggesting `git update-index --assume-unchanged deployment/ansible/group_vars/development` to prevent changes to this file from being tracked by git. The reverse is `--no-assume-unchanged` and that'd be good to include also in case someone wants to update the default development settings.

.gitignore
```diff
@@ -80,6 +80,3 @@ docs/_build/
 
 # PyBuilder
 target/
-
-# group_vars config
-deployment/ansible/group_vars/all
```
ddohler: We should also delete `deployment/ansible/group_vars/all.example` as part of this PR.

.travis/deploy_staging.sh
```diff
@@ -13,20 +13,16 @@ done
 # Set up group vars
 echo "setting up group vars..."
 set +x
-grep -v 'nominatim_key\|app_version:' deployment/ansible/group_vars/all.example \
-    > deployment/ansible/group_vars/all
-echo "app_version: \"latest\"" \
-    >> deployment/ansible/group_vars/all
 echo "web_js_nominatim_key: \"${NOMINATIM_API_KEY}\"" \
```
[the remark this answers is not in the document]
shreshthkhilani: They aren't in staging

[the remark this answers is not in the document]
ddohler: It looks like they still are (even though they're blank), i.e. https://github.com/WorldBank-Transport/DRIVER/pull/716/files#diff-b24154a4151c546c7f0a31a4625034dbR45 (I meant removing the variables entirely, not emptying their values; sorry if that was unclear).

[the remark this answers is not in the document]
shreshthkhilani: Sorry, I was looking at the wrong version of the file!

shreshthkhilani:
This should be good to look at again. Thanks!

PR #717, opened 2018-09-20
Fix map centering after geocoding address

shreshthkhilani:
## Overview

Fix map centering after geocoding address. Currently, after searching the first time and zooming/panning the map, the second search won't work center correctly on the first click. 

### Checklist

### Notes

Kind of unrelated: How do we handle cases when Pickpoint returns no data? Currently, we do nothing. 

Some addresses that work: `64 xavier st`, `160 Luna Mencias`

## Testing Instructions

 * Search for a latlng or an address in the Philippines. 
 * Pan/zoom a bit.
 * Search for another latlng or an address.
 * Ensure that it centers correctly on the first click.

Closes [#155396666](https://www.pivotaltracker.com/story/show/155396666)

### DIFF
--- a/web/app/scripts/state/zoom-to-address-directive.js
+++ b/web/app/scripts/state/zoom-to-address-directive.js
@@ -30,8 +30,7 @@
                     if (latlng === null) {
                         return;
                     }
-                    map.setZoom(16);
-                    map.panTo(new L.LatLng(parseFloat(latlng[0]), parseFloat(latlng[1])));
+                    map.setView([parseFloat(latlng[0]), parseFloat(latlng[1])], 16);
                 };
 
                 var geocodeAddress = function(query) {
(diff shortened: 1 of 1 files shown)


ddohler:
I think it would be good to show some kind of alert when there's no geocoding result, but I think it would be good to wait until the WB and UI/UX have a chance to look at this; we can bring up that question as part of those discussions (although I suspect it'll come up naturally even if we don't).

shreshthkhilani:
Thanks! @ddohler should I add a trigger to search when hitting enter? Or wait on Design/WB for that too?

ddohler:
If you want to and think it'll be easy, I'd say go for it -- even if we end up changing it later, it'll at least make things easier on us for the time being.

shreshthkhilani:
@pcaisse, could you test that hitting enter triggers the search? Thank you!

pcaisse:
@shreshthkhilani LGTM!

PR #719, opened 2018-09-21
Save original record every time a Record is created

shreshthkhilani:
## Overview

Extends the model for `RecordAuditLogEntry` and writes the entire record (as a JSON) as well as an MD5 hash signature of the JSON string to it on create. 

### Checklist

## Testing Instructions

 * Create a new record on the UI
 * Verify that for that new record, an audit log entry was created and has a valid `log` and `signature` field
 * On the Django shell, get that `RecordAuditLogEntry` and ensure that `rale.verify_log()` returns `True`
 * Change the `log` field, save the record, and ensure that `rale.verify_log()` returns `False`

Closes [#160206283](https://www.pivotaltracker.com/story/show/160206283)


app/data/models.py
```diff
@@ -56,6 +57,24 @@ def as_list(cls):
 
     action = models.CharField(max_length=6, choices=ActionTypes.choices)
 
+    weather = models.CharField(max_length=50, null=True, blank=True)
```
ddohler: A couple preliminary thoughts (and apologies if this wasn't intended for review yet):
- This adds a pretty tight coupling between DriverRecords and RecordAuditLogEntries; if we change a field on DriverRecord we'd always need to change this (and the creation code down below) too.
- One of the things I liked about the idea of putting this change tracking on the audit log entries was the possibility for later extending it to track other types of changes, like modifications or deletions. Do you have thoughts about how we might do that with this architecture? It's not necessary as part of this task but I think it'd be good to think it through.

shreshthkhilani: This commit was just work I did to check if I could change models, make Django migrations work, and see the required results on creating records via the UI. 

I plan on editing the audit log model to not replicate the driver record field-to-field. In the past, I've set up audit logs to have a jsonb field with a json of all the info a model has, and I intended to ask on the Slack channel whether that's something we'd like to do. I think that that would also solve the two problems you've raised?

shreshthkhilani: More detail about how a jsonb field would help track changes. The JSON could look something like this: 
```json
{
  "old": {},
  "new": {}
}
```
The `old` and `new` keys would have the state of the object prior to this action and the `new` key would have the state of the object after this action. 
In case of a create, the `old` key would be `null`. In case of delete, `new` would be `null`. In case of an update, `old` and `new` would both be non null.

Alternatively, we could also just track the `new` state.

ddohler: Gotcha; that all makes sense to me. One thing I'm wondering about with JSONB is how to consistently generate the hash since it will be sensitive to things like white space which might differ depending on how the JSONB field gets serialized. But that's a more minor thing -- I think overall a JSONB field makes sense and should do the trick for resolving both those issues.

shreshthkhilani: To hash the JSON, we'd need to convert it to a string and we could just use `json.dumps(dict, sort_keys=True)`?

rmartz: Do we need searching for this field? Concern over whether a static JSONB value has consistent representation in Postgres [seems valid to me](https://www.postgresql.org/docs/9.4/static/datatype-json.html):

> Because the `json` type stores an exact copy of the input text, it will preserve semantically-insignificant white space between tokens, as well as the order of keys within JSON objects.  [...] By contrast, `jsonb` does not preserve white space, does not preserve the order of object keys, and does not keep duplicate object keys.

JSONB's advantages are that it converts data into a more efficient but representationally lossy binary form and is searchable, whereas JSON seems to be basically just a wrapper around a TEXT field. Since we likely would prefer a consistent representation and don't need searchability, JSONB seems like it may not be the best fit of the two.

shreshthkhilani: That's fair. I wasn't aware that Postgres had a just JSON type as opposed to a JSONB. It doesn't need searching on that field, so JSON sounds good.

shreshthkhilani: But does it really matter which field it's stored in the db? If we're hashing the dict, it would be on the Python code (`hashlib.md5(json.dumps(dict, sort_keys=True)).digest()`) and then stored in another column, so we're agnostic to whether it's JSON or JSONB?

app/data/models.py
```diff
@@ -56,6 +57,14 @@ def as_list(cls):
 
     action = models.CharField(max_length=6, choices=ActionTypes.choices)
 
+    # The log JSON will contain `old` and `new` state of the model
+    log = models.TextField(null=True)
+    # Singature will contain an MD5 hash of the log field
+    signature = models.BinaryField(null=True)
```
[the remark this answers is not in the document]
shreshthkhilani: I don't know of an advantage. Would you like me to change it to be a string?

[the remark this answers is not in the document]
ddohler: I probably would: it'll make debugging easier, and if we want to display this to users we'd have to convert the binary to hex anyway.

shreshthkhilani:
@ddohler I made the change; do you want to have a look before I merge?

ddohler:
Looks good to me!

PR #731, opened 2018-11-07
Translate "Log in to DRIVER" as a complete phrase

ddohler:
## Overview

The WB requested that we change the translation on the login page. This fixes some issues that arose while trying to make that change (and makes that change).

### Checklist

### Demo

screenshot from 2018-11-07 14-31-01

### Notes

We used to display the phrase "Log in to DRIVER" as `{{ 'LOGIN.LOG_IN_TO' | translate }} {{ 'COMMON.APP_NAME' | translate }}`. However, this assumes that the translation of "Log in to" + the translation of "DRIVER" is the same in every language. There are numerous grammatical reasons that this could be incorrect, and it appears that something like this motivated the requested change to the Laos translation, because the new translation is something along the lines of (according to Google Translate) "access system".

I didn't want to change the APP_NAME to be "system", so instead I combined the two tokens into a single phrase so that it can be translated properly. I copied the existing text into the new phrase for all the other languages we support. It may be that none of the other languages that DRIVER has been translated into so far have exhibited this problem because no one has complained yet. It's also possible that the other languages do have this problem but no one has noticed--in this case it's now easy to fix.

## Testing Instructions

 * Add Laos and perhaps a few other languages to your `group_vars/development`
 * `vagrant provision app`
 * Confirm that Laos displays ການເຂົ້າສູ່ ລະບົບ on the login-screen. 
 * Confirm that no other languages have changed.

Closes #161781136


rmartz:
We should file a bug for `LOG_IN_WITH` as well, and potentially any other translations we have that rely on automatically concatenating two strings... not all languages structure their sentences the same way as English does.

ddohler:
That's a good point, I've added an issue for it!

PR #738, opened 2018-12-20
Fix duplicates filtering on Django app

shreshthkhilani:
## Overview

The duplicates view on the app is broken. This is because the filtering on Django returns a Type Error. This PR fixes this error by modifying the arguments of a method to make it what the `django_filters` package expects. This likely broke sometime during the Django upgrade so is only broken on staging and not on production. 

### Checklist

## Testing Instructions

 * Make sure you have some duplicates. Add a record that is at the same time and same lat lon as an existing record then run the management command to generate duplicates in the app container `./manage.py find_duplicate_records`.
 * Next, navigate to the duplicates window on the UI to recreate the error.
 * Pull this branch, see the error disappear. 

Closes #XXX


shreshthkhilani:
Build failing for now because of Quay outage.

PR #760, opened 2019-03-07
Reconcile differences in production.j2 and production.example

rmartz:
## Overview
Over time, a small number changes have been made to either `production.example` but not `production.j2` or vise versa. This confuses which one should be used, which is a significant impediment when creating a new instance or performing an instance upgrade.

This applies observed changes to the corresponding file, so the only differences should be differences of value.

### Checklist

## Testing Instructions
- Run
```
diff deployment/ansible/group_vars/production.{j2,example}
```
- The only differences should be variable value templating and `{% raw %}`/`{% endraw %}`

### DIFF
--- a/deployment/ansible/group_vars/production.example
+++ b/deployment/ansible/group_vars/production.example
@@ -22,6 +22,7 @@ app_version: "0.0.0"
 app_domain_name: "your.driver-domain.here"
 
 ## DOCKER SETTINGS
+docker_image_tag: "{{app_version}}"
 # The docker repository you want your app to pull images from. The default is "quay.io/azavea/".
 # The trailing / is required.
 docker_repository: "quay.io/azavea/"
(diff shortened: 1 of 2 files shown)


deployment/ansible/group_vars/production.example
```diff
@@ -22,6 +22,7 @@ app_version: "0.0.0"
 app_domain_name: "your.driver-domain.here"
 
 ## DOCKER SETTINGS
+docker_image_tag: "{{app_version}}"
```
rmartz: This seems like something that might fit in the `ADVANCED` section at the bottom, since it's just a reassignment of an already provided value that I suspect the user won't need to set themselves, like what we do with `allowed_host`:
https://github.com/WorldBank-Transport/DRIVER/blob/3ae5b47e182e4f2074beef0a0a0d2f31d80084a4/deployment/ansible/group_vars/production.example#L260-L263

PR #763, opened 2019-03-18
 Remove remaining traces of Ashlar

rmartz:
## Overview
Removes references to Ashlar as able. Some remnants remain, in:
- CHANGELOG
- Existing migrations (Provisioning a new DRIVER instance installs migrations successfully)
- Ashlar Schema Editor

## Notes
Also fixes problems encountered preventing provisioning by upgrading Docker to version 17.

### Checklist

## Testing Instructions
- Make sure that no `ashlar/` directory exists in the same directory as your `DRIVER/` repo
- `vagrant provision`
 - The web interface should still work
- (Extra credit)
  - Delete the VMs with `vagrant destroy`
  - Reprovision from scratch with `vagrant up`
    - The web interface should still work

Closes [#164466001](https://www.pivotaltracker.com/story/show/164466001)

### DIFF
--- a/app/Dockerfile.base
+++ b/app/Dockerfile.base
@@ -3,7 +3,7 @@ FROM quay.io/azavea/django:1.11-python2.7-slim
 # N.B. This will not work for docker versions above 1.10.x; Docker 1.11 and above contain
 # multiple binaries in a tarball; if we upgrade Docker to the 1.11 branch or beyond then
 # the RUN command below will need to be updated.
-ENV DOCKER_VERSION 1.10.3
+ENV DOCKER_VERSION 17.06.2
 
 # install libgdal1-dev
 RUN echo "deb http://deb.debian.org/debian jessie main contrib non-free " > /etc/apt/sources.list.d/jessie.list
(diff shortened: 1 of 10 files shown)


rmartz:
> I needed to remove and touch /etc/exports on my host machine for Vagrant to stop throwing errors for me.

That's really weird - I'm not familiar with that file or what it's used for, do you think that could be a sign of some kind of problem?

shreshthkhilani:
I don't think it's a problem. I think it's an NFS on Mac thing––it was looking for the directory mapping because it previously wrote it to that file. https://stackoverflow.com/questions/20726248/vagrant-error-nfs-is-reporting-that-your-exports-file-is-invalid/33852720

rmartz:
Ah ok, glad you got that figured out!

app/Dockerfile.base
```diff
@@ -3,7 +3,7 @@ FROM quay.io/azavea/django:1.11-python2.7-slim
 # N.B. This will not work for docker versions above 1.10.x; Docker 1.11 and above contain
```
pcaisse: Can this comment be deleted now?

rmartz: It appears so - in theory something must have happened previously to make it outdated because I didn't notice any problems for this, but now that we are on >1.10.x it's pretty clearly obsolete now.

PR #765, opened 2019-03-22
Prevent failed post-stop script task from stopping Upstart respawning

rmartz:
## Overview
After one of our Docker processes ends, we configure Upstart to kill the container and delete it. However, because the process has ended, the command Upstart is configured to kill the container fails:
```
Error response from daemon: Cannot kill container driver-gradle: Container 6c9afdbe941ae917002f069761f1584d0093ac420f330333a364241a9c9a5d28 is not running
```

This failed `docker kill` command would cause Upstart to abort the process respawn, leaving the service stopped.

It's not clear if `docker kill` is needed in the `post-stop script` section, but in order to prevent a failure from stopping Upstart from attempting to respawn the process this appends `|| true` to each command so that they will be respawned even if one of the `post-stop` commands fails.

### Checklist

## Testing Instructions
- `vagrant provision celery`
- Open a terminal and SSH into the celery server
- `sudo tail -f /var/log/upstart/driver-gradle.log`
- Open a second terminal and SSH into the database server
- `sudo service redis-server stop`
  - You should see messages in the `celery` SSH terminal showing the Gradle process being repeatedly respawned and failing 
- `sudo service redis-server start`
  - You should see the messages stop and diagnostic messages showing the process is waiting for messages
- Exit the database server
- `vagrant reload database`
  - You should see messages in the `celery` SSH terminal showing the Gradle process being respawned, and recover when the database server has finished reloading

Closes [#164140129](https://www.pivotaltracker.com/story/show/164140129)

### DIFF
--- a/deployment/ansible/roles/driver.app/tasks/main.yml
+++ b/deployment/ansible/roles/driver.app/tasks/main.yml
@@ -33,10 +33,10 @@
   command: /usr/bin/docker pull {{ docker_repository }}driver-app:{{ docker_image_tag }}
   when: staging or production
 
-- name: Configure Driver application service definition
+- name: Configure DRIVER application service definition
   template: src=upstart-app.conf.j2 dest=/etc/init/driver-app.conf
 
-- name: Restart Driver application
+- name: Restart DRIVER application
   service: name=driver-app state=restarted
 
 - name: Set up monit monitoring of driver-app container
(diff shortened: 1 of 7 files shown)


rmartz:
Just pushed a commit that should fix provisioning the Celery server not restarting the Gradle server and leaving it running an old version of the code.

PR #770, opened 2019-04-05
Fix Blackspot failures for Mumbai

rmartz:
## Overview
Fixes a variety of errors that would prevent Mumbai from making use of Blackspots. Errors include:
- Including deleted records in CSV export (And so Blackspot analysis)
- Incorrect process for installing Docker inside of Docker (Appears to be the warned fallout from ambiguous comment https://github.com/WorldBank-Transport/DRIVER/pull/763#discussion_r267941831)
- Failed to install R package `caret` due to R version requirement shift

Also changes:
- Fixes error in API causing deleted records to still be shown to the user (This was thought to potentially be connected to the similar CSV export error)
- Moves blocking asynchronous Celery calls in blackspot logic to happen in process (Blackspot logic already runs in the Celery Docker environment, and having the tasks passed through Celery added overhead and caused obfuscated error messages)
- Changes CSV export to use generator structure (Reduce memory pressure storing duplicate values and improve performance)
- Creates and compresses tarball in Python (Previously used bash command to compress, resulting in obfuscated error message if the task failed)

### Checklist

### Demo
Screen Shot 2019-04-05 at 11 27 31 AM

## Testing Instructions
- Load database dump of Mumbai records
- Change `group_vars/development` to reflect Mumbai values
```diff
diff --git a/deployment/ansible/group_vars/development b/deployment/ansible/group_vars/development
index 254dd5b..48b1d56 100644
--- a/deployment/ansible/group_vars/development
+++ b/deployment/ansible/group_vars/development
@@ -80,10 +80,10 @@ oauth_client_id: ''
 oauth_client_secret: ''

 ## Localization settings
-local_time_zone_id: 'Asia/Manila'
-local_country_code: 'ph'
-local_center_lat_lon: [12.375, 121.5]
-osm_extract_url: 'https://download.geofabrik.de/asia/philippines-latest.osm.pbf'
+local_time_zone_id: "Asia/Calcutta"
+local_country_code: 'in'
+local_center_lat_lon: [19.076, 72.878]
+osm_extract_url: "https://download.geofabrik.de/asia/india-latest.osm.pbf"

 # List of languages to allow in the application. There must be at least one defined,
 # and the first one in the list is used as the application's default language.
```
- Provision VMs
- SSH in to `web` VM and become root
- Open `shell_plus` terminal
  - `docker exec -it driver-app python ./manage.py shell_plus`
- Delete all `RoadSegmentsShapefile` records
  - `RoadSegmentsShapefile.objects.all().delete()`
- In a new terminal, SSH into `celery` VM and become root
- Run blackspot analysis
  - `docker exec $(docker ps -q -f name=driver-celery) python ./manage.py calculate_black_spots`
  - Warning: This likely will take about 3 hours, and is very CPU intensive. If you have a laptop you'll want to be plugged in while this is running.
- It should complete
- Open web interface
  - There should be red blackspots shown in the blackspots map

Closes [PT163071747](https://www.pivotaltracker.com/story/show/163071747)


rmartz:
I'm currently doing some memory profiling to attempt to minimize the risk of OOM failures for this process. Because memory optimizations were done along side other fixes I don't have a graph of memory usage in stock code, but with the code as it exists currently this is the `mprof` graph of memory usage over time:
Screen Shot 2019-04-09 at 11 30 03 AM

I believe the initial 3 flat tiers correspond to the 3 steps of generating a segments tar file (`load_road_network()`, `get_segments_shp()`, and `create_segments_tar()`), as in prior testing I generated a profile where those steps were skipped and the spike after those tiers was much earlier  (Unfortunately I did not retain the generated profile from then).

The spike concerns me, and I believe it corresponds with `get_training_noprecip()`, which internally uses a `SpooledTemporaryFile()` to write a generated CSV file in memory rather than to disk.

The tier after the spike I believe is while the final stage of the R machine learning is being performed, which also concerns me as the graph does not include memory required by that process and having 250MB of memory already reserved by this process may be unnecessary pressure on the system. I'm currently working on structuring the calculation logic as best I can to minimize the amount of content that Python has in scope during that type, in hopes that garbage collection can reduce the memory used passively while the R code is running to a relatively low level.

rmartz:
Did some restructuring to encapsulate intermediate values in function scopes so they can be cleaned up after they're used. This brought the Python background memory use overhead during the R machine learning portion down slightly (About 10%), not sure what would be needed to reduce it further:
Screen Shot 2019-04-09 at 12 56 32 PM
(This graph is missing the three tiers mentioned above as I ran the test with a pre-existing segments TAR file, this may affect results but this test took 20 minutes and generating the TAR file takes about 3 hours, so in the interest of speediness I figured it was a reasonable compromise)

My impression is that most values in memory when we start the `driver-analysis` container are  IDs and paths, but 220-230MB isn't something to sneeze at (Especially when we expect to have our machine learning process running concurrently). For the moment this may be a point to settle with the performance we're seeing now, and if we continue to run into memory problems later on we can investigate further.

Of note, in investigating if we can reduce the spike seen, it appears functions in `generate_training_input.py` are built with multiple purposes merged together ([Single Responsibility principle](http://www.principles-wiki.net/principles:single_responsibility_principle)). This is convenient for combining repeated correlated tasks into the least number of constituent functions, but structuring for the least number of functions comes at the cost that those functions provide all-or-nothing behavior. In our case, that cost means that we have very limited flexibility in reorganizing steps to minimize what values are kept in memory concurrently.

If the values produced share the same lifespan then having all-of-nothing behavior can work out, but for example `get_training_noprecip()` requires `segments` for later tasks while 
`records` is only used once, which makes it an ideal candidate for being removed from scope after use. However, as both values are currently produced by `get_data()` (Though they are produced from different files, along with a large number of other potentially disentangleable values) it's impossible to design a function where one value drops out of scope implicitly. We would have to add in explicit value management in that a function returns one value but not the other, which exacerbates the problem of having inflexible functions with confusing (and seemingly random) return values.

rmartz:
Rebased to remove reverted commit with script that tried to migrate bad records in-place, also fixed a couple typos in commit messages (Doh)

PR #774, opened 2019-04-18
Use auto-incrementing labels for related items

kshepard:
## Overview

Users were confused about the way we were handling labels of related objects. We decided to instead opt for using the object name followed by an auto-incrementing label. Some of the required changes for this were done in the `json-editor` -- see https://github.com/azavea/json-editor/pull/11

### Checklist

### Demo

Screenshot_2019-04-18 DRIVER

Screenshot_2019-04-18 DRIVER(2)

### Notes

This intentionally doesn't alter any behavior of the Android app, and will also not interfere with how the Android app is performing the labeling. There may be future changes to the Android app to align it appropriately.

## Testing Instructions

 * Provision your `app` VM and load the site up in a browser
 * Go to the record list view and experiment with viewing/creating/updating records with relationships
 * Verify the auto-incremented labels are displayed in each place the relationships are displayed or are selectable

Closes [PT164899426](https://www.pivotaltracker.com/story/show/164899426)


web/app/scripts/details/details-reference-controller.js
```diff
@@ -1,33 +1,25 @@
...
-                    }
-                    // TODO: do we need to make changes for any displayed substrings
-                    // (and limitTo filters) in the app for right-to-left languages?
-                    return reference[key].substring(0, maxKeyLength) + ellipsis;
-                }).join(' ');
+            // There will only ever be one matching index. All other related info objects will
+            // have a -1 populated here, so we can just take the maxiumum of the array
+            var index = _.max(candidateIndices);
+            if (index > -1) {
+                ctl.referenceDisplay = ctl.property.propertyName + ' ' + (index + 1);
```
rmartz: Since we have the `if (index > -1) {` check, do we need a fallback branch in case no candidate matches?

rmartz: Talked with @kshepard and did some testing - it looks like situations where this falls through happens fairly frequently whenever the record is blank, so the default behavior leaving `referenceDisplay` unchanged seems perfectly fine.

